Issue #40, opened 2022-03-16
`onyo init` fails in an existing git repo


aqw:
According to the README, `onyo init` can be run on an existing git repository.

> [...] The directory will be initialized as a git repository (if it is not one already), the `.onyo/` configuration folder created, and template files generated and committed. [...]

However, the following fails:
```
❱ mkdir /tmp/tst
❱ cd /tmp/tst
❱ git init
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint:   git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint:   git branch -m <name>
Initialized empty Git repository in /tmp/tst/.git/
❱ onyo init
ERROR:onyo:/tmp/tst is already a git-repository.
```

TobiasKadelka:
That it again prints "succesfully initialized the onyo repo" should not happen. Should it print a warning/info/nothing instead?

aqw:
I think you're missing the point of the bug report.

The README states that the user can run `onyo init` on an existing git repository. This is currently not the case. Onyo errors instead, and does not create a `.onyo` folder.

```
❱ ls -lah
total 48K
drwxr-xr-x  3 aqw  aqw  4.0K Mar 18 11:16 .
drwxrwxrwt 23 root root  36K Mar 18 11:16 ..
drwxr-xr-x  7 aqw  aqw  4.0K Mar 18 11:16 .git
❱ onyo init
ERROR:onyo:/tmp/tst is already a git-repository.
❱ ls -lah
total 48K
drwxr-xr-x  3 aqw  aqw  4.0K Mar 18 11:16 .
drwxrwxrwt 23 root root  36K Mar 18 11:16 ..
drwxr-xr-x  7 aqw  aqw  4.0K Mar 18 11:16 .git
```

aqw:
> I think for now I will print an info if already is onyo and git, but print a warning if it is already git, but not onyo, and still create an .onyo folder.

Sorry, I just saw this (since it was edited).

This is still not quite correct.

If the dir is empty:
  - init git and onyo
  - print success

 If the dir contains a git repo but not onyo dir
  - print information about presence of git repo
  - init onyo
  - print success

 If the dir contains both git and onyo dirs
[...]

PR #46, opened 2022-03-23
Update onyo Init output

TobiasKadelka:
PR gives more differentiated and specific ouput for different situations in which `onyo init` is called (e.g. in an existing git repository, in an empty repository, in an onyo repository).
Always displays info when `onyo init` is called, and updates tests for the new output of `onyo init`.


aqw:
A few comments on this PR:
- despite closing #40, this does not seem to address the situation where an onyo directory is present, but it is not a git repository.
- The language distinction between ".onyo directory" and "git repository" is confusing to me. As a user, I'm left wondering what distinction is trying to be communicated to me.
  - I recommend either consolidating both to use the term "repository" or adding a descriptive phrase to "directory", such as ".onyo configuration directory".

PR #47, opened 2022-03-23
onyo: add debug mode that prints all commands that are run

TobiasKadelka:


onyo/main.py
```diff
@@ -16,6 +16,15 @@ def parse_args():
     parser = argparse.ArgumentParser(
         description='A text-based inventory system backed by git.'
     )
+    parser.add_argument(
+        '-d',
+        '--debug',
+        #metavar='debug',
```
aqw, line 22: Is this planned for something?

onyo/main.py
```diff
@@ -16,6 +16,15 @@ def parse_args():
...
+        required=False,
+        default=False,
+        action='store_true',
+        help='wayyy more output'
```
aqw, line 26: I think a dry "enable debug logging" communicates a similar level of detail.

PR #48, opened 2022-03-23
tests: runable from other directories/systems; adds github action to run tests automatically

TobiasKadelka:
Instead of using a hardcoded file path, the default_tests.py uses the location relative to itself to find the files to compare the ouput to, and to creates a folder to run all the tests in. Instead of using a set path in the output files, replaces paths in the output with a path for each test, depending on the location the test was run from.

Sets up a github action file to run the tests automatically for each push.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,18 +6,17 @@ jobs:
   build-linux:
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 5
+      matrix:
+        python-version: ['3.5', '3.6', '3.7', '3.8', '3.9', '3.10', 'pypy3']
 
     steps:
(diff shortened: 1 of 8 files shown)


aqw:
If there's no compelling reason to use `conda`, I recommend simply using pip.

.github/workflows/main.yml
```diff
@@ -6,14 +6,15 @@ jobs:
   build-linux:
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 5
+      matrix:
+        python-version: [ '2.x', '3.x', 'pypy2', 'pypy3' ]
```
aqw, line 10: No need to test `2.x` if `3.5` is the minimum supported version.

PR #63, opened 2022-03-29
CI: trigger with PR, add flake8 linting

TobiasKadelka:
CI will be triggered with pushs and PRs.

Adds linting with flake8, and a flake8 config file (specifies which rules are to follow/ignore) and updates code base to these rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,15 @@ laptop asset should be converted to a folder, and the license moved into it.
 
   Print the contents of `asset` to the terminal without parsing or validating
   the contents.
+- `onyo rm [--quiet, -q] [--yes, -y] asset | directory...`:
+
+  Delete the `asset`(s) and `directory`(s).
+
+  Onyo will present a complete list of all files and folders to delete, and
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -22,6 +23,10 @@ jobs:
     - name: Install dependencies
       run: |
         pip install .
+        pip install flake8
+    - name: flake8 linting
+      run: |
+        flake8 $(git ls-files onyo/*.py onyo/commands/*.py tests/*.py)
```
aqw, line 29: `find` might be a better solution here, as it will be recursive, and thus less brittle. Something like:
```
find . -type f -name "*.py"
```

aqw (changes requested):
Overall, this PR looks good to me. The `find` change would be nice, as it decreased the change of missing new dirs (and subdirs) that are introduced in the future.

The one blocker that I see is the first merge commit. This PR should be rebased on top of `main`, and only have your 3 commits.

Issue #42, opened 2022-03-16
`onyo new` does not handle spaces correctly


aqw:
The README does not disallow spaces for the `type`, `make`, `model`, and `serial` fields (and perhaps it should). But `onyo new` handles spaces poorly. An example:

```
❱ onyo new groups/dp/aqw    
<type>*:laptop
<make>*:With an absolute path, a unix local named pipe is used for control. The file is created with user and group that is configured and access bits are set to allow members of the group access. Further access can be controlled by setting permissions on the directory containing the control socket file. The key and cert files are not used when control is via the named pipe, because access control is via file and directory permission.
<model*>:T490
<serial*>:abc123
ERROR:onyo:fatal: pathspec 'an' did not match any files

❱ onyo tree
/tmp/tst
├── a
├── absolute
├── access
├── access.
├── allow
├── an
├── and
├── are
├── be
├── because
├── bits
├── by
├── can
├── cert
├── configured
├── containing
├── control
├── control.
├── controlled
├── created
├── directory
├── file
├── file.
├── files
├── for
├── Further
├── group
├── groups
│   └── dp
│       └── aqw
│           └── laptop_With
├── is
├── key
├── local
├── members
├── named
├── not
├── of
├── on
├── path,
├── permissions
├── permission._T490.abc123
├── pipe
├── pipe,
├── set
├── setting
├── socket
├── that
├── the
├── The
├── to
├── unix
├── used
├── user
├── via
├── when
└── with

3 directories, 54 files

```

TobiasKadelka:
I wrote a fix that will allow for spaces in all fields (and create just 1 file in the right location), but I think, at least when added as last/first character of the file name, I should still trim them.

aqw:
I disagree about the trimming. Shell already splits on white space, so file arguments will already be trimmed. /If/ Onyo receives a file with a leading or trailing space, it's because the user explicitly asked for it using quotes or escaping with `\`. In such cases (and as always), we should do what the user asks. They may be silly, but we should not make it impossible to be silly when explicit.

TobiasKadelka:
While the line `onyo new groups/dp/aqw` gets interpreted by CLI, this dialog:
```
<type>*:laptop
<make>*:With an absolute path, a unix local named pipe is used for control. The file is created with user and group that is configured and access bits are set to allow members of the group access. Further access can be controlled by setting permissions on the directory containing the control socket file. The key and cert files are not used when control is via the named pipe, because access control is via file and directory permission.
[...]

Issue #41, opened 2022-03-16
`onyo new` allows reserved characters to be used


aqw:
According to the README, `.` and `_` are reserved characters in some fields:

> The `type`, `make`, or `model` fields reserve the `_` and `.` characters.
>
> The `serial` field has no restrictions.

However, I am able to use those characters with `onyo new`:
```
❱ onyo new groups/dp/aqw 
<type>*:laptop_
<make>*:len_o_vo
<model*>:T.4_9.0
<serial*>:abc123
(vermin) aqw@strange in /tmp/tst on git:master
❱ onyo tree
/tmp/tst
└── groups
    └── dp
        └── aqw
            └── laptop__len_o_vo_T.4_9.0.abc123
```

Issue #34, opened 2022-03-16
`onyo new` errors (with a misleading message) on folders with spaces


aqw:
This manifests in many ways.

With a repository that looks like the following:
```
❱ onyo tree  
/tmp/tst
└── groups
    └── dp
```
If I run the following, it throws an error
```
❱ onyo new groups/dp/Alex\ Waite
ERROR:onyo:/tmp/tst/groups/dp/Alex Waite is no onyo repository.
```

It also creates empty folders at `groups/dp/Alex` and `Waite`:
```
❱ onyo tree
/tmp/tst
├── groups
│   └── dp
│       └── Alex
└── Waite
```

If I quote the argument rather than escaping the space, it still results in the same behavior:
```
❱ onyo new "groups/dp/Alex Waite" 
ERROR:onyo:/tmp/tst/groups/dp/Alex Waite is no onyo repository.
❱ onyo tree
/tmp/tst
├── groups
│   └── dp
│       └── Alex
└── Waite
```

If I create `groups/dp/Alex Waite/` using `mkdir`, it also chokes, albeit differently:
```
❱ mkdir groups/dp/Alex\ Waite
❱ onyo new groups/dp/Alex\ Waite/
<type>*:laptop
<make>*:lenovo
<model*>:T490
<serial*>:abc123
ERROR:onyo:touch: cannot touch 'Waite/laptop_lenovo_T490.abc123': No such file or directory
```
And it creates `groups/dp/Alex`:
```
❱ onyo tree
/tmp/tst
└── groups
    └── dp
        ├── Alex
        └── Alex Waite
```

PR #71, opened 2022-04-01
Fix/Update handling of keyboard input

TobiasKadelka:
This PR updates the handling from keyboard input at multiple places to fix problems from these issues:
#34, #41, #42

Spaces in directories/file names from CLI are possible and respected by all onyo functions, tests are updated for this, and reserved characters in `onyo new` are checked.


onyo/commands/new.py
```diff
@@ -14,16 +14,31 @@
...
 
 def build_commit_cmd(file, git_directory):
-    return ["git -C " + git_directory + " commit -m", "\'new " + file + "\'"]
+    return ["git -C " + git_directory + " commit -m", "\'new \"" + file + "\"\'"]
+
+
+def read_new_field(field_description):
+    # read word for field from keyboard
+    field = str(input(field_description))
+    # if field contains one of the reserved characters, read a new word.
```
aqw, line 27: If the field contains a reserved character, it should bark at the user, inform them which characters are reserved, and tell them to try again.

Nitpick: `word` should be normalized

aqw (changes requested):

onyo/commands/new.py
```diff
@@ -14,16 +14,32 @@
...
 
 def run_onyo_new(directory):
-    type_str = str(input('<type>*:'))
-    make_str = str(input('<make>*:'))
-    model_str = str(input('<model*>:'))
-    serial_str = str(input('<serial*>:'))
+    type_str = read_new_word('<type>*:')
+    make_str = read_new_word('<make>*:')
+    model_str = read_new_word('<model*>:')
+    serial_str = read_new_word('<serial*>:')
```
aqw, line 42: `read_new_word()` checks for `reserved_characters`. But according to the README, the `serial` is allowed to use reserved characters. Only `type`, `make`, and `model` are limited.

PR #75, opened 2022-04-08
add flag --non-interactive to suppress opening of editor

TobiasKadelka:


aqw:
It's uncommon to use `_` in flags. It's best to stay with `-` (i.e. `--non-interactive`).

How do you see this working with respect to the (as of yet unspec-ed) `-m <msg>, --message=<msg>` flag for commands which result in a git commit?

TobiasKadelka:
> How do you see this working with respect to the (as of yet unspec-ed) `-m <msg>, --message=<msg>` flag for commands which result in a git commit?

Like one would assume, without opening the editor, but instead using the passed message.

Without the -m flag (currently and only mode) it is using a default message, so if this is changed later to "opening the editor by default when commits are created", then there would be the need for a change, because in that case, with the non-interactive flag it should not open the editor, and just use the current generated message. But currently there is no conflict.

PR #78, opened 2022-04-21
Check files before running command(s), create one commit per command

TobiasKadelka:
The commands that can get multiple inputs check all of them now before running any operation, so onyo can inform about problems beforehand, and then after running all operations, onyo creates 1 commit (before, it did one commit per changed/created file/folder).

During this, some commands (like edit) got changed and do now accept multiple inputs. This is part of the PR, so the check_sources() functions do not have to be done first for one source, and then re-written for multiple ones.

### DIFF
--- a/onyo/commands/cat.py
+++ b/onyo/commands/cat.py
@@ -9,23 +9,26 @@ logging.basicConfig()
 logger = logging.getLogger('onyo')
 
 
-def build_cat_cmd(file):
-    this_asset = file
+def build_cat_cmd(files):
+    list_of_cat_commands = []
+    problem_str = ""
     onyo_repository_dir = os.environ.get('ONYO_REPOSITORY_DIR')
-    if not os.path.isfile(this_asset) and onyo_repository_dir is not None:
-        this_asset = os.path.join(onyo_repository_dir, this_asset)
-    if not os.path.isfile(this_asset):
-        logger.warning(file + " does not exist.")
-        return None
-    return "cat \"" + this_asset + "\""
+    for file in files:
+        if os.path.isfile(file):
+            list_of_cat_commands.append("cat \"" + file + "\"")
+        elif not os.path.isfile(file) and onyo_repository_dir is not None:
+            list_of_cat_commands.append("cat \"" + os.path.join(onyo_repository_dir, file) + "\"")
+        else:
+            problem_str = problem_str + "\n" + file + " does not exist."
+    if problem_str != "":
+        logger.warning(problem_str)
+    return list_of_cat_commands
 
 
 def cat(args):
-    for file in args.file:
-        # build command
-        cat_command = build_cat_cmd(file)
-
+    # check paths and build commands
+    list_of_cat_commands = build_cat_cmd(args.file)
+    for command in list_of_cat_commands:
         # run commands
-        if cat_command is not None:
-            output = run_cmd(cat_command)
-            print(output.strip())
+        output = run_cmd(command)
+        print(output.strip())
--- a/onyo/commands/edit.py
+++ b/onyo/commands/edit.py
@@ -18,39 +18,56 @@ logging.basicConfig()
 logger = logging.getLogger('onyo')
 
 
-def build_commit_cmd(file, git_directory):
-    return ["git -C \"" + git_directory + "\" commit -m", "\'edit " + file + "\'"]
+def build_commit_cmd(files, git_directory):
+    return ["git -C \"" + git_directory + "\" commit -m", "edit files\n" + "\n".join(files)]
+
+
+def check_sources(sources):
+    problem_str = ""
+    list_of_sources = []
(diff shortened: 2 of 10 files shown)


onyo/utils.py
```diff
@@ -68,10 +68,10 @@ def get_git_root(path):
 
 def get_full_filepath(git_directory, file):
     full_filepath = os.path.join(git_directory, file)
-    if not os.path.isfile(full_filepath):
+    if not os.path.exists(full_filepath):
```
aqw, line 71: Unsure what `exists()` checks for, but in shell it is very broad and includes symlinks, sockets, etc.

I'm not sure if Onyo should care, and if someone finds a way to abuse features for their use-case, more power to them. But I just wanted to register that this might be broader than just files/folders.

onyo/commands/tree.py
```diff
@@ -17,10 +19,48 @@ def build_tree_cmd(directory):
     return "tree \"" + directory + "\""
 
 
+def check_sources(sources):
```
aqw, line 22: This is relevant to all commands, and we already talked about this when we synced on video, but I just want to register my objections to the vague-ness of the name, and that it returns a modified copy of the sources, rather than a boolean.

onyo/commands/cat.py
```diff
@@ -9,23 +9,26 @@
...
     onyo_repository_dir = os.environ.get('ONYO_REPOSITORY_DIR')
-    if not os.path.isfile(this_asset) and onyo_repository_dir is not None:
-        this_asset = os.path.join(onyo_repository_dir, this_asset)
-    if not os.path.isfile(this_asset):
-        logger.warning(file + " does not exist.")
-        return None
-    return "cat \"" + this_asset + "\""
+    for file in files:
+        if os.path.isfile(file):
+            list_of_cat_commands.append("cat \"" + file + "\"")
```
aqw, line 18: Not a blocker for this PR, but I'm unaware of anything fancy that `cat` itself might be doing.

In general, we should stand on the shoulders-of-giants-written-in-C. Let git, etc al do the heavy lifting, and trust them. But it's not clear to me what `cat` might be giving us, and if it would be better to simply open and dump the contents of files in Python, and remove the overhead of calling a subprocess.

onyo/commands/edit.py
```diff
@@ -18,39 +18,56 @@
...
-    # check if file is in git
-    run_output = run_cmd("git -C \"" + git_directory + "\" ls-tree -r HEAD ")
-    if git_filepath not in run_output:
-        logger.error(git_filepath + " is not in onyo.")
+def check_sources(sources):
+    problem_str = ""
+    list_of_sources = []
+    if isinstance(sources, str):
+        sources = ["".join(sources)]
+    git_directory = get_git_root(sources[0])
```
aqw, line 30: Registering the outcome of our discussion: commands should not be involved in determining the repo root. This should be done at one point (invocation of onyo) and the result passed down to all commands.

onyo/commands/edit.py
```diff
@@ -18,39 +18,56 @@
...
+            sys.exit(1)
+
+        # change file
+        if not args.non_interactive:
+            edit_file(source)
+
+        # check if changes happened and add them
+        repo = Repo(git_directory)
+        changed_files = [item.a_path for item in repo.index.diff(None)]
+        if len(changed_files) != 0:
```
aqw, line 67: Is it possible to have a negative length (error state or something?) If so, the check should be `> 0`.

aqw (approved):
I did two quick skims. None of what I've pointed out here or in our conversation need to be addressed in this specific PR.

onyo/utils.py
TobiasKadelka: While this is not the last word on the topic, my thought was to allow these and all other types of files, as long as they "exist". This might introduce weird things and behaviour, but as long as I don`t see anything breaking for sure (which seems not to be the case when building a path) I prefered to allow instead of block stuff with onyo.

As a note: when it definitely does matter (like for mkdir, mv, etc) the onyo commands check more specificly, what the path contains.

onyo/commands/edit.py
TobiasKadelka: changed_files should be a normal list, so I don't think that could happen.
But if it would happen, I am equally not sure, if we wanted to "go on and add the changes".
It might be worth trying to break this part, and if we find something adjust the behaviour, but I think I will leave it like that until we do harder tests to break all layers of onyo.

onyo/utils.py
aqw: Sounds good to me. :-)

onyo/commands/edit.py
aqw: Seems like `len()` is ok, and will only return > 0.

https://stackoverflow.com/questions/42521449/how-does-python-ensure-the-return-value-of-len-is-an-integer-when-len-is-cal

PR #79, opened 2022-04-22
Clean and Update the Readme

TobiasKadelka:
After discussions, updates and changes in onyos behaviour, the readme needed to be updated, and mistakes in it were corrected too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,11 +196,12 @@ laptop asset should be converted to a folder, and the license moved into it.
 
   - `--quiet`: Silence the output (requires the `--yes` flag)
   - `--yes`: Respond "yes" to the prompt and run non-interactively
-- `onyo edit asset`:
+- `onyo edit asset...`:
 
-  Open the `asset` file using the default text editor specified by the
+  Open the `asset` file(s) using the default text editor specified by the
   environment variable `EDITOR` (`nano` and then `vi` are used as fallbacks).
 
+  When multiple asset files are given, Onyo will open them in sequence.
   - `--non-interactive`: Suppress opening of editor
 
   TODO: Describe validation
--- a/README.md
+++ b/README.md
@@ -320,9 +320,9 @@ TODO:
 
 ## Examples
 
-**Inventory a new asset; add it to the warehouse**:
+**Inventory a new asset; add it to the shelf**:
 ```
-onyo new warehouse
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -145,8 +145,9 @@ laptop asset should be converted to a folder, and the license moved into it.
   created, and template files generated and committed.
 
   The current working directory will be initialized if neither `directory` nor
-  the `ONYO_REPOSITORY_DIR` environment variable are specified. If both are set,
-  Onyo will throw an error.
+  the `ONYO_REPOSITORY_DIR` environment variable are specified. If the
+  environment variable is set, but a directory variable is given, Onyo will use
+  the given directory.
```
aqw, line 150: I disagree with this change in behavior (what if we ad a `-C` option like `git`; how should that behave?). But, if you do plan to make this change, the wording should be made more explicit:

> If the both are set, Onyo will print a warning, but still initialize the supplied directory.

README.md
```diff
@@ -197,11 +197,12 @@ laptop asset should be converted to a folder, and the license moved into it.
...
   - `--quiet`: Silence the output (requires the `--yes` flag)
   - `--yes`: Respond "yes" to the prompt and run non-interactively
-- `onyo edit asset`:
+- `onyo edit asset...`:
 
-  Open the `asset` file using the default text editor specified by the
+  Open the `asset` file(s) using the default text editor specified by the
   environment variable `EDITOR` (`nano` and then `vi` are used as fallbacks).
 
+  When multiple asset files are given, Onyo will open them one after another.
```
aqw, line 205: The current wording is fine. Another way of saying it is: "Onyo will open them in sequence."

aqw (changes requested):
The only controversial thing here for me is the change in behavior for `onyo init`. I'm not a fan of guessing what the user meant when they provide two conflicting pieces of information. I think that's a rabbit hole that never ends, and creates ambiguity rather than decreases it. The ambiguity will increases even further if/when we allow the user to provide information in other ways (such as `-C`).

Outside of that; this all looks good to me. :-)

README.md
TobiasKadelka: I reverted this change. The code will be updated later. I will also create an issue for -C, where we can discuss the implementation and the behaviour of it.

PR #82, opened 2022-04-29
Cleanup code

TobiasKadelka:
This PR contains a lot of little changes, cleanups and movings of code, that were put of "for later", in issues and video meetings.

`onyo edit` checks if the editor set by $EDITOR exists, if not, if nano exists and newly, uses now vi as a fallback, in case nano is not found, too. Using vi as a fallback was in the readme for a long time and now is finally part of the code.

The other changes are updates of output (like errors), and renaming of variables and functions in the code.

### DIFF
--- a/onyo/utils.py
+++ b/onyo/utils.py
@@ -79,8 +79,23 @@ def get_full_filepath(git_directory, file):
 
 def get_editor():
     editor = os.environ.get('EDITOR')
-    if editor is None:
+    if not editor:
+        logger.info("$EDITOR is not set.")
+    elif editor and run_cmd("which " + editor).rstrip("\n") == "":
+        logger.warning(editor + " could not be found.")
+    else:
+        return editor
+    # try using vi/nano as editor
+    if run_cmd("which nano").rstrip("\n") != "":
+        logger.info("nano is used as editor.")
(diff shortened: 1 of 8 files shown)


onyo/utils.py
```diff
@@ -79,8 +79,21 @@ def get_full_filepath(git_directory, file):
 
 def get_editor():
     editor = os.environ.get('EDITOR')
-    if editor is None:
-        editor = 'nano'
+    if (editor is None) or (editor == "") or (run_cmd("which " + editor).rstrip("\n") == ""):
```
aqw, line 82: Two comments here:

This check for two different things:
1) if no EDITOR is specified
2) if the specified editor cannot be found in the path

The first can be simplified to use truthiness:
```
if not editor:
```

The second check should be its own if-check (before this one) and throw a warning. Then proceeding to the fallbacks makes sense.

aqw (changes requested):
Overall LGTM. I left one comment about the editor detection/fallback. Otherwise: all good. :-)

onyo/utils.py
TobiasKadelka: I included both changes. Thanks!

PR #109, opened 2022-08-08
set: add command to onyo

### LINKED ISSUE #15: implement `onyo set`

### LINKED ISSUE #108: default value for `onyo set --depth NUM`
Currently I have choosen 0 as depth, so by default one can add just real assets, so `onyo set RAM=10 shelf/` would be the same as `onyo set --depth 0 RAM=10 shelf/`, and thereby it would add no assets. However, to add the assets directly in that folder, one can do either `onyo set --depth 1 RAM=10 shelf/` or `onyo set RAM=10 shelf/*` so the question of this issue is not about functionality. Instead I want to make sure, that the default behaviour is intuitive.

So, alternatively, we might want to have a default of 1, so that when the user runs `onyo set RAM=10 shelf/` the assets directly in shelf will be added?

TobiasKadelka:
This enables to set one or multiple values for one or multiple assets.
It changes existing values, or adds new ones.

The flags --rename and --filter are still missing and must be
added later. Dictionary keys and lists as values are not implemented.

Close #108 because it implements the changes discussed.
Close #15 because the command is implemented.

Issues about the missing flags and functionalities will be added,
together with a cleanup of the readme.

### DIFF
--- a/onyo/commands/set.py
+++ b/onyo/commands/set.py
@@ -1,0 +1,197 @@
+#!/usr/bin/env python3
+
+import logging
+import os
+import sys
+import yaml
+import glob
+from ruamel.yaml import YAML
+
+from git import Repo
+
+from onyo.utils import (
+    run_cmd,
+    build_git_add_cmd,
+    validate_rule_for_file
+)
+
+logging.basicConfig()
+logger = logging.getLogger('onyo')
+
+
+def build_commit_cmd(files_to_change, onyo_root):
+    return ["git -C " + onyo_root + " commit -m", "set values.\n\n" +
+            "\n".join(files_to_change)]
+
+
+def read_asset(file, onyo_root):
+    yaml = YAML(typ='safe')
+    asset = {}
+    with open(os.path.join(onyo_root, file), "r") as stream:
+        try:
+            asset = yaml.load(stream)
+        except yaml.YAMLError as e:
+            print(e)
+    if asset is None:
+        return {}
+    return asset
+
+
+def write_asset(data, file, onyo_root):
+    with open(os.path.join(onyo_root, file), "w") as stream:
+        yaml.dump(data, stream)
+
+
+def set_value(file, git_filepath, keys, onyo_root):
+    asset = read_asset(git_filepath, onyo_root)
+    for key in keys.items():
+        try:
+            asset[key[0]] = key[1]
+        # when a YAML asset just has a line with a key like "RAM" without value:
+        except TypeError as e:
+            return "Can't update asset " + git_filepath + ": " + str(e)
+    write_asset(asset, git_filepath, onyo_root)
+    return ""
+
+
+def add_assets_from_directory(directory, asset_list, depth, onyo_root):
+    for path in os.listdir(os.path.join(onyo_root, directory)):
+        asset = os.path.join(os.path.join(onyo_root, directory), path)
+        if path in [".git", ".onyo", ".anchor"]:
+            continue
+        if os.path.isfile(asset):
+            asset_list.append(asset)
+        elif os.path.isdir(asset):
+            if depth == 0:
+                return asset_list
+            else:
+                if depth > 0:
+                    depth = depth - 1
+                asset_list = add_assets_from_directory(os.path.join(directory, path), asset_list, depth, onyo_root)
+    return asset_list
+
+
+def diff_changes(file_list, keys, onyo_root):
+    output_str = ""
+    asset = []
+    ru_yaml = YAML(typ='safe')
+    for file in file_list:
+        with open(os.path.join(onyo_root, file), "r") as stream:
+            try:
+                asset = ru_yaml.load(stream)
+                if not asset:
+                    asset = []
+            except yaml.YAMLError as e:
+                print(e)
+        asset_changed = ""
+        for key in keys:
+            # if the value exists and is the same as before, do nothing:
+            if key in asset and asset[key] == keys[key]:
+                continue
+            # if existing value gets updated, show the removal of the old value:
+            if key in asset:
+                asset_changed = asset_changed + "\n-\t" + key + ": " + str(asset[key])
+            # display new value:
+            asset_changed = asset_changed + "\n+\t" + key + ": " + str(keys[key])
+        if asset_changed:
+            output_str = output_str + "\n" + os.path.relpath(file, onyo_root) + asset_changed
+    return output_str
+
+
+def simulate_validation_after_change(file, rules_file, keys, onyo_root):
+    problem_str = ""
+    temp_file = os.path.join(onyo_root, os.path.join(".onyo/temp/", os.path.basename(file)))
+    run_cmd("cp \"" + file + "\" \"" + temp_file + "\"")
+    problem_str = problem_str + set_value(temp_file, temp_file, keys, onyo_root)
+    for path_of_rule in rules_file:
+        if os.path.join(onyo_root, file) in glob.glob(os.path.join(onyo_root, path_of_rule), recursive=True):
+            for rule in rules_file[path_of_rule]:
+                problem_str = problem_str + validate_rule_for_file(temp_file, rule, path_of_rule, file, onyo_root)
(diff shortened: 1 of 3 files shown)


onyo/commands/set.py
```diff
@@ -0,0 +1,165 @@
...
+    with open(os.path.join(onyo_root, file), "w") as stream:
+        yaml.dump(data, stream)
+
+
+def set_value(file, git_filepath, keys, onyo_root):
+    asset = read_asset(git_filepath, onyo_root)
+    for key in keys.items():
+        try:
+            asset[key[0]] = key[1]
+        # when in YAML asset just a line with key like "RAM" without value:
```
aqw, line 50: Missing "is" (I think)

onyo/commands/set.py
```diff
@@ -0,0 +1,165 @@
...
+        except TypeError as e:
+            print(e)
+    write_asset(asset, git_filepath, onyo_root)
+
+
+def add_assets_from_directory(directory, asset_list, depth, onyo_root):
+    for path in os.listdir(os.path.join(onyo_root, directory)):
+        asset = os.path.join(os.path.join(onyo_root, directory), path)
+        if path == ".git" or path == ".onyo" or path == ".anchor":
```
aqw, line 59: Using a list reads more cleanly for me:

```
if path in ['.git', '.onyo', '.anchor']:
```

onyo/commands/set.py
```diff
@@ -0,0 +1,165 @@
...
+            continue
+        if os.path.isfile(asset):
+            asset_list.append(asset)
+        elif os.path.isdir(asset):
+            # if a depth is given:
+            if depth > 0:
```
aqw, line 65: Purely style, but I'd write this a bit differently:

```
if depth == 0:
    return asset_list
else:
    if depth > 0:
        depth = depth - 1
    asset_list = add_assets_from_directory(os.path.join(directory, path), asset_list, depth, onyo_root)

File: onyo/main.py [hunk 4]
```diff
@@ -232,6 +260,60 @@ def parse_args():
...
+        type=int,
+        required=False,
+        default=-1,
+        help='Limit the maximum number of levels of directories to search beneath source'
+    )
+    cmd_set.add_argument(
+        'keys',
+        action=StoreDictKeyPair,
+        metavar="keys",
+        help='Key value pairs to set in sources'
```
aqw, line 309: I'm not a fan of the word "source" here. Do you mean asset files?

onyo/main.py
```diff
@@ -232,6 +260,60 @@ def parse_args():
...
+        default=False,
+        action='store_true',
+        help='Silence output (requires the --yes flag)'
+    )
+    cmd_set.add_argument(
+        '-y', '--yes',
+        required=False,
+        default=False,
+        action='store_true',
+        help='Respond "yes" to the prompt'
```
aqw, line 288: "To any prompts"

onyo/main.py
```diff
@@ -232,6 +260,60 @@ def parse_args():
...
+    )
+    cmd_set.add_argument(
+        '-R', '--recursive',
+        required=False,
+        default=False,
+        action='store_true',
+        help='Respond "yes" to the prompt'
```
aqw, line 295: Copy/paste mistake

onyo/main.py
```diff
@@ -232,6 +260,60 @@ def parse_args():
...
+    )
+    cmd_set.add_argument(
+        '-d', '--depth',
+        metavar='depth',
+        type=int,
+        required=False,
+        default=-1,
+        help='Limit the maximum number of levels of directories to search beneath source'
```
aqw, line 303: Source -> "target directory"

onyo/main.py
```diff
@@ -12,6 +12,34 @@
...
+        super(StoreDictKeyPair, self).__init__(option_strings, dest, nargs=nargs, **kwargs)
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        my_dict = {}
+        for kv in values.split(","):
+            if "=" not in kv:
+                logger.error("\nNo \"=\" in key/value pair \"" + str(kv) + "\"")
+                sys.exit(1)
+            if kv.count("=") > 1:
+                logger.error("\n\"" + str(kv) + "\" contains to many \"=\". To separate multiple key/value pairs use a space.")
```
aqw, line 28: to -> too

aqw (commented):
I didn't get to review everything, nor run the actual code. But I found a few bits while looking through.

onyo/commands/set.py
TobiasKadelka: Updated to "when a YAML asset just has a line with a key like "RAM" without value:"

onyo/main.py
TobiasKadelka: Urrgh.

TobiasKadelka:
Thanks for the feedback, I added the changes to the PR and squashed them into the commit.

onyo/main.py
```diff
@@ -232,6 +260,60 @@ def parse_args():
...
+        default=False,
+        action='store_true',
+        help='Respond "yes" to any prompts'
+    )
+    cmd_set.add_argument(
+        '-R', '--recursive',
+        required=False,
+        default=False,
+        action='store_true',
+        help='When a directory is given, set values recursively for all assets in it'
```
aqw, line 295: The way this is written is odd to me. If a directory is not provided, but the `-R` is passed, then onyo will ignore it and operate non-recursively?

In general, onoy assumes the top-level dir if no argument is provided, and I think we should be consistent here. If the user says recursive, then onyo will operate recursively.

The following wording is closer to what I'm thinking:

> Set values for all (matching) assets under each directory, recursively,

It's still a bit vague on the "even when no directory is provided" topic, but it feels more clear to me.

onyo/main.py
```diff
@@ -232,6 +260,60 @@ def parse_args():
...
+    )
+    cmd_set.add_argument(
+        '-d', '--depth',
+        metavar='depth',
+        type=int,
+        required=False,
+        default=-1,
+        help='Limit the maximum number of levels of directories to search beneath target directory'
```
aqw, line 303: This should reference the `--recursive` flag in some way. Something like:

> Descend at most "n" levels of directories below the starting-point(s). Used only with `--recursive`.

onyo/main.py
```diff
@@ -232,6 +260,60 @@ def parse_args():
...
+    )
+    cmd_set.add_argument(
+        'keys',
+        action=StoreDictKeyPair,
+        metavar="keys",
+        help='Key value pairs to set in asset files'
```
aqw, line 309: An example would help users here, and also mentioning comma-separation.

> Key value pairs to set in asset files. Multiple pairs can separated by commas (e.g. key=value,key2=value2).

onyo/main.py
TobiasKadelka: > The way this is written is odd to me. If a directory is not provided, but the -R is passed, then onyo will ignore it and operate non-recursively?

Yes, when an asset is also given, for example  `onyo set --recursive USB=2 shelf/asset user/`, it will ignore the `--recursive` flag for `shelf/asset` and just run the command recursive for `user/`

TobiasKadelka: > In general, onoy assumes the top-level dir if no argument is provided, and I think we should be consistent here. If the user says recursive, then onyo will operate recursively.

Does it assume the top level, or the current working dir?

TobiasKadelka: "Set values recursively for assets in directories"

aqw: > Yes, when an asset is also given, for example onyo set --recursive USB=2 shelf/asset user/, it will ignore the --recursive flag for shelf/asset and just run the command recursive for user/

Well, by definition, it cannot operate recursively on a file, so this makes sense.

But to be clear, if two directories are passed to onyo, both will be operated on recursively. Yes?

But the point I was trying to make is that the wording made it sound like that if we weren't providing any arguments (thus implying the entire repo), then the --recursive flag would no-op. Which I disagree with (and I don't think you intended to say, but it reads that way).

PR #115, opened 2022-08-11
WIP: Round up

### LINKED ISSUE #104: Validation/Templates: which files to provide with onyo default installation?
Currently I am providing with the repository, the installation of onyo, and `onyo init` some templates and validation files.

These are just place holder as a proof of concept and to demonstrate the functionalities, I do not think that they otherwise provide any real value to the user.

We should discuss which template files and which validation rules we want to give with onyo per default with every installation, or if we want to leave it like it is right now and let the user decide/overwrite everything manually.

### LINKED ISSUE #111: Design and Implement faux serial numbers
The information in the readme about faux serials is the following:
```
Assets for which the identity is not important receive a unique
faux-serial to prevent filename conflicts. These serials are
prepended with the word faux.
```

Currently I plan to just check if the serial field value is  `== "faux"`, and in that case call a function that generates a serial value which provides a *unique file name*, e.g. `laptop_apple_macbookpro.faux5252`. (I know we want the laptop serials to not be faux because the identity matters, it is just an example).

We talked about using something like https://docs.python.org/3/library/uuid.html. These generate fairly long names, so we should decide if we want:

- using them completely
- using them, but cut to certain length
- just generate an own unique but random number

How long should faux serial numbers be?

TobiasKadelka:
This PR contains a lot of small pieces that were missing, cleanups, and bugfixes:
- Most importantly: cleanup the readme to reflect the actual/current state of onyo
  - remove not implemented parts
  - update and normalize names, examples
  - add new flags
- add faux serial numbers
  - Close #111 
- removes test validation and template files to prepare onyo for real usage
  - comment out validation.yaml rules, to give examples without giving default rules
  - rename templates, e.g laptop to laptop.example
  - Close #104
- allow reserved characters for serial field (this was always in readme, but not implemented)
- Bug fixes:
  - fix a bug regarding empty file validation.yaml (created non-iterable noneType object)
  - fix bug where `onyo mv --force --rename` allowed to create a duplicate asset name
  - fix error message for `onyo cat <dir>` on existing <dir> from `<dir> does not exist` to `onyo cat expects files, but <dir> is folder`


onyo/utils.py
```diff
@@ -141,6 +140,8 @@ def validate_file(file, original_file, onyo_root):
     with open(os.path.join(onyo_root, ".onyo/validation/validation.yaml"), "r") as stream:
         try:
             rules_file = ru_yaml.load(stream)
+            if rules_file is None:
```
aqw, line 143: I think truthiness might be better here:

```
if not rules_file:
    rules_file = []

aqw (approved):
This all looks great to me.

In the commit message for the README cleanup, it would be nice to have links to the issues you created to implement the features that are being removed from the README. But it's not essential.

onyo/utils.py
TobiasKadelka: Good point. I will update it, and the readme commit too.

TobiasKadelka:
I updated the PR so it contains the feedback, and added additional bug-fixes (and edited them to the description above)

PR #118, opened 2022-08-26
onyo: add "onyo -C <dir>", remove $ONYO_REPOSITORY_DIR

### LINKED ISSUE #62: define behavior when `ONYO_REPOSITORY_DIR` is set and cwd is in an onyo repo
Currently, it is undefined what Onyo should do when `ONYO_REPOSITORY_DIR` is set and the current-working-directory is in an onyo repository.

The most straightforward solution I see is the following:

- if `ONYO_REPOSITORY_DIR` if set, always use it
  - if in an onyo repo warn/inform user that onyo is using `ONYO_REPOSITORY_DIR`
- if not set, use current-working-directory
- if not set and not in onyo repo, error

There are other approaches, though most seem more complicated on the surface.

In general, we should aim for a solution that is easy to explain and intuit, rather than a deep algorithmic tree.

### LINKED ISSUE #80: Readme: how to work with multiple onyo directories
The current readme is written as if it operates on the assumption, that there will be only one onyo directory.

There are many ways we discussed how to work with multiple onyo directories (e.g. aliases), and we should add a section to the readme, that explains how to do this.

### LINKED ISSUE #81: onyo -C </path/to/run/> <command>
We discussed multiple times, if we want to have a -C flag.
The implementation would be fairly straight forward, so I want to first discuss the behaviour and consequences of this. In general: It would enable to run an onyo command in an onyo repository B, while being in repository A.

1. Should this overwrite `$ONYO_REPOSITORY_DIR`, or throw an error if the var is set and -C is used?
2. How should this behave (assuming `$ONYO_REPOSITORY_DIR` is unset) :
`onyo -C /path/a/ git -C /path/b/ status`
I would argue, it should basically run as if double-redirected and show`git status` for /path/b/

TobiasKadelka:
This removes the environmental variable, instead it adds the option
`onyo -C <dir>` to explicitely specify the directory to run onyo from.

The changes mainly effect main.py, since it passes the variable
"onyo_root" to the other commands. onyo_root is now either cwd (like
before) or the `<dir>` from `onyo -C <dir>` (instead of the old var
$ONYO_REPOSITORY_DIR), so the commands should in general all
work as before.
In utils.py there where path checks for $ONYO_REPOSITORY_DIR
removed, too.

There were many test-cases checking that the environmental variable
works, these are removed. Instead, I added test cases, that run the
onyo commands from outside the onyo directory with `onyo -C <dir>`.
Further tests will be added soon, this should just cover the absolute
minimum requirement to check if the path with -C gets passed to the
commands correctly in general.

The readme got updated, the environmental variable is now not
mentioned anymore. Instead it mentions the -C option.

This implements #81 and removes the needs for issues about
$ONYO_REPOSITORY_DIR, so these can be closed:
close #81
close #62
close #80


aqw:
I don't have time to review this, but wanted to register a big +1 for this simplification.

aqw:
One comment: I think there should be a note somewhere in the README about using an alias in `.bashrc`/`.zshrc` to allow invoking onyo for a repo from anywhere. If that doesn't happen in this PR, then an issue should be created, and I'll write it later. :-)

PR #148, opened 2022-09-08
Move argparse to utils; update help text, metavars, and var names.

aqw:
This moves the argument parsing to `utils.py`, where it can be imported by the shell completion command. Perhaps it should be in its own file, but this seemed like an ok place for now.

This also includes numerous language tweaks I made to the help text and metavars for readability and consistency.

Lastly, the argument vars themselves were updated to match the metavars, which required some changes to the command files themselves, In general, I went for updating all variables, but sometimes I went for a less intrusive approach, as the entire file would change, and there was already other visible refactoring to be done. I left those for later.


aqw:
All tests pass. Merging.

Issue #126, opened 2022-08-29
DOC: setup Read The Docs


aqw:
It's becoming increasingly clear that the scope of documentation for Onyo would be better suited for something like Read The Docs rather than a single README.

See also #125 , #119 , #105 , #20 , #17 , and the templating/validation system.

loj:
Done! There's a link to the docs in the `README` via the RTD bagde.

PR #149, opened 2022-09-09
DOC: setup and populate docs

### LINKED ISSUE #125: DOC: divide README's contents in dedicated files and make it a summary with links
The README is now functioning as both a development guideline and a go-to for testers, but it's getting impractical to navigate.

I suggest to relocate its contents (as they are) in separate files:

- `DESIGN.md`: Just Files and Folders, Inventory Concepts, Asset Name Scheme, File Contents, Config Files, Template Files, Field Validation, Environment Variables
- `COMMANDS.md`: Commands
- `EXAMPLES.md` Examples, Example Templates, Example Validation

Then put in the README only the following:

- a oneliner description of Onyo
- how to install it (pip in virtualenv, for now, for who doesn't have a preferred way to compile from source)
- links to the above mentioned files
- how to contribute (which can be a file on its own but doesn't need to be just now).

One more needed file is a quickstart that is a mix of what the examples are for now but in a way that one can use it as a small demo.

loj:
This sets up basic documentation using sphinx that can eventually be linked up with Read the Docs.

* all previous documentation (overview, design concepts, examples, etc) contained in the `README` was moved over to the docs
* the previous command documentation in the `README` was moved to docstrings within each command and then linked to `argparse` via `description`.
* With the `sphinx-argparse` extension, the command line documentation is generated directly from `argparse`.

Once the documentation is ready to go live, it can be imported to [Read the Docs](https://docs.readthedocs.io/en/stable/tutorial/index.html#getting-started). This also should allow linking with CI to [trigger builds](https://docs.readthedocs.io/en/stable/tutorial/index.html#trigger-a-build-from-a-pull-request) directly with pull requests.

Closes #125 
Towards #126


aqw:
Thanks for the PR! I'm excited about this one. :-D

Looks like the tests are failing on linting. You can run the following:
```
flake8 $(find . -type f -name "*.py")
```

The lint errors seems to be in `conf.py`. Either fix them, or ignore the file entirely (by adding `# flake8: noqa` to the first line of the file).

Nits:
- `make.bat` is a windows file, and does not need to be included
- it would be nice to know where those example sphinx files came from. Explaining (or even just the command) in the commit message would be nice.
- The last commit fixes #127

loj:
Thanks for the feedback. :-) I  pushed an update that should resolve the failing tests as well as address your comments above.

aqw:
Test for Python 3.7 seem to fail. It seems to be related to `importlib.metadata`.

According to the [importlib.metadata docs](https://docs.python.org/3/library/importlib.metadata.html), it was added in Python 3.8.

I havn't looked closely. Perhaps it just needs to be installed (or a specific version). It is available on [PyPi](https://pypi.org/project/importlib-metadata/).

loj:
So it turns out that `flake8` and `sphinx` don't like each other. This provided an opportune moment to split the install dependencies for tests and docs out of the main package install. They are now in `extras_requires` for `tests` and for `docs` in `setup.py`. This allows for the dependencies of tests and dependencies for docs to be installed separately when desired instead of installing them by default with the main package.

Additionally I found that it is a good idea to upgrade pip before installing packages, and have added this step to the main test workflow.

Finally, I added a workflow to test build the docs.

aqw:
FINALLY! Thanks for pushing through this and getting the sphinx vs flake8 stuff figured out.

Merging. :-D

PR #167, opened 2022-09-16
DOC: add installation instructions and remove empty introduction page

loj:
Right now, I don't think we need an introduction page. Instead, I've added a page with installation instructions. Then eventually a quickstart or common work-flows page might make sense to add.

With this change, I think the docs are ready to go live with Read the Docs.


aqw:
Thanks!

Issue #193, opened 2022-09-25
`cat`: adds newline to end of output


loj:
This differs from the traditional behavior of `cat`.

aqw:
Fixed in 0cc6e4e

PR #194, opened 2022-09-25
RF: `cat`

### LINKED ISSUE #192: `cat`: errors do not return correct exit code
Currently errors output an exit code of `0` rather than `1`.

loj:
This adds tests for `cat` and also refactors the command to:
* improve readability
* fix error handling
* use python to print file content to stdout rather than calling cat via a sub process

Fixes #192, #193

### DIFF
--- a/onyo/commands/cat.py
+++ b/onyo/commands/cat.py
@@ -2,29 +2,54 @@
 
 import logging
 import os
+import sys
 
-from onyo.utils import run_cmd
 from onyo.commands.fsck import read_only_fsck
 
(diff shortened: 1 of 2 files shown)


aqw:
The failure is due to https://github.com/readthedocs/sphinx_rtd_theme/issues/1343

There's no sense in pinning a version to avoid this. Upstream will fix this soon enough, as it's affecting everyone.

The docs build correctly locally. Merging.

aqw:
Thanks!

PR #196, opened 2022-09-25
RF: `config`

aqw:
This PR:

- rewrites `onyo config`, basically turning it into a wrapper for `git config -f .onyo/config`
- adds tests
- adds documentation
- moves all configuration options to an "onyo" namespace, so that they can be set in git without fear of conflict
- the editor spawned by onyo is now configurable via `onyo.core.editor`
- adds a function `get_config_value()` to simplify and improve support in the codebase for using config values


aqw:
The docs build failure is the same as #194.

The docs build locally. Merging.

PR #215, opened 2022-10-04
RF: `mv`

aqw:
This PR refactors `mv` to bring it in line with how other commands (such as `rm`) operate.

The biggest behavioral change is that renaming files is no longer allowed. That will only be possible using `set`, making it the only command that can change keys (pseudo or otherwise).

Changes:
* addition of `--quiet`
* `--force` -> `--yes`
* retirement of `--rename`

The logic for `mv` is more complicated than I initially expected, as it can both move and rename files/directories, with different sanity checks required.

This adds docstrings and a boatload of tests to cover the many corner-cases.


aqw:
@TobiasKadelka Ready for a review. :-)

TobiasKadelka:
Looks good to me, I will merge :)

PR #216, opened 2022-10-05
RF/TST (pathlib, doc-strings, f-strings): `new`; `utils.edit_file()` and `utils.generate_faux_serial()`, `utils.get_list_of_assets()`

TobiasKadelka:
Changes in `onyo new`:
- uses pathlib.Path, shutil, and git.Repo
- updates imports
- adds doc-strings, f-strings and more whitespace
- adds minimal tests for `onyo new`
  - at different folder depths
  - white and without spaces in name
  - with and without faux serial

Changes in `utils`:
- modernize `edit_file()` (pathlib, doc-strings, f-strings)
- refactor `utils.get_list_of_assets()`
- move code `create_faux()` from `new` to `utils` and rename the function to `generate_faux_serial()`
- use `random.choices()` instead of `uuid1()` for the creation of faux serial numbers
- add also pathlib, doc-strings and f-strings
- test `generate_faux_serial()`
  - check for generation of different faux numbers with default length
  - test generation of different faux serial numbers with varying length
  - test for correct error code if faux serial numbers with a length outside of the allowed interval are requested

### DIFF
--- a/onyo/commands/new.py
+++ b/onyo/commands/new.py
@@ -2,12 +2,12 @@
 
 import logging
 import sys
-import uuid
 import shutil
 from git import Repo
 from pathlib import Path
 
 from onyo.utils import (
+    generate_faux_serial,
     get_config_value,
     get_list_of_assets,
     get_git_root,
@@ -74,20 +74,6 @@ def run_onyo_new(directory, template, non_interactive, onyo_root, repo):
     return filename
 
 
-def create_faux(onyo_root, list_of_assets, faux_length=8):
-    """
-    Create a new faux serial number for the serial field of an asset name.
-    Check for uniqueness of the faux serial in the repository.
-
-    Returns on success a unique faux serial number.
-    """
-    faux = "faux" + str(uuid.uuid1())[:faux_length]
-    for asset in list_of_assets:
-        if faux in asset[1]:
-            return create_faux(onyo_root, list_of_assets, faux_length)
-    return faux
-
-
 def create_filename(onyo_root, template):
     """
     Read fields required for an asset name, check the validity of the fields and
@@ -100,7 +86,11 @@ def create_filename(onyo_root, template):
     for field in ["type", "make", "model", "serial"]:
         word = read_new_word('<' + field + '>*:')
         if field == "serial" and word == "faux":
-            word = create_faux(onyo_root, get_list_of_assets(onyo_root))
+            try:
+                word = generate_faux_serial(onyo_root)
+            except ValueError as e:
+                print(e)
+                sys.exit(1)
         words.append(word)
     filename = words[0] + "_" + words[1] + "_" + words[2] + "." + words[3]
 
--- a/onyo/commands/new.py
+++ b/onyo/commands/new.py
@@ -1,13 +1,13 @@
 #!/usr/bin/env python3
 
 import logging
-import os
 import sys
 import uuid
+import shutil
+from git import Repo
+from pathlib import Path
 
 from onyo.utils import (
-    build_git_add_cmd,
-    run_cmd,
     get_config_value,
     get_list_of_assets,
     get_git_root,
@@ -21,44 +21,66 @@ logger = logging.getLogger('onyo')
 reserved_characters = ['_', '.']
 
 
-def build_commit_cmd(file, onyo_root):
-    return ["git -C \"" + onyo_root + "\" commit -m", "new asset.\n\n" + file]
-
-
 def read_new_word(word_description):
+    """
+    Read a new word via user input for the creation of a asset name.
+    Checks for the serial field if it contains reserved characters, and checks
+    for all fields that they are not empty. If a check is not successful, it
+    tries to read the same word again.
+
+    Returns the word on success.
+    """
     # read word for field from keyboard
-    word = str(input(word_description))
+    word = input(word_description)
+
     # if word contains reserved character, inform and read new word
     for char in reserved_characters:
         if word_description != "<serial>*:" and char in word:
-            logger.info(char + " is in list of reserved characters: " + ", ".join(reserved_characters))
+            logger.info(f"'{char}' is in list of reserved characters: {reserved_characters}")
             return read_new_word(word_description)
+
     # if enter pressed without input, read new word
     if len(word) == 0:
         return read_new_word(word_description)
     return word
 
 
-def run_onyo_new(directory, template, non_interactive, onyo_root):
-    # create an asset file, first without content
+def run_onyo_new(directory, template, non_interactive, onyo_root, repo):
+    """
+    Read a new asset name, check it for uniqueness and validity, create the
+    asset (from a template if one was given), edit the new asset (if not
+    suppressed with flag) and stage it with git.
+
+    Returns the newly created asset's filename on success.
+    """
+    # create an asset file
     filename = create_filename(onyo_root, template)
-    if os.path.exists(os.path.join(directory, filename)):
-        logger.error(os.path.join(directory, filename) + " asset already exists.")
+    filename = Path(directory, filename).resolve()
+    template = Path(template).resolve()
+
+    if filename.exists():
+        logger.error(f"{filename}' asset already exists.")
         sys.exit(1)
-    if template:
-        run_cmd("cp \"" + template + "\" \"" + os.path.join(os.path.join(onyo_root, directory), filename) + "\"")
+    if template.is_file():
+        shutil.copyfile(template, filename)
     else:
-        run_cmd(create_asset_file_cmd(directory, filename))
+        filename.touch()
+
     # open new file?
     if not non_interactive:
-        edit_file(os.path.join(directory, filename), onyo_root, onyo_new=True)
+        edit_file(filename, onyo_root, onyo_new=True)
     # add file to git
-    git_add_cmd = build_git_add_cmd(directory, filename)
-    run_cmd(git_add_cmd)
-    return os.path.join(directory, filename)
+    repo.git.add(filename)
+    return filename
 
 
 def create_faux(onyo_root, list_of_assets, faux_length=8):
+    """
+    Create a new faux serial number for the serial field of an asset name.
+    Check for uniqueness of the faux serial in the repository.
+
+    Returns on success a unique faux serial number.
+    """
     faux = "faux" + str(uuid.uuid1())[:faux_length]
     for asset in list_of_assets:
         if faux in asset[1]:
@@ -67,6 +89,12 @@ def create_faux(onyo_root, list_of_assets, faux_length=8):
 
 
 def create_filename(onyo_root, template):
+    """
+    Read fields required for an asset name, check the validity of the fields and
+    uniqueness of the build-together asset name.
+
+    Returns the new asset name on success.
+    """
     words = []
     # build new name, read these words
     for field in ["type", "make", "model", "serial"]:
@@ -75,32 +103,36 @@ def create_filename(onyo_root, template):
             word = create_faux(onyo_root, get_list_of_assets(onyo_root))
         words.append(word)
     filename = words[0] + "_" + words[1] + "_" + words[2] + "." + words[3]
-    # check if the new filename is actually unique in onyo repository
+
+    # check if the new asset name is actually unique in onyo repository
     assets = get_list_of_assets(onyo_root)
     for asset in assets:
         if filename == asset[1]:
-            logger.info(filename + " exists already in " + asset[0] + "\nCreate a new filename:")
+            logger.info(f"{filename} exists already in {asset[0]}\nCreate a new filename:")
             return create_filename(onyo_root, template)
     return filename
 
 
-def create_asset_file_cmd(directory, filename):
-    return "touch \"" + os.path.join(directory, filename) + "\""
+def sanitize_paths(directory, template, onyo_root):
+    """
+    Check and normalize the directory and template for the creation of a new
+    asset.
 
-
-def prepare_arguments(directory, template, onyo_root):
-    directory = os.path.join(onyo_root, directory)
+    Returns the absolute path of the directory and of the template on success.
+    """
+    directory = Path(onyo_root, directory).resolve()
     if not template:
         template = get_config_value('onyo.new.template', onyo_root)
 
-    template = os.path.join(get_git_root(onyo_root), os.path.join(".onyo/templates", template))
+    template = Path(get_git_root(onyo_root), ".onyo/templates", template).resolve()
 
     problem_str = ""
-    if not os.path.isfile(template):
-        problem_str = problem_str + "\nTemplate file " + os.path.join(get_git_root(onyo_root), os.path.join(".onyo/templates", template)) + " does not exist."
-    if not os.path.isdir(directory):
-        problem_str = problem_str + "\n" + directory + " is not a directory."
-    if problem_str != "":
+    if not template.is_file():
+        problem_str += f"\nTemplate file {template} does not exist."
+    if not directory.is_dir():
+        problem_str += f"\n{directory} is not a directory."
+
+    if problem_str:
         logger.error(problem_str)
         sys.exit(1)
     return [directory, template]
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -133,6 +133,55 @@ def test_get_editor_precedence():
...
+    faux.
+    """
+    onyo_root = './'
+    faux = []
+    number_of_faux = range(0, 100)
+
+    # technically, creating multiple faux serial numbers without actually
+    # creating assets might lead to duplicates. In practice, this should not
+    # happen if they are sufficiently long and random
+    for i in number_of_faux:
```
aqw, line 149: This loop can be consolidated to:
```
faux = [ utils.create_faux(onyo_root) for x in range(0,100) ]
assert len(faux) == len(set(faux))
```
And will also execute faster.

tests/test_utils.py
```diff
@@ -133,6 +133,55 @@ def test_get_editor_precedence():
...
+        assert new_faux not in faux
+        faux.append(new_faux)
+
+    # check that the correct number of unique faux serials was created
+    assert len(faux) == len(number_of_faux)
+
+
+def test_create_faux_different_length():
+    """
+    Create many faux with different length. Because of the different length,
```
aqw, line 161: "Faux" is not a noun. Many faux(fake) serials are generated.

Length -> lengths.

Is there a max length? If so, that should be documented and checked for in the `create_faux()` function.

tests/test_utils.py
```diff
@@ -133,6 +133,55 @@ def test_get_editor_precedence():
...
+    onyo_root = './'
+    faux = []
+    # this is somewhat limited by the default length of uuid.uuid1(), but 35
+    # should also be sufficiently to test the function.
+    number_of_faux = range(0, 35)
+
+    # technically, creating multiple faux serial numbers without actually
+    # creating assets might lead to duplicates. In practice, this should not
+    # happen if they are sufficiently long and random
+    for i in number_of_faux:
```
aqw, line 173: Same comments here as for the previous for-loop.

aqw (changes requested):

aqw (changes requested):

onyo/commands/new.py
```diff
@@ -84,23 +113,26 @@ def create_filename(onyo_root, template):
...
 
-    template = os.path.join(get_git_root(onyo_root), os.path.join(".onyo/templates", template))
+    template = Path(get_git_root(onyo_root), ".onyo/templates", template).resolve()
 
     problem_str = ""
-    if not os.path.isfile(template):
-        problem_str = problem_str + "\nTemplate file " + os.path.join(get_git_root(onyo_root), os.path.join(".onyo/templates", template)) + " does not exist."
-    if not os.path.isdir(directory):
+    if not template.is_file():
+        problem_str = problem_str + "\nTemplate file " + template + " does not exist."
```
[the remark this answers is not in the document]
TobiasKadelka: I will. 

The idea of not erroring immediately is that onyo can print out a list of problems that can be solved at once before calling onyo again, instead of re-running the command multiple times and getting everytime a new error. That said, in some of the instances it is probably not necessary.

onyo/commands/new.py
```diff
@@ -67,6 +89,12 @@ def create_faux(onyo_root, list_of_assets, faux_length=8):
 
 
 def create_filename(onyo_root, template):
```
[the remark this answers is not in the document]
TobiasKadelka: That is true. 
There was already a discussion about using e.g. the git index to get a list of files and changing the function. I will for  not update this now, instead I created an issue for this and work on it in another PR for the refactoring of some of `util`s functions.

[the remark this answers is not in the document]
TobiasKadelka: #218 describes this.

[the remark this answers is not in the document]
aqw: At bare minimum, this can be trivially reduced from 5x to 1x.

The full problem does not need to be solved immediately in order to drastically reduce the amount of work done here.

onyo/commands/new.py
```diff
@@ -21,44 +21,66 @@
...
+        edit_file(filename, onyo_root, onyo_new=True)
     # add file to git
-    git_add_cmd = build_git_add_cmd(directory, filename)
-    run_cmd(git_add_cmd)
-    return os.path.join(directory, filename)
+    repo.git.add(filename)
+    return filename
 
 
 def create_faux(onyo_root, list_of_assets, faux_length=8):
```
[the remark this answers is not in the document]
TobiasKadelka: I think this is a comment about a commit in between. The comment is correct, and the function got moved from`new.py` to `utils.py`, and renamed to `generate_faux_serial()`, where it uses the variable to get a list of existing assets.

onyo/commands/new.py
```diff
@@ -21,44 +21,66 @@
...
+    """
+    # create an asset file
     filename = create_filename(onyo_root, template)
-    if os.path.exists(os.path.join(directory, filename)):
-        logger.error(os.path.join(directory, filename) + " asset already exists.")
+    filename = Path(directory, filename).resolve()
+    template = Path(template).resolve()
+
+    if filename.exists():
+        logger.error(filename + " asset already exists.")
```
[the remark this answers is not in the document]
aqw: This is a new line that you're adding. We should not make existing problems worse.

onyo/commands/new.py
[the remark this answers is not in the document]
aqw: > I think this is a comment about a commit in between. 

Yes, that was the case.

Marking as resolved.

onyo/commands/new.py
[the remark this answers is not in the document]
aqw: There are ways to do that while not assigning everything to one generic variable.

onyo/utils.py
```diff
@@ -74,6 +75,28 @@ def get_editor(onyo_root):
...
+    Generate a unique faux serial number. The faux serial is verified to be
+    unique in the repository before it is returned.
+    The requested length of faux serial numbers has to be between 1 and 37.
+
+    Returns on success a unique faux serial number.
+    """
+    # check that the requested faux serial number has a valid length
+    if faux_length < 1 or faux_length > 37:
+        logger.error("Length of faux serial numbers must be between 1 and 37")
+        sys.exit(1)
```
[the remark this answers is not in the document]
TobiasKadelka: I decided to raise a valueError as the most appropriate thing to do.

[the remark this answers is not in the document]
TobiasKadelka: The tests for `utils.generate_faux_serial()` and the function in `onyo new` which can call `generate_faux_serial()` are updated to reflect the new behaviour.

tests/test_utils.py
```diff
@@ -133,6 +134,49 @@ def test_get_editor_precedence():
...
+    assert len(faux) == len(set(faux))
+    assert len(faux) > 0
+
+
+def test_generate_faux_serials_different_lengths():
+    """
+    Create many faux serial numbers with different lengths. Because of the
+    different lengths, they should all be different.
+    """
+    onyo_root = './'
```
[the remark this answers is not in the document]
TobiasKadelka: That is true. Should all functions in `utils` that get `onyo_root` test if it is actually the onyo repository root directory? I think it might be better to check this beforehand in the onyo main or onyo commands that call sub-functions from utils.
Otherwise there are a lot of tests that would need to be run for many of the utils function calls.

onyo/commands/new.py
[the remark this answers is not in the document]
TobiasKadelka: I was not exactly sure how a trivial partial solution would look like, instead I decided to do this with the git index.

The newest commit is a refactoring of the function and it would be nice if you could review the changes in `get_list_of_assets()` that I made there.

onyo/utils.py
```diff
@@ -261,13 +300,18 @@ def get_config_value(name, onyo_root):
 
 
 def get_list_of_assets(repo_path):
```
aqw, line 302: This is what I am using as part of my fsck work:

```
def get_list_of_assets(repo_path):                                               
    assets = {x for x in repo.git.ls_files('--full-name').split('\n')            
              if not is_protected_path(x)}

    return assets
```

onyo/commands/new.py
```diff
@@ -21,86 +21,108 @@
...
-    if template:
-        run_cmd("cp \"" + template + "\" \"" + os.path.join(os.path.join(onyo_root, directory), filename) + "\"")
+    if template.is_file():
+        shutil.copyfile(template, filename)
     else:
-        run_cmd(create_asset_file_cmd(directory, filename))
+        filename.touch()
+
     # open new file?
     if not non_interactive:
```
aqw, line 70: No need to change it for this PR, but double negatives are hard to read. Take a look at what I did to `onyo history` and how it uses `interactive` as the destvar.

onyo/utils.py
```diff
@@ -165,59 +191,72 @@ def check_float(value):
...
                 break
             elif edit_temp == 'n':
-                run_cmd("cp \"" + file + "\" \"" + temp_file + "\"")
+                shutil.copyfile(file, temp_file)
                 break
     further_editing = 'y'
     while further_editing == 'y':
         # do actual editing:
-        os.system(editor + " \"" + temp_file + "\"")
+        os.system(f"{editor} \"{temp_file}\"")
```
aqw, line 232: Avoid escaping quotes unless you absolutely need to.

```
os.system(f"{editor} '{temp_file}'")
```
or
```
os.system(f'{editor} "{temp_file}"')
```

onyo/utils.py
```diff
@@ -165,59 +191,72 @@ def check_float(value):
...
                     elif further_editing == 'n':
                         output_str = "No changes made."
                         # onyo new should neither create a new file nor a
                         # temp-file, and have a special info message
                         if (onyo_new):
-                            run_cmd("rm \"" + file + "\"")
-                            output_str = "No new asset \"" + os.path.relpath(file, onyo_root) + "\" created."
-                        run_cmd("rm \"" + temp_file + "\"")
+                            Path.unlink(file)
+                            output_str = f"No new asset \"{file.relative_to(onyo_root)}\" created."
```
aqw, line 255: Same comment on escaping.

onyo/utils.py
```diff
@@ -261,14 +300,10 @@ def get_config_value(name, onyo_root):
...
-        if os.path.isfile(elem):
-            # when assets are in .gitignore, they should not be listed as such
-            if run_cmd("git -C \"" + repo_path + "\" check-ignore --no-index \"" + elem + "\""):
-                continue
-            assets.append([os.path.relpath(elem, repo_path), os.path.basename(elem)])
-    return assets
+    """
+    Return a list of all assets in an onyo repository.
+    """
+    return [[x[0][0], Path(x[0][0]).name] for x in Repo(repo_path).index.entries.items() if not is_protected_path(x[0][0])]
```
aqw, line 306: I recommend writing this on two lines:

```
return [[x[0][0], Path(x[0][0]).name] for x in Repo(repo_path).index.entries.items()
                                      if not is_protected_path(x[0][0])]
```

tests/commands/test_new.py
```diff
@@ -0,0 +1,83 @@
...
+             'three',
+             'overlap/one',
+             'overlap/two',
+             'overlap/three',
+             'very/very/very/deep'
+             ]
+
+
+def populate_test_repo(path):
+
```
aqw, line 26: Nit: unneeded blank line.

tests/commands/test_new.py
```diff
@@ -0,0 +1,83 @@
...
+def test_new_non_interactive():
+    populate_test_repo('top_level')
+    os.chdir('top_level')
+
+    # create new asset for all different folders
+    for i in range(0, len(test_dirs)):
+        input_str = 'laptop\napple\nmacbookpro\n' + str(i)
+        file = 'laptop_apple_macbookpro.' + str(i)
+        ret = subprocess.Popen(['onyo', 'new', '--non-interactive', test_dirs[i]],
```
aqw, line 50: Is there a reason for using `Popen` here rather than `subprocess.run`? It's meant exactly for this use case. Less boilerplate and no need to use `communicate`.

tests/commands/test_new.py
```diff
@@ -0,0 +1,83 @@
...
+    os.chdir(original_cwd)
+
+
+def test_new_non_interactive():
+    populate_test_repo('top_level')
+    os.chdir('top_level')
+
+    # create new asset for all different folders
+    for i in range(0, len(test_dirs)):
+        input_str = 'laptop\napple\nmacbookpro\n' + str(i)
```
aqw, line 48: f-strings can be used outside of `print` statements, and will auto-cast variables when possible.
```
input_str = f'laptop\napple\nmacbookpro\n{i}'
```

tests/commands/test_new.py
```diff
@@ -0,0 +1,83 @@
...
+    # return to home
+    os.chdir(original_cwd)
+
+
+def test_new_non_interactive():
+    populate_test_repo('top_level')
+    os.chdir('top_level')
+
+    # create new asset for all different folders
+    for i in range(0, len(test_dirs)):
```
aqw, line 47: You probably want `enumerate` here:
```
for idx, dir in enumerate(test_dirs):
    ...
```

It's more readable (IMO) and allows you to just use `dir` rather than `test_dirs[i]`.

tests/commands/test_new.py
```diff
@@ -0,0 +1,83 @@
...
+             'two',
+             'three',
+             'overlap/one',
+             'overlap/two',
+             'overlap/three',
+             'very/very/very/deep'
+             ]
+
+
+def populate_test_repo(path):
```
aqw, line 25: Is there a reason why a new repo is generated for each test?

`rm` and `mv` were destructive, and it made it difficult to know what was where. But `new` is additive.

Creating new repos is time consuming, and if the work isn't needed, we should avoid it, for the sake of keeping test runtime down.

tests/test_utils.py
```diff
@@ -133,6 +134,52 @@ def test_get_editor_precedence():
...
+    assert ret.type == ValueError
+
+    # test that faux serial numbers can't be requested longer than 37 and the
+    # function raises the correct ValueError:
+    with pytest.raises(ValueError) as ret:
+        utils.generate_faux_serial(onyo_root, faux_length=38)
+    assert ret.type == ValueError
+
+    # technically, creating multiple faux serial numbers without actually
+    # creating assets might lead to duplicates. In practice, this should not
```
aqw, line 176: This comment is unnecessary. They cannot possibly conflict, as they are all different lengths.

aqw (changes requested):
I added a few comments, until it turned into a review. So there might be a weird mix of "single comments" and review comments.

Overall, this looks a lot better. Primary concerns are strings (f-string and unnecessary escaping) and tests (Popen, and `populate_test_repo()`)

tests/commands/test_new.py
```diff
@@ -0,0 +1,64 @@
...
+    ret = subprocess.run(['onyo', 'mkdir'] + test_dirs)
+    assert ret.returncode == 0
+
+    # return to home
+    os.chdir(original_cwd)
+
+
+def test_new_non_interactive():
+    populate_test_repo('testing_new')
+    os.chdir('testing_new')
```
aqw, line 43: This is actually not necessary.

`change_cwd_to_sandbox(()` (in `conftest.py`) puts every test-file into its own "sandbox" directory. It's just a dir, so it still needs an `onyo init`, but you don't need to create yet another subdir inside of it if you're only going to use one.

You can simply `populate_test_repo('./')` and remove the `os.chdir` in all of the tests.

onyo/utils.py
```diff
@@ -74,6 +76,30 @@ def get_editor(onyo_root):
...
+    between 1 and 37.
+
+    Returns on success a unique faux serial number which does not appear in any
+    other asset file name in any directory of the repository.
+    """
+    # check that the requested faux serial number has a valid length
+    if faux_length < 1 or faux_length > 37:
+        raise ValueError("Length of faux serial numbers must be between 1 and 37")
+
+    # generate a new faux serial number while it is in list_of_assets, and
```
aqw, line 93: This comment confused me, specifically the "while it is in list_of_assets". That is bringing the code logic into the comment.

Comments should provide the "straight dope" of what's being achieved.

aqw (changes requested):
Almost there. One last bit with the `os.chdir()`. Then it's ready to merge.

aqw (approved):
Looks good!

PR #217, opened 2022-10-07
RF: onyo invocation; add tests

aqw:
This PR refactors main, improves its readability, adds tests and docstrings, and fixes multiple bugs.

The tests focus on behavior when invoking  `onyo` itself and its flags. Helper functions are also added, which should be useful for other tests which could benefit from [powersets](https://en.wikipedia.org/wiki/Power_set).

Bugs fixed:

- insufficient number of args to onyo would still exit 0
- args named 'config' to subcommands would result in subsequent args being ignored
- onyo -d -C /tmp (and similar) would crash with Python trace

### DIFF
--- a/docs/source/cmd_cat.rst
+++ b/docs/source/cmd_cat.rst
@@ -3,7 +3,7 @@ onyo cat
 
 .. argparse::
    :module: onyo.main
-   :func: parse_args
+   :func: setup_parser
    :prog: onyo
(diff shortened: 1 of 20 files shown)


TobiasKadelka:
Especially the better handling of the argument "config" makes me happy :)

aqw:
Thanks! Merging. :-)

PR #223, opened 2022-10-13
RF: `onyo edit`

TobiasKadelka:
`onyo edit` is refactored and ` onyo new` is updated to adapt to changes, since the commands and functions now expect e.g. pathlib objects, use the repo class, return values changed, and error handling was updated.

Done:
- `onyo edit`
  - `utils.edit_file()` got moved to `onyo edit` and its code simplified
  - `utils.get_editor()` got moved to `onyo edit` together with its tests
  - prints diff-like output and asks if should be saved

- `onyo new`
  - Updates flag `onyo new --non-interactive` (like in `onyo history`) to simplify code
  - Adapts command `onyo new` to use new class, functions, etc (preparation for refactoring)

- `test_edit.py`:
  - move tests for editor from `utils.edit_file()`
  - add `test_edit_single_asset()`
  - add `test_edit_multiple_assets()`
  - add `test_edit_protected()`
  - add `test_edit_without_saving_changes()`
  - add `test_continue_edit_no()`
  - add `test_edit_non_existing_file()` 
  - add `test_edit_without_changes()`
  - add `test_edit_with_dot_dot()`


onyo/commands/edit.py
```diff
@@ -4,40 +4,43 @@
...
+    return [f'git -C "{onyo_root}" commit -m "edit asset(s).\n\n"', "\n".join(files)]
 
 
 def prepare_arguments(sources, onyo_root):
     problem_str = ""
     list_of_sources = []
+    asset_list = get_list_of_assets(onyo_root)
+
     if isinstance(sources, str):
         sources = ["".join(sources)]
```
aqw, line 31: I'm confused here. Perhaps I'm misunderstanding, but it looks like it wants to make sure that `sources` is a list. If so, why the `join`? Why not simply:
```
sources = [sources]
```

onyo/commands/edit.py
```diff
@@ -4,40 +4,43 @@
...
+        current_source = Path(source)
+        if not current_source.is_file():
+            current_source = Path(onyo_root, source)
+        if not current_source.exists():
+            problem_str += f"\n{current_source} does not exist."
         # check if file is in git
-        run_output = run_cmd("git -C \"" + onyo_root + "\" ls-tree -r HEAD ")
-        if source not in run_output:
-            problem_str = problem_str + "\n" + current_source + " is not in onyo."
+        elif True not in [source in asset[0] for asset in asset_list]:
```
aqw, line 39: `True not in []` is more complicated than `False in []`. But Python's truthiness is your friend. You can simplify this to `not []`. For example:

```
elif not [asset[0] for asset in asset_list if source in asset[0]]
```

onyo/commands/edit.py
```diff
@@ -4,40 +4,43 @@
...
+            problem_str += f"\n{current_source} is not in onyo."
```
aqw, line 40: This error message (and the supporting comments) could be more clear here.

If I understand it correctly, this is trying to guard against the situation where a user attempts to edit an untracked file, which would have to be in `.gitignore` (in order to pass `onyo fsck`). Yes?

Or... I see that you're using `asset_list`, which /currently/ returns a list of all files, but eventually will in fact only return files which match the asset pattern. Are you guarding against `onyo edit` being used on non-assets?

In either case, better messaging and comments would help understand the programmer's intent here.

onyo/commands/edit.py
```diff
@@ -62,17 +65,24 @@ def edit(args, onyo_root):
 
     # check and set paths
     list_of_sources = prepare_arguments(args.asset, onyo_root)
+    changed_files = []
+
     # iterate over file list, edit them, add changes
     for source in list_of_sources:
         git_filepath = os.path.relpath(source, onyo_root)
```
aqw, line 72: Missed an `os.path`

onyo/commands/new.py
```diff
@@ -67,10 +69,17 @@ def run_onyo_new(directory, template, non_interactive, onyo_root, repo):
         filename.touch()
 
     # open new file?
```
aqw, line 71: Why is the comment uncertain?

onyo/utils.py
```diff
@@ -190,72 +190,97 @@ def check_float(value):
...
     """
-    Check if the selected editor exists. Check if a file exists, create a copy
-    of it in .onyo/temp/ to change with the editor, and check the temp file
-    after editing for correct yaml syntax and validity specified by the onyo
-    validation file.
-    If the asset is not valid, open a dialog to enable to correct the asset, or
-    remove the changes.
-
-    Returns True on success.
+    Decide for an existing asset file, if a temporary file needs to be created,
```
aqw, line 195: "Decide for". Do you mean "Determine if"?

onyo/utils.py
```diff
@@ -190,72 +190,97 @@ def check_float(value):
...
+    or if one already exists and should be used.
+    If no temporary file exists, create a copy of the given asset.
```
aqw, line 197: Why would a temp file already exist? And wouldn't it be dangerous to use an unknown stale temp file that may have no relation to the file that the user asked to edit?

onyo/utils.py
```diff
@@ -190,72 +190,97 @@ def check_float(value):
...
+    If one exists, ask the user if it should be further edited, or if the
+    changes should be discarded, in which case the temporary file gets
+    overwritten by a new copy of the input asset.
+
+    Returns on success the Path.object of the temporary asset file.
```
aqw, line 202: Nit: "Returns on success the Path() of the temporary asset file."

onyo/commands/new.py
```diff
@@ -67,10 +69,17 @@ def run_onyo_new(directory, template, non_interactive, onyo_root, repo):
...
-    if not non_interactive:
-        edit_file(filename, onyo_root, onyo_new=True)
-    # add file to git
-    repo.git.add(filename)
+    creation_successful = True
```
aqw, line 72: Personal bias, but I think `successful` is clear (and shorter).

onyo/utils.py
```diff
@@ -190,72 +190,97 @@ def check_float(value):
...
-    # verify existence of file to edit
-    file = Path(file)
-    if not file.is_file():
-        logger.error(f"{file} does not exist.")
-        sys.exit(1)
-
-    # create and edit a temporary file, and if that is valid replace original
     temp_file = Path(get_git_root(onyo_root), ".onyo/temp/", file.name)
+
+    # if no temporary file exists, just create a copy of the input file
```
aqw, line 206: This comment just restates what is already clearly readable on the next line.

Beware comments that start with "if".

onyo/utils.py
```diff
@@ -190,72 +190,97 @@ def check_float(value):
...
     if not temp_file.is_file():
         shutil.copyfile(file, temp_file)
-    # When temp-file exists, ask if to use it
-    elif temp_file.is_file():
+
+    # Otherwise ask if it should be used, or overwrite it with a new copy of the
```
aqw, line 210: "Otherwise" == "else"

onyo/utils.py
```diff
@@ -190,72 +190,97 @@ def check_float(value):
...
-            edit_temp = str(input(f"Temporary changes for {file} exist. Continue editing? (y/N) "))
-            if edit_temp in ['y', 'Y', 'yes']:
+            edit = input(f"Temporary changes for {file} exist. Continue editing? (y/N) ")
+            if edit in ['y', 'Y', 'yes']:
                 break
-            elif edit_temp == 'n':
+            elif edit == 'n':
                 shutil.copyfile(file, temp_file)
                 break
+    # return the temporary file
```
aqw, line 220: Again, this restates what is already crystal clear on the next line.

onyo/utils.py
```diff
@@ -190,72 +190,97 @@ def check_float(value):
...
+    return temp_file
+
+
+def validate_yaml_syntax(file):
```
aqw, line 224: `valid_yaml()` or `is_valid_yaml()` are shorter and will read more naturally in an `if` statement.

aqw (changes requested):

onyo/commands/edit.py
TobiasKadelka: I think `False in []` equals not `not []`:
```
>>> False in [False, False, True]
True
>>> not [False, False, True]
False
```

So I will for now simplify to your proposed  `False in []`, thanks!

aqw: > I think False in [] equals not not []:

That's why I suggested it. ;-P

> Further update, not any([source in asset[0] for asset in asset_list]) seems to work best and is also fairly readable.

Have you profiled that? I'm not sure if that is faster, as it will need to assign a lot of True/False values and then check them.

Profiling...:
```
def straight(source, asset_list):
    if not [asset[0] for asset in asset_list if source in asset[0]]:
        yope='yope'

# my mispronunciation of "any"
def ernie(source, asset_list):
    if not any([source in asset[0] for asset in asset_list]):
        yope='yope'

asset_list = [(f"{x}", f"{x}") for x in range(1,1000) ]
```
```
In [236]: timeit.timeit(lambda: straight('1', asset_list), number=50000)
Out[236]: 1.3253654120489955

In [237]: timeit.timeit(lambda: straight('no_match', asset_list), number=50000)
Out[237]: 1.0206323569873348

In [238]: timeit.timeit(lambda: ernie('1', asset_list), number=50000)
Out[238]: 1.4346348019316792

In [239]: timeit.timeit(lambda: ernie('no_match', asset_list), number=50000)
Out[239]: 1.5336767170811072
```

TobiasKadelka: I did indeed miss-read this, and will update accordingly, thanks.

onyo/commands/edit.py
```diff
@@ -3,39 +3,79 @@
...
-    if problem_str != "":
-        log.error(problem_str)
+    if problem_with_source:
+        print(problem_with_source, file=sys.stderr)
         sys.exit(1)
     return list(dict.fromkeys(list_of_sources))
 
 
-def edit(args, onyo_root):
+def edit(args, onyo_root: str):
```
aqw, line 78: Missing return type for the function.

onyo/commands/edit.py
```diff
@@ -57,16 +98,19 @@ def edit(args, onyo_root):
...
+        successful = edit_asset(editor, source)
+        if successful:
+            repo._git(['add', source])
+        else:
+            repo._git(['restore', source])
+            print(f'"{source}" not updated.')
 
     # commit changes
-    files_staged = repo.files_staged
+    files_staged = '\n'.join([file.name for file in repo.files_staged])
```
aqw, line 114: The new `Repo().commit()` accepts lists and allows you to avoid this.

onyo/commands/edit.py
```diff
@@ -57,16 +98,19 @@ def edit(args, onyo_root):
         sys.exit(1)
 
     # check and set paths
-    list_of_sources = prepare_arguments(args.asset, onyo_root)
-    # iterate over file list, edit them, add changes
+    list_of_sources = prepare_arguments(args.asset, repo)
+    editor = get_editor(repo.root)
+
+    # change files and add changes to git
     for source in list_of_sources:
```
aqw, line 105: Please retire the name "source". This is a copy/paste from `mv` and others. There is no "source" here. They are assets.

`list_of_sources` is needlessly verbose. Try:
```
for a in assets:
```

onyo/commands/edit.py
```diff
@@ -3,39 +3,79 @@
...
+                further_editing = input("Continue editing? No discards changes. (y/N) ")
+                if further_editing in ['y', 'Y', 'yes', 'n', 'N', 'no']:
+                    break
+
+        if further_editing in ['n', 'N', 'no']:
+            return False
+    return False
+
+
+def prepare_arguments(sources: list, repo: Repo) -> list:
```
aqw, line 57: What does this function do?

I assumed it prepared the arguments, but it seems to return a list of assets.

A better name and a docstring would help reduce the ambiguity.

Please also specify the types inside the lists when possible. Such as:
```
list[Path]
```

aqw (changes requested):
I primarily reviewed `edit.py`. There are enough changes necessary that I chose not to review the others.

onyo/commands/edit.py
```diff
@@ -47,6 +87,7 @@ def edit(args, onyo_root):
     problems are found, the choice will be offered to reopen the editor to fix
     them, or abort and return to the original state.
     """
+    repo = ""
```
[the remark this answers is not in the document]
TobiasKadelka: Pyre thinks different, because it does not know about `sys.exit(1)` in catch and thinks repo would not be initialized in the error case:
`Uninitialized local [61]: Local variable `repo` is undefined, or not always defined`
(https://github.com/psyinfra/onyo/actions/runs/3291764861/jobs/5426323059 shows me trying to push it without the line.)

https://pyre-check.org/docs/errors/ has a section about initilizing in try/catch specifically, which says one should either initilize before, or alternatively if that is not liked, move all the following code that uses repo into the try block, which is considered bad practice.
Repo has no empty initilizer, so I choose the string.
I will check for different solutions again later, but this is why this is right now unfixed.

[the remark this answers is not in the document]
aqw: Then `repo = None` would be more appropriate.

tests/commands/test_edit.py
```diff
@@ -0,0 +1,225 @@
...
+    try:
+        repo = Repo(onyo_root)
+    except OnyoInvalidRepoError:
+        sys.exit(1)
+
+    # check edit for a list of assets all at once
+    ret = subprocess.run(['onyo', 'edit'] + list(repo.assets), input='y', capture_output=True, text=True)
+    assert ret.returncode == 0
+    assert "Save changes?" in ret.stdout
+    assert "+key: multiple_assets" in ret.stdout
```
aqw, line 70: This string should be present `n` times, correct? If so, a `count` would be more appropriate here.

tests/commands/test_edit.py
```diff
@@ -0,0 +1,225 @@
...
+    assert not ret.stderr
+
+    # verify the changes were saved for all assets and the repository is clean
+    for asset in repo.assets:
+        ret = subprocess.run(['cat', asset], capture_output=True, text=True)
+        assert 'key: multiple_assets' in ret.stdout
+    assert not repo.fsck()
+
+
+def test_edit_without_saving_changes():
```
aqw, line 80: `without_saving_changes` -> `discard`

tests/commands/test_edit.py
```diff
@@ -0,0 +1,225 @@
...
+    os.environ['EDITOR'] = "printf 'key: dot_dot' >"
+    try:
+        repo = Repo(onyo_root)
+    except OnyoInvalidRepoError:
+        sys.exit(1)
+
+    # check edit with a path containing a ".." that leads outside the onyo repo
+    # and then inside again
+    asset = Path("../test_edit.py/laptop_apple_macbookpro.0")
+    assert asset.is_file()
```
aqw, line 121: Nothing has changed yet. I don't see the need for this assert.

aqw (changes requested):
This is trending in the right direction. `edit` looks approachable now.

I skimmed `new`, and read through most of the tests.

onyo/commands/edit.py
```diff
@@ -1,41 +1,95 @@
...
+    """
+    Checks for a list of assets if they are valid paths to files that exist as
+    onyo assets in the repository.
+
+    Returns a list of Path() objects if the assets exist in the onyo repository.
+    """
+    valid_assets = []
+
+    for asset in assets:
+        asset = Path(asset).resolve().relative_to(repo.root)
```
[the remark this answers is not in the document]
TobiasKadelka: The next 4 lines print the paths out in error messages to the user, therefore I will use the relative path for them.

onyo/commands/edit.py
```diff
@@ -1,41 +1,95 @@
...
+        if not asset.exists():
```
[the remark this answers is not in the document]
TobiasKadelka: The idea was to give a more precise information in case the file exists but is not an asset.

onyo/commands/edit.py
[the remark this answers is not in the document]
TobiasKadelka: Update: `repo.assets` does not use absolute paths, so I can't check `if asset in repo.assets` if not doing this relative, too. If we decide to change `repo.assets` I will update this.

tests/commands/test_edit.py
```diff
@@ -0,0 +1,225 @@
+import os
+import subprocess
+import sys
+from pathlib import Path
+from onyo import commands  # noqa: F401
+
+from onyo.lib import Repo, OnyoInvalidRepoError
+
+onyo_root = "./"
```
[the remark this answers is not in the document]
TobiasKadelka: I do not understand what you mean with this.
`onyo_root` is used in multiple functions, even if just every time once to initialize the repository. Do you want the `"./"` string to be used directly in each function instead once on top?

aqw (changes requested):

aqw (approved):

onyo/commands/edit.py
```diff
@@ -3,39 +3,116 @@
...
+        asset = Path(asset).resolve().relative_to(repo.root)
+        if asset not in repo.assets:
+            print(f"\n{asset} is not an asset.", file=sys.stderr)
         else:
-            list_of_sources.append(current_source)
-    if problem_str != "":
-        log.error(problem_str)
+            valid_assets.append(asset)
+
+    if len(valid_assets) != len(assets):
```
aqw, line 110: This should print an error to `stderr` rather than exiting silently.

onyo/commands/edit.py
```diff
@@ -57,16 +135,26 @@ def edit(args, onyo_root):
...
-        repo._git(['add', git_filepath])
+    assets = sanitize_assets(args.asset, repo)
+    editor = get_editor(repo.root)
+
+    for asset in assets:
+        if edit_asset(editor, asset):
+            repo.add(asset)
+        else:
+            """
+            If user wants to discard changes, restore the assets state
```
aqw, line 146: assets -> asset's

onyo/commands/edit.py
```diff
@@ -57,16 +135,26 @@ def edit(args, onyo_root):
...
 
     # commit changes
     files_staged = repo.files_staged
     if files_staged:
-        repo._git(['commit', '-m', 'edit asset(s).\n\n' + '\n'.join(files_staged)])
+        print(diff(repo))
+        if request_user_response("Save changes? No discards all changes. (y/n) "):
+            repo.commit('edit asset(s).', files_staged)
+        else:
+            for asset in files_staged:
```
aqw, line 158: `git restore` accepts multiple arguments. Rather than a for-loop, you can simply pass them all directly to `git`. That way it's only one `git` call rather than n-edited-assets.

onyo/commands/new.py
```diff
@@ -42,37 +42,45 @@ def read_new_word(word_description):
...
     if template.is_file():
-        shutil.copyfile(template, filename)
+        shutil.copyfile(template, temp_asset)
     else:
-        filename.touch()
+        log.error(f"{template} does not exist.")
+        sys.exit(1)
+
+    if new_asset.exists():
+        log.error(f"{new_asset}' asset already exists.")
```
aqw, line 65: Nit: missing opening `'`.

Not a blocker, as `new` will be refactored soon.

onyo/commands/new.py
```diff
@@ -42,37 +42,45 @@ def read_new_word(word_description):
...
+    new_asset = Path(directory, create_filename(repo))
+    temp_asset = Path(repo.root, ".onyo/temp/", new_asset.name)
+
+    template = Path(template).resolve()
     if template.is_file():
-        shutil.copyfile(template, filename)
+        shutil.copyfile(template, temp_asset)
     else:
-        filename.touch()
+        log.error(f"{template} does not exist.")
```
aqw, line 61: Nit: I like to quote user-controlled variables in single quotes, just in case something is empty or has spaces. It makes it easier debug and more obvious to the user what's going on.

Not a blocker, as `new` will be refactored soon.

tests/commands/test_edit.py
```diff
@@ -0,0 +1,300 @@
+import os
+import pytest
```
aqw, line 2: Nit: convention[1] states that imports of 3rd-party modules should come after standard-lib imports.

Not a blocker, just FYI.

[1] https://peps.python.org/pep-0008/#:~:text=Imports%20should%20be%20grouped%20in,Local%20application%2Flibrary%20specific%20imports.

tests/commands/test_edit.py
```diff
@@ -0,0 +1,300 @@
...
+    Get the editor from git settings.
+    """
+    # set the editor
+    ret = subprocess.run(["git", "config", "onyo.core.editor", 'vi'],
+                         capture_output=True, text=True)
+    assert ret.returncode == 0
+
+    assert get_editor(repo.root) == 'vi'
+
+    # cleanup
```
aqw, line 28: Since each test is separate, there's no need to cleanup. And indeed, it's arguably better if it is not cleaned up, as it's easier to inspect after-the-fact.

aqw (changes requested):
This looks almost ready to merge.

Most of my comments are nits and not blockers.

The only stuff I'd like to see addressed is some `stderr` prints and passing all args to `git restore` at once. The rest is up to you whether to address.

tests/commands/test_edit.py
TobiasKadelka: Thanks, I adjusted it for all editor-tests.

aqw:
LGTM!

PR #235, opened 2022-10-24
TST/BF: make test_history.py platform independent

TobiasKadelka:
Change command calls from e.g. `/bin/printf` to `/usr/bin/env printf` to use tests platform independent.

`test_history.py` used built-in commands e.g. `/bin/printf`. These differ on macbooks (e.g. `/usr/bin/printf`) so the tests for the `onyo history` command failed on macOS.


aqw:
Thanks!

PR #238, opened 2022-10-26
NF/RF: mv -> Repo.mv() ; add tests

aqw:
Notable changes:

- introduce `Repo().mv()` (with tests)
- add a bunch of new tests, covering a lot of corner cases
- fix bugs around explicit move (`file-1` -> `subdir/file-1`)

As with the other conversion, the tests for the command `mv` have been left intact. They will be address later when the logging overhaul happens.

### DIFF
--- a/onyo/commands/mv.py
+++ b/onyo/commands/mv.py
@@ -1,181 +1,19 @@
-import logging
-import re
 import sys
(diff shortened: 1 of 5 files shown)


TobiasKadelka:
Very nice 👍

PR #240, opened 2022-10-27
RF: move setup_parser() to `main.py`; simplify key-value parsing

aqw:
This PR includes #239, to reduce compute time.

Notable changes are:

- move `setup_parser()` (and supporting functions) to `main.py`
- pass the `parser` object directly into `shell-completion`
- remove a bunch of unneeded imports of `onyo.commands` that were guards against circular imports
- simplify parsing of key-value pairs

The key-value parsing should handle all situations except for excess commas. I had a solution that abused shlex in order to handle commas correctly, but it requires quotes. All quotes are stripped by the time it reaches this function, rendering that fanciness useless.

The `csv` module would seem like a natural fit for parsing the key-value pairs, but it chokes on quoting portions of a field (e.g. key='value') and mixed single and double quotes.

----

Leaving the shelex solution here, for posterity, just in case it can be of use elsewhere:

```
def parse_key_values(string):
    lex = shlex.shlex(string, posix=True)
    # split on commas
    lex.whitespace = ','
    # remove meaning for all shell chars
    lex.wordchars += '~-./*?= '

    results = {k: v for k, v in (pair.split('=') for pair in list(lex))}
    for k, v in results.items():
        try:
            results.update({k: int(v)})
        except ValueError:
            try:
                results.update({k: float(v)})
            except ValueError:
                pass

    return results
```


TobiasKadelka:
LGTM :)

PR #241, opened 2022-10-27
remove old validation code

### LINKED ISSUE #93: Add functionality for regular expressions
While the terminal automatically parses and expands some regular expressions (e.g. `onyo rm shelf/laptop_*`), there are additional cases mentioned in the readme, which should be newly discussed, implemented and then documented in the correct form.

> Field Validation
> [...]
> Entries can be strings or, for more complex needs, any valid [Python Regular Expression](https://docs.python.org/3/library/re.html#regular-expression-syntax).
> Values are checked for reserved characters before they are checked against the validation strings/patterns.
> One can effectively disable validation for a field by adding .* to the corresponding validation list.

Nothing about this is true right now, `onyo new` checks fields depending on the template used (including default, when nothing is specified, see PR #92 ), the validation can not be ignored in the way discribed. Therefore I will safe the text here, and remove it from the current readme with PR #91.

### LINKED ISSUE #94: Validation: keys "deprecated" and "disallowed"
The file containing Validation rules for assets should make it possible to have fields to warn or disallow certain subfields for specific assets.

These might look e.g. like:
```
laptop_*
Ram:
- deprecated/warning: True
- error/disallow: True
```

When one of these rules is added, `onyo` should either warn when fields are used, or even block the user from adding these fields to certain assets.

### LINKED ISSUE #95: Validation: allow subcommands for field validation
We want to allow the user to validate fields/values with certain commands. This might happen in the validation file like:

```
RAM:
- Validation: /program/to/validate {asset.disk}
```

We need to define how these fields will look like and how they are used by onyo.
For this, jinja might be a good fit.

### LINKED ISSUE #96: Validation: add key to allow only defined asset fields
When creating certain types of assets, it might be wished, that only explicetely allowed defined fields are added to it and all other added fields not.
This might happen e.g.:

```
laptop_*
Ram:
- type: string
only_defined_keys: true
```

### LINKED ISSUE #97: Validation: allow user specific warnings/messages in error cases
When a user creates a new asset, the validation might fail, e.g. for syntax reasons.
We might want to allow the user to add a certain message to the validation file for each field, that looks like that:
```
laptop_*
display:
   - Validate: [integer]″
   - message: screensize need to be given in format: [integer]″
```

### LINKED ISSUE #98: Validation: allow regex
Regex might be the easiest way to define the format that values for fields should have.

```
laptop_*
RAM:
- validation: <some regex>
```

### LINKED ISSUE #102: Validation: enable additional type bool
Booleans are a common type one might want to have as metadata.
One has to keep in mind and test, that python implements booleans as numbers (https://stackoverflow.com/questions/37888620/comparing-boolean-and-int-using-isinstance)
```
print isinstance(True, (float, int))
True
```

So the function which checks the type should make sure, that onyo differentiates between these.
And, possibly more important, it should differentiate between False as a boolean and a string, to keep this information consistent along assets.

### LINKED ISSUE #103: Validation: enable lists?
The general value "list" might be easy to add and check with python, but I am not sure how one would go on and specify in the validation file, which types the values in the list have, or how to specify the lenght of the list in flexible ways.

This might be easier and more obvious, if the #98 is implemented with regex.

### LINKED ISSUE #110: validation: "required" flag
It should be possible to add a key/flag to the validation.yaml that forces the user for assets in a certain locations (or with certain names) to define a key that has to be added to an asset, for it to be valid.

### LINKED ISSUE #219: consolidate to a single YAML parser
Currently both `yaml` and `ruamel` are in use. Only one should be used (`ruamel`).

### LINKED ISSUE #220: ruamel should always be in "rt" mode

aqw:
This will eventually be replaced with an alternative (likely Cerberus). Until then, this is dead code that gets in the way.

I kept `fsck`'s `asset-validity` as it's already in a usable form.

Invalidate many old-validation-system issues:

closes #93
closes #94
closes #95
closes #96
closes #97
closes #98
closes #102
closes #103
closes #110

This also allows us to finally use only one YAML parser. This drops the last bits of pyyaml, and changes all used of ruamel to use "roundtrip" mode.

closes #219
closes #220


aqw:
Merging this giga issue-closing PR. :-D

PR #242, opened 2022-10-28
WIP: TST/RF: assorted cleanup

aqw:
This is a collection of seeming-random, but actually interconnected changes.

Changed of note:
- quick-and-dirty modernization of `tree`. No thought was put into repo-ification, which also seems largely unneeded
- renaming of `onyo_root` to either `opdir` (which is what it usually means) or `repo_root` (which it sometimes means)
- update the `utils.py` tests to use the `repo` fixture and update some comments
- finally drop `run_cmd()`, `build_git_add_cmd()`, and `get_git_root()` from utils, as the above were the last consumers.

### DIFF
--- a/onyo/commands/cat.py
+++ b/onyo/commands/cat.py
@@ -8,7 +8,7 @@ logging.basicConfig()
 log = logging.getLogger('onyo')
 
 
-def sanitize_paths(paths, onyo_root):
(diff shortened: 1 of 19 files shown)


aqw:
+1 from Tobias off-list. Merging.

Issue #147, opened 2022-09-05
create multiple new assets at once


aghisla:
One of our common use cases (and where we avoid a ton of typing + typos) is batch creation/clone of assets, for example when we get several (from 2 to 10) identical laptops or screens.

In GLPI we solved this by cloning an existing device and editing the clones one by one, replacing information that should be unique (serial number and later inventory number), but it was the best GLPI could offer and it has no validation on unique metadata, which is problematic.

For onyo, Jan and I have initially thought of an `onyo clone` command with the following options:

- asset to be cloned
- how many new clones
- where to put the clones (default: shelf)
- open each new clone in an editor (default) or quick-edit (onyo prompts only for the serial number of each clone)

but we are open to other workflows+implementations that are as linear as this one.

I thought for example about `onyo new` and a almost partially filled template, but I have not yet really decided if templates should only contain keys, at least the fundamental templates - and how onyo should deal with partially filled templates.

aqw:
This is a use case Tobias and I have talked about a bit, but havn't yet come up with a clear answer.

The main topics have been around:

Use cases:
1) import a large database (such as GLPI->onyo). This is likely a one-time event (`onyo import`). This becomes complicated, because it spans across locations and asset types. This, in and of itself isn't a problem, but validation could make the user's experience of `import`ing 1,000 assets painful. How does one skip, resume, etc.
2) add 1 new asset that is very similar to another asset (`onyo clone`). This overlaps quite a bit with templates, and I don't think will be that useful/ergonomic, because a user would have to find another asset somewhere in the tree.
3) add 10 identical (except for serial) laptops (`new --batch`). `onyo new` also allows piping into it (albeit not very ergonomically). So the current solution to this problem is to create a template, create a list of serial numbers, loop over that list and pipe into `onyo new` 10 times. An alternative is to add an option (say `--batch`, `--bulk`, or `--csv`) to `onyo new` to allow multiple assets to be creating in one invocation. This is very, very similar to `onyo import`, but can be interesting with `onyo new --template`.

I currently find option 3 to be most flexible, but the way it will work is still not entirely clear to me. Should `onyo new --batch` create 10 commits, or 1? Should it allow using a template (pre-filling out some fields), and if so, should it allow specifying the template in the CSV file to allow for complicate situations (thus satisfying much of the `onyo import` scenario)?

aghisla:
[...]

Issue #123, opened 2022-08-29
`onyo new` accepts leading whitespace


aghisla:
While testing how Onyo reacts to existing serial numbers, I typed " foo" as new serial, and the resulting filename had a space: `headset_Corsair_Virtuoso RGB. foo`

Nothing breaks because of it so far but I would avoid it - and I think I typed a space as I would have done when editing an asset file by hand.

TobiasKadelka:
I think that was the state in which onyo was a few months ago, a decission was made to allow white space in all fields.

I am fine with automatically stripping the whitespace for each field, but that decission has then to be made actively.

Relevant in that context might be #70, since the python function gets the string pre-parsed from the terminal. Otherwise I would have said one could just do the same escaping for space with `\` like normal, instead of making the choice which has to be made now.

TobiasKadelka:
Cool, thanks a lot👌

aqw:
I like the idea of allowing as much as possible (including whitespaces). That being said, I do see a lot of practical, day-to-day value in stripping any leading or trailing whitespace (not just spaces).

TobiasKadelka:
I think they can't do that, because  of #70.

There are a few projects on the internet that try to emulate the whitespace behaviour from the shell for python, and obviously also some that can handle escaping, but when I looked for a good solution, I did not find something that did the whole job. There were always some other input-strings that did not work then anymore in the ways we wanted it to work.

This is not to say, that I do not still want it, but that problem together with the idea of "allowing as much as possible" lead to the decission to have it not strip the whitespace.

aqw:
> I think they can't do that, because of https://github.com/psyinfra/onyo/issues/70.

Kinda. But they're completely different worlds.

#70 is not a bug in my opinion.

This issue is solvable by simply running a trim on input.

TobiasKadelka:
Yes, if we decide that we do not allow leading/trailing whitespace anymore.
But I think we should allow it there, as much as we allow it in the middle of a field.

TobiasKadelka:
The new way for `onyo new` to behave is that it does not have the interactive dialog anymore, instead it expects a full directory and asset name. These are still allowed to contain leading whitespace and no trimming happens in the code, but these need to be escaped in the terminal, since the shell is now parsing the arguments.
Therefore the incomming PR #243 will close this issue.

PR #243, opened 2022-10-28
RF: `onyo new`

TobiasKadelka:
Code Updates:
- refactor `onyo new` and add tests for it
- rename template "standard" -> "empty"
- retire `.onyo/temp/`, build new assets the destination directory
- Updates existing tests so they are valid with the changes in `new`
  - includes test_invoking.py and reference output
- updates imports and uses modernized code (e.g. uses Repo class)
- updates argument parsing in main.py for changed `onyo new`
- adds `validate_name_scheme()` to repo class

User-visible changes in `onyo new`:
- adds `--yes`, `--set`, `--tsv` and `--edit` flags
- Instead of interactive keyboard input onyo expects the full
  asset name with directory path, checks it for validity and adds
  a faux serial number if needed (close https://github.com/psyinfra/onyo/issues/123)
- can now get one or multiple input-paths (close https://github.com/psyinfra/onyo/issues/147)
- can now read information for new assets from tsv file
- adds diff-like output and asks user to verify changes


onyo/commands/new.py
```diff
@@ -1,154 +1,318 @@
...
+            #       given for the same value?
+
+            # assets must have minimal information to be created
+            if not new_asset.destination_path or not new_asset.temp_path or not new_asset.template:
+                print(f"Information incomplete for {new_asset}", file=sys.stderr)
+                continue
+            new_assets.append(new_asset)
+
+    # If the information was valid and complete, each row contains one asset
+    if row_count != len(new_assets):
```
aqw, line 149: IMO, problems should exit right away, and this guard can be removed.

onyo/lib/onyo.py
```diff
@@ -900,3 +900,46 @@ def _rm_sanitize(self, paths: Iterable[Union[Path, str]]) -> list[Path]:
                                          '\n'.join(error_path_protected))
 
         return paths_to_rm
+
+    def validate_name_scheme(self, name: str) -> bool:
```
aqw, line 904: IMO, this belongs in the "NEW" section.

Also `validate_name_scheme()` -> `valid_name()`

is there a reason why you felt that `re` wasn't the right approach here? To me it feels like a better fit, and would greatly simplify the function.

aqw (changes requested):
I read through about half of the code.

Some nits here and there, but my primary concern is about the size of `read_assets_from_tsv()`, the duplication of logic between the TSV and CLI methods, and the necessity of the Asset class.

onyo/commands/new.py
```diff
@@ -1,154 +1,318 @@
...
-    new_asset = Path(directory, create_filename(repo))
-    temp_asset = Path(repo.root, ".onyo/temp/", new_asset.name)
 
-    template = Path(template).resolve()
-    if template.is_file():
-        shutil.copyfile(template, temp_asset)
-    else:
-        log.error(f"'{template}' does not exist.")
-        sys.exit(1)
+def move_assets_to_destination(assets: list[Asset], repo: Repo) -> None:
```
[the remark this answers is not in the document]
TobiasKadelka: This happens now "in place" a.k.a. `destination_path`, so that no moving to destinations, copying of trees or anything like that is needed.

onyo/main.py
```diff
@@ -327,10 +319,39 @@ def setup_parser():
...
+        action='store_true',
+        help='Open new assets to edit them before creation'
+    )
+    cmd_new.add_argument(
+        '-s', '--set',
+        action=StoreDictKeyPair,
+        metavar="KEYS",
+        help='key-value pairs to set in assets; multiple pairs can be separated by commas (e.g. key=value,key2=value2)'
+    )
+    cmd_new.add_argument(
```
aqw, line 334: Usually arguments are put at the end of a command's list.

onyo/commands/new.py
```diff
@@ -1,154 +1,296 @@
...
 
-def run_onyo_new(directory, template, interactive, repo: Repo):
+def create_assets_in_destination(assets: dict, repo: Repo) -> None:
     """
-    Read a new asset name, check it for uniqueness and validity, create the
-    asset (from a template if one was given), edit the new asset (if not
-    suppressed with flag) and stage it with git.
-
-    Returns the newly created asset's filename on success.
+    Create assets and set their contents, and if their directory does not exist
```
aqw, line 54: "Create and populate assets. Parent directories are create if necessary."

onyo/commands/new.py
```diff
@@ -1,154 +1,296 @@
...
+    yet, create it.
     """
+    for asset in assets.keys():
+        # create missing directories
+        if not asset.parent.is_dir():
```
aqw, line 59: What happens in the pathological case where the "parent" is actually a file/asset?

onyo/commands/new.py
```diff
@@ -1,154 +1,296 @@
...
+        # iterate over the table from the beginning to read asset information.
+        tsv_file.seek(0)
+        table = csv.DictReader(tsv_file, delimiter='\t')
+        for row in table:
+            row_count += 1
+            new_path = ""
+            contents = {}
+
+            # error if any required field is empty
+            if not row['type'] or not row['make'] or not row['model'] or not row['serial'] or not row['directory']:
```
aqw, line 96: `any()` can simplify this.

onyo/commands/new.py
```diff
@@ -1,154 +1,296 @@
...
+                print("The fields 'type', 'make', 'model', 'serial' and 'directory' are required, but missing in line {row_count}.", file=sys.stderr)
```
aqw, line 97: > but missing in line

This is an incomplete clause.

From a user standpoint, I think it would be helpful to also print the entire offending line.

onyo/commands/new.py
```diff
@@ -1,154 +1,296 @@
...
+
+            # verify that the asset name is valid and unique in repo and table
+            valid_asset_destination(new_path, repo, existing_assets, new_assets)
+
+            # either a template is given in table, CLI, or onyo config
+            if "template" in row.keys():
+                if row['template'] == "":
+                    print(f"Template value missing in {row_count}", file=sys.stderr)
+                    sys.exit(1)
+                template = Path(repo.root, ".onyo/templates", row['template']).resolve()
```
aqw, line 115: This feels like something that should be provided by the `repo` class.

Possibly a property `repo.templates` that is a set of Paths of all templates. And/or something that returns the Path object of the default template. The second I am uncertain about, but the first seems like a useful feature.

onyo/commands/new.py
```diff
@@ -1,154 +1,296 @@
...
-    if not template:
-        template = repo.get_config('onyo.new.template')
+    new_assets = {}
 
-    template = Path(repo.root, ".onyo/templates", template).resolve()
+    if tsv:
+        new_assets = read_assets_from_tsv(tsv, template, set_values, repo)
+    else:
+        new_assets = read_assets_from_CLI(assets, template, set_values, repo)
+    if len(new_assets) == 0:
```
aqw, line 202: Truthiness can be used here:
```
if not new_assets:
```

onyo/commands/new.py
```diff
@@ -1,154 +1,296 @@
...
+            print(f"{args.tsv} does not exist.", file=sys.stderr)
+            sys.exit(1)
+        with open(args.tsv, 'r') as tsv_file:
+            table = csv.reader(tsv_file, delimiter='\t')
+            header = []
+            # check that the first row (header) is complete
+            for row in table:
+                header = row
+                break
+            if 'type' not in header or 'make' not in header or 'model' not in header or 'serial' not in header or 'directory' not in header:
```
aqw, line 234: `any()` can simplify this.

aqw (changes requested):
I've read through most of this PR.

I did not review the tests closely. I will try to do so next time.

`read_assets_from_tsv()` is still unclear to me. I feel that the current approach mixes checking TSV validity/sanity, new asset sanity, and then builds everything. It's more intertwined than I would like.

onyo/commands/new.py
```diff
@@ -1,154 +1,296 @@
...
-    tries to read the same word again.
+    Select the template to use. If `--template` is not given, use the template
+    from the repository config file.
 
-    Returns the word on success.
+    Returns the template path on success, or leaves with error.
     """
-    # read word for field from keyboard
-    word = input(word_description)
+    if template:
```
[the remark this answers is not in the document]
TobiasKadelka: It is not a required CLI argument, if it is not explicitely given this is false, and it uses the default template from the `.onyo/config` file.

onyo/lib/onyo.py
```diff
@@ -833,6 +833,50 @@ def generate_faux_serials(self, length: int = 6, num: int = 1) -> set[str]:
...
+            log.error(f"'{asset.name}' must be like '<type>_<make>_<model>.<serial>'")
+            return False
+
+        # check against special characters in fields 'type', 'make', 'model'
+        if '_' in type or '.' in type or '_' in make or '.' in make or '_' in model or '.' in model:
+            log.error(f"'{asset.name}' must be like '<type>_<make>_<model>.<serial>'")
+            return False
+
+        # none of the fields is allowed to be empty
+        if not type:
```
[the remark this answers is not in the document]
TobiasKadelka: Since the change already drops the information which field is empty (which is fine), I change it not to `any()`, but to `if not all([type, make, model, serial]):`

onyo/commands/new.py
TobiasKadelka: To shame me even more, it also should have better been an f-string.

I added the full row, thanks.

aqw (approved):
This PR looks good. I have a few comments here-and-there, but no large objections to the overall approach.

As a general comment, this would have been much easier to review if it was split into multiple commits. I would have liked to see smaller features added (such as `templates`, `valid_name`, etc), and then the main `new` stuff.

Overall, LGTM. And thanks for working on this! It's a /big/ improvement. :-)

onyo/lib/onyo.py
```diff
@@ -833,6 +847,65 @@ def generate_faux_serials(self, length: int = 6, num: int = 1) -> set[str]:
...
+        Returns True for valid asset names, and False if invalid.
+        """
+        # verify that a dot exists
+        if "." not in asset.name:
+            log.error("Asset names must have a '.'")
+            return False
+
+        # split fields
+        try:
+            # reverse order, because re reads from left to right and splits,
```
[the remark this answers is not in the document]
TobiasKadelka: Great point, I corrected it 👍

onyo/lib/onyo.py
```diff
@@ -833,6 +847,65 @@ def generate_faux_serials(self, length: int = 6, num: int = 1) -> set[str]:
...
+        for template in self.templates:
+            if template.name == template_name.name:
+                return template.relative_to(self.root)
+            elif template_name.is_file() and template.samefile(template_name):
+                return template.relative_to(self.root)
+
+        log.error(f"Template {template_name} does not exist.")
+        raise ValueError(f"Template {template_name} does not exist.")
+
+    def valid_name(self, asset: Path) -> bool:
```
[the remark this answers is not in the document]
TobiasKadelka: I will change this not now, but with issue #259 in the next PR, which will be about `onyo get` and getting a list of assets, too.

PR #246, opened 2022-11-01
NF/TST: add `repo(init=bool)`, and assorted test modernization

aqw:
This PR covers a variety of areas:

Notable changes:
- add `Repo(init=True)` with tests
- port `init` and test helpers to use `Repo(init=True)`
- drop GitPython as a dependency
- `generate_faux_serials()` will only generate > 0 serials
- expand and improve tests for `generate_faux_serials()`
- update tests for `history` and `cat` to use the `repo` fixture (and some other general modernization)

The test runtime is significantly reduced (~50%). I expected an improvement, but the extent is surprising. The new `init` codepath spawns fewer subshells, and GitPython was always slow in our benchmarks.

Locally:
- without coverage: 73 sec -> 38 sec (48% drop)
- with coverage: 120 sec -> 50 sec (59% drop)

On CI:
- `3.9`: 246 sec -> 110 sec (55% drop)
- `3.10`: 252 sec -> 123 sec (51% drop)
- `3.11`: 393 sec -> 320 sec (19% drop)

### DIFF
--- a/onyo/commands/init.py
+++ b/onyo/commands/init.py
@@ -1,60 +1,13 @@
-import logging
 import sys
-import shutil
 from pathlib import Path
-import git
-
-logging.basicConfig()
-log = logging.getLogger('onyo')
+from onyo import Repo
 
 
-def get_skel_dir():
-    """
-    Return the path of the skel/ dir in the onyo module directory.
-    """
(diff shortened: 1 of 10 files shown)


TobiasKadelka:
Looks good :)

PR #257, opened 2022-11-25
RF `onyo set` and add `onyo unset`

### LINKED ISSUE #4: design `onyo unset`
This is for removing keys entirely.

This is functionally the same as `onyo set key=[unset] asset_file`.

### LINKED ISSUE #112: Add `onyo set --rename` flag
This was not implemented with `onyo set`, because first it is needed to define name schemes and the pseudo keys in an accessable way (see #101).

After implementing, this information has to be added to the readme (it was removed from the specification readme):
```
`onyo set [--rename, -r]`

  The `type`, `make`, `model`, and `serial` pseudo-keys (see "File Contents")
  can be set when the `--rename` flag is used. It will result in the file(s)
  being renamed.

  - `--rename`: Permit assigning values to pseudo-keys that would result in the
    file(s) being renamed.
```

### LINKED ISSUE #249: `set` key/value arguments should be space separated
`onyo set` accepts an arbitrary number of key/value pairs, and an arbitrary number of paths.

To work around this awkwardness, we chose to pass key/value pairs as one long argument separated with commas. This, unfortunately, creates its own wide set of awkwardness (see #116, #240).

For example:
```
onyo set key1=value1,key2=value2 path1 path2
```

The most comprehensive solution is to tell parser that we'll do our own parsing. Go through all key/value pairs until we find a non-pair. After that, it's all paths. If another key/value pair is found, throw and error and exit, informing the user that it's not allowed to mix the two.

For example:
```
onyo set key1=value1 key2=value2 path1 path2
```

The method needs to be flexible enough to work with both `onyo set` and `onyo new --set`.

TobiasKadelka:
Code changes:
- refactor set.py:
  - modernizes code (e.g. docstrings, pyre, Repo class)
  - add flag `onyo set --rename`
  - moves most functionalities to Repo class (e.g. read/write)
  - updates `new` and `edit` to use the new/moved code
- adds tests to `onyo set`
- Updates the existing tests and reference output
- move code `test_asset_path_and_name_available()` from onyo new
  to repo class, to make it available for set
- adds command `unset`
- adds tests for `onyo unset`

User-visible:
- allow `onyo set` setting pseudo fields to rename files (close #112)
- Adds command `onyo unset` to remove keys from assets (close #4)
- key value pairs are now space separated (close #249)

Missing for set/unset:
- the used ruamel functions allow writing of lists/dicts, and the onyo lib (especially the read/write functions) can handle these data-types, too. But the command line interface for the user has currently no way of allowing lists to be set/unset or otherwise handled. This needs to be handled separately, after the syntax for lists and dicts is defined.
- flag `--filter` is not added in this PR, it will happen in the next one, together with `onyo get` and tests.


aqw:
Please rebase this on top of `master`, and then I'll review it. :-)

onyo/lib/onyo.py
```diff
@@ -906,6 +930,227 @@ def valid_name(self, asset: Path) -> bool:
...
+                              [str(file) for file in self.files_staged])
+                # reset renaming needs double-restoring
+                if self.files_staged:
+                    self._git(['restore', '--source=HEAD', '--staged', '--worktree'] +
+                              [str(file) for file in self.files_staged])
+                raise ValueError(e)
+
+        return self._diff_changes()
+
+    def _select_assets_from_directory(self, directory: Path, depth: Union[int, None],
```
aqw, line 978: The only consumer of this is `_set_sanitize()`, yet there is a significant amount of overlap of functionality and checks between the two functions.

I have not thought about this at length, but my impulse is to consolidate these. Perhaps getting a list of all assets/directories first, and /then/ checking them?

onyo/lib/onyo.py
```diff
@@ -906,6 +930,227 @@ def valid_name(self, asset: Path) -> bool:
...
+                continue
+
+            if full_path.is_dir():
+                paths_to_set += self._select_assets_from_directory(full_path, depth, recursive)
+            else:
+                paths_to_set.append(full_path)
+
+        if error_path_absent:
+            log.error('The following paths do not exist:\n' +
+                      '\n'.join(error_path_absent) + '\n' +
```
aqw, line 1078: This will fail on Paths. I recommend using `self._n_join`. It'll take care of casting and iterating, and allow you to avoid quite a few bugs.

onyo/lib/onyo.py
```diff
@@ -1218,3 +1218,32 @@ def _rm_sanitize(self, paths: Iterable[Union[Path, str]]) -> list[Path]:
...
+        if any([key in ["type", "make", "model", "serial"] for key in keys]):
+            log.error("Can't unset pseudo keys (name fields are required).")
+            raise ValueError("Can't unset pseudo keys (name fields are required).")
+
+        for asset in assets_to_unset:
+            contents = self._read_asset(asset)
+
+            for field in keys:
+                if field not in contents:
+                    # TODO: Should this error if field does not exist?
```
aqw, line 1241: No, it should not error. Otherwise, the user would /always/ need to use `filter` in order to constrain the asset list to only those which have the key.

onyo/lib/onyo.py
```diff
@@ -1218,3 +1218,32 @@ def _rm_sanitize(self, paths: Iterable[Union[Path, str]]) -> list[Path]:
...
+
+        if any([key in ["type", "make", "model", "serial"] for key in keys]):
+            log.error("Can't unset pseudo keys (name fields are required).")
+            raise ValueError("Can't unset pseudo keys (name fields are required).")
+
+        for asset in assets_to_unset:
+            contents = self._read_asset(asset)
+
+            for field in keys:
+                if field not in contents:
```
aqw, line 1240: Nit: IMO, it is better to lead with the positive. 

Here you have "if not" making "else" the true statement. It's a bit backwards and can be confusing.

onyo/lib/onyo.py
```diff
@@ -1218,3 +1218,32 @@ def _rm_sanitize(self, paths: Iterable[Union[Path, str]]) -> list[Path]:
...
+                    # TODO: Should this error if field does not exist?
+                    log.info(f"Field {field} does not exist in {asset}")
+                else:
+                    contents.pop(field)
```
aqw, line 1244: `pop()` throws `KeyError` if the key does not exist. You could skip the `if`/`else` and use `try`/`except`.

Arguably, `del contents[field]` is a better fit (also with `try`/`except`), as it does not return a value to operate on (as `pop()` does).

onyo/lib/onyo.py
```diff
@@ -906,6 +930,227 @@ def valid_name(self, asset: Path) -> bool:
...
+            self._git(["mv", str(asset), str(new_name)])
+
+    def _write_asset(self, asset: Path, contents: dict) -> None:
+        """
+        Write contents into an asset file.
+        """
+        if contents == {}:
+            asset.open('w').write("")
+        else:
+            yaml = YAML(typ='rt')
```
aqw, line 1151: Does round-trip mode work here? The object isn't shared across `_read_asset()` and `_write_asset()`...

aqw (changes requested):
I've reviewed about half of this PR. Specifically about half of the `set` commit and all of `unset` (minus the tests).

onyo/lib/onyo.py
TobiasKadelka: This function might also be re-usable for later commands like `onyo get`, and depending on how we implement `onyo ls`.

Alternatively, depending on how we solve https://github.com/psyinfra/onyo/issues/259 this might also have an impact of the code of both functions `_set_sanitize()` and `_select_assets_from_directory()`, since the functionality would be similar.

This is not to say it should never be changed again, but I think it might be more valueable as a separate part for now and be changed then together with the other issue.

PR #267, opened 2023-01-23
DEMO: add example workflow to build an inventory

TobiasKadelka:
Add a list of onyo commands to `generate_demo_repo.sh`, and `inventory.tsv` for the command `onyo new -y --tsv inventory.tsv`


aqw:
LGTM!

PR #271, opened 2023-01-23
NF: add function `Repo.restore()`

### LINKED ISSUE #258: Add function `Repo.restore()`
Multiple commands can make changes, that might be discarded later in the commands execution (e.g. `onyo set` if user wants to discard changes after seeing diff output) and do the restoring with:
`self._git(['restore', '--source=HEAD', '--staged', '--worktree'] + [str(file) for file in self.files_staged])`

Since this is multiple times used in the code, I think this should happen in the repository class. I am a bit divided if this function should get a list of assets, and just restore these, or completely restore all changes, because when assets get renamed and changed with one command, they need to be restored twice.

The downside of restoring all staged changes would be, that if the repo is not clean at the begin of the commands execution, this might reset other staged but uncommited changes. On the otherhand, the current approach is to not run commands that can change assets when there are staged changes, so this would not be an issue at all.

TobiasKadelka:
The function restores all committed changes, e.g. if user responds with "no" to promt when asked if assets should be updated.
Close #258

### DIFF
--- a/onyo/commands/edit.py
+++ b/onyo/commands/edit.py
@@ -138,6 +138,5 @@ def edit(args, opdir: str) -> None:
         if request_user_response("Save changes? No discards all changes. (y/n) "):
             repo.commit('edit asset(s).', staged)
         else:
-            repo._git(['restore', '--source=HEAD', '--staged', '--worktree'] +
-                      [str(file) for file in staged])
+            repo.restore()
             print('No assets updated.')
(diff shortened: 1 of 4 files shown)


aqw:
IMO, some of these situations should not do a restore (for example dry-run). However, that conversation is independent of this change, which does improve upon the status quo. I agree that `Repo.restore()` is useful.

Given that this is a public method, I would really prefer to see a few tests.

PR #278, opened 2023-01-25
BF: allow special characters

### LINKED ISSUE #277: BUG (and TST): `onyo new` allows escaped quote for asset names, but `onyo fsck` fails then
Obviously it should not happen, that something is allowed to be entered into onyo, that then leads to every following command to break, and forces the user to fix manually with non-onyo commands.
As an example: `onyo new laptop_mac\"book_macbook.faux` has an escaped quote and works, and fsck (which is run by every other command in the beginning) fails then:

```
(onyo) tkadelka in ~/new-onyo/test on git:master
❱ onyo new laptop_mac\"book_macbook.faux  
The following will be created:
"laptop_mac\"book_macbook.fauxOQI6r9"
Create assets? (y/n) y
(onyo) tkadelka in ~/new-onyo/test on git:master
❱ onyo fsck
  File "/Users/tkadelka/.venvs/onyo/bin/onyo", line 8, in <module>
    sys.exit(main())
  File "/Users/tkadelka/.venvs/onyo/lib/python3.9/site-packages/onyo/main.py", line 603, in main
    args.run(args, args.opdir)
  File "/Users/tkadelka/.venvs/onyo/lib/python3.9/site-packages/onyo/commands/fsck.py", line 24, in fsck
    repo.fsck()
  File "/Users/tkadelka/.venvs/onyo/lib/python3.9/site-packages/onyo/lib/onyo.py", line 367, in fsck
    if not all_tests[key]():
  File "/Users/tkadelka/.venvs/onyo/lib/python3.9/site-packages/onyo/lib/onyo.py", line 468, in _fsck_yaml
    YAML(typ='rt').load(Path(self.root, asset))
  File "/Users/tkadelka/.venvs/onyo/lib/python3.9/site-packages/ruamel/yaml/main.py", line 430, in load
    with stream.open('rb') as fp:
  File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 1252, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
  File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 1120, in _opener
    return self._accessor.open(self, flags, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/tkadelka/new-onyo/test/"laptop_mac\\"book_macbook.fauxOQI6r9"'
```

I strongly assume, that this happens for almost all commands and probably multiple special characters, that can be escaped, but later not handled by onyo.

To fix this, we have to go trough all commands, probably just setting asset names in quotes in the code, especially when subprocesses are called, and add test-cases to all of them to make sure that onyo can handle these characters too.

Alternatively, if we do not want to allow these characters, it should be explicitely decided, stated in the docs, and then inputs should be checked for special characters, so that they will not be entered.

TobiasKadelka:
This PR fixes two bugs, both connected to certain special characters.

- Parsing of `git ls-files` output contains quotes if files have some special characters:
  - Use `git ls-files -z` to get list of files in `Repo._get_files()` without additional quotes
  - Update parsing of this list accordingly
  - close #277 

- `os.system` parses paths with spaces, quotes etc. differently then `subprocess.run`
  - replace `os.system` with `subprocess.run` for `onyo tree`
  - use `shlex.quote()` for giving paths as input to `os.system`

### DIFF
--- a/onyo/commands/edit.py
+++ b/onyo/commands/edit.py
@@ -1,10 +1,11 @@
 import logging
 import os
 import sys
-
-from onyo import Repo, OnyoInvalidRepoError
 from pathlib import Path
+from shlex import quote
+
 from ruamel.yaml import YAML, scanner  # pyre-ignore[21]
+from onyo import Repo, OnyoInvalidRepoError
 
 logging.basicConfig()
 log = logging.getLogger('onyo')
@@ -42,8 +43,7 @@ def edit_asset(editor: str, asset: Path) -> bool:
     if the user wants to discard the changes.
     """
     while True:
-        os.system(f'{editor} "{asset}"')
-
+        os.system(f'{editor} {quote(str(asset))}')
         try:
             YAML(typ='rt').load(asset)
             # TODO: add asset validity here
--- a/onyo/commands/history.py
+++ b/onyo/commands/history.py
@@ -3,6 +3,7 @@ import os
 import shutil
 import sys
 from pathlib import Path
+from shlex import quote
 
 from onyo import Repo, OnyoInvalidRepoError
 
@@ -84,7 +85,7 @@ def history(args, opdir):
     orig_cwd = os.getcwd()
     try:
         os.chdir(opdir)
-        status = os.system(f"{history_cmd} '{path}'")
+        status = os.system(f"{history_cmd} {quote(str(path))}")
     except:  # noqa: E722
         pass
     finally:
--- a/onyo/commands/tree.py
+++ b/onyo/commands/tree.py
@@ -1,5 +1,5 @@
 import logging
-import os
+import subprocess
 import sys
 from pathlib import Path
 
@@ -9,7 +9,7 @@ logging.basicConfig()
 log = logging.getLogger('onyo')
 
 
-def sanitize_directories(directories: list, opdir: str) -> str:
+def sanitize_directories(directories: list, opdir: str) -> list[str]:
     """
     Check a list of directories. If any do not exist or are a file, and error
     will be printed.
@@ -32,7 +32,7 @@ def sanitize_directories(directories: list, opdir: str) -> str:
               file=sys.stderr)
         sys.exit(1)
 
-    return ' '.join(str(x) for x in dirs)
+    return dirs
 
 
 def tree(args, opdir: str) -> None:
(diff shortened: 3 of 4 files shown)


onyo/commands/history.py
```diff
@@ -4,6 +4,7 @@
 import sys
 from pathlib import Path
 
+from shlex import quote
```
aqw, line 7: nit: `shlex` is in the standard library, and thus should be up one line (underneath `pathlib`). The newline belongs between stdlib and other modules.

onyo/commands/tree.py
```diff
@@ -53,7 +54,8 @@ def tree(args, opdir: str) -> None:
     orig_cwd = os.getcwd()
     try:
         os.chdir(opdir)
-        status = os.system(f"tree {dirs}")
+        ret = subprocess.run(['tree'] + dirs, capture_output=True, text=True)
```
aqw, line 57: This should use `subprocess`'s `cwd`. Then `os` can be removed entirely, and the `chdir` dance can be avoided.

onyo/lib/onyo.py
```diff
@@ -126,7 +126,7 @@ def _get_files(self) -> set[Path]:
         Return a set of all files in the repository (except under .git).
         """
         log.debug('Acquiring list of files')
-        files = {Path(x) for x in self._git(['ls-files']).splitlines() if x}
+        files = {Path(x) for x in self._git(['ls-files', '-z']).split('\0') if x}
```
aqw, line 129: Doesn't this fix also need to be applied to the other `_get_files_*()` functions?

aqw (changes requested):

onyo/commands/edit.py
```diff
@@ -5,6 +5,7 @@
 from onyo import Repo, OnyoInvalidRepoError
 from pathlib import Path
 from ruamel.yaml import YAML, scanner  # pyre-ignore[21]
+from shlex import quote
```
aqw, line 8: nit: this should be above `ruamel` and  under `pathlib`. (and `onyo` should be at the bottom)

aqw (approved):
LGTM!

Issue #253, opened 2022-11-02
TST: retire `clean_sandboxes()` and `change_cwd_to_sandbox()`


aqw:
Depends on #251 and #252

Issue #251, opened 2022-11-02
TST: all tests should be independent


aqw:
Each test should run in its own isolated repository, independent of all other tests. The `repo` fixture should be used. When that is not appropriate. `tmp_path` should be used.

PR #280, opened 2023-01-27
TST: modernize `test_new.py` and `test_mkdir.py` and add special character test cases to them

TobiasKadelka:
This modernizes the tests for `onyo mkdir` and `onyo new`.

Since `test_mkdir.py` was fairly old, I now introduce fixtures, doc-strings and other modernizations to it.

I use in `test_new.py` now `@pytest.mark.repo_dirs(*directories)`, instead of listing all directories to be used in the test case separately.

The PR #278 did introduce a BF that stops onyo from breaking when special characters are used in asset or directory names, so that `onyo mkdir` and `onyo new` now both have additional tests for special characters.

### DIFF
--- a/tests/commands/test_mkdir.py
+++ b/tests/commands/test_mkdir.py
@@ -1,147 +1,147 @@
 import subprocess
 from pathlib import Path
 
+import pytest
+from onyo.lib import Repo
 
-def anchored_dir(directory):
+
+directories = ['simple',
+               's p a c e s',
+               's p a/c e s',
+               'overlap/one',
+               'overlap/two',
+               'overlap/three',
+               'r/e/c/u/r/s/i/v/e',
+               'spe\"cial\\char\'actஞers',
+               'very/very/very/deep'
+               ]
+
+
+@pytest.mark.parametrize('directory', directories)
+def test_mkdir(repo: Repo, directory: str) -> None:
     """
-    Returns True if a directory exists and contains an .anchor file.
-    Otherwise it returns False.
+    Test that `onyo mkdir <dir>` creates new directories correctly for different
+    depths and directory names.
     """
-    if Path(directory).is_dir() and Path(directory, '.anchor').is_file():
-        return True
+    ret = subprocess.run(['onyo', 'mkdir', directory], capture_output=True, text=True)
 
-    return False
-
-
-def test_onyo_init():
-    ret = subprocess.run(["onyo", "init"])
+    # verify output
+    assert not ret.stdout
+    assert not ret.stderr
     assert ret.returncode == 0
 
+    # verify folders and anchors exist
+    d = Path(directory)
+    while not d.samefile(repo.root):
+        assert Path(d).is_dir()
+        assert Path(d, ".anchor").is_file()
+        d = d.parent
 
-def test_simple_dir():
-    ret = subprocess.run(["onyo", "mkdir", "simple/"])
+    # verify that the repository is clean
+    repo.fsck()
+
+
+def test_mkdir_multiple_inputs(repo: Repo) -> None:
+    """
+    Test that `onyo mkdir <dirs>` creates new directories all in one call when
+    given a list of inputs.
+    """
+    ret = subprocess.run(['onyo', 'mkdir'] + directories, capture_output=True, text=True)
+
+    # verify output
+    assert not ret.stdout
+    assert not ret.stderr
     assert ret.returncode == 0
-    assert anchored_dir('simple')
+
+    # verify folders and anchors exist
+    for directory in directories:
+        d = Path(directory)
+        while not d.samefile(repo.root):
+            assert Path(d).is_dir()
+            assert Path(d, ".anchor").is_file()
+            d = d.parent
+
+    # verify that the repository is clean
+    repo.fsck()
 
 
-def test_dir_exists():
-    ret = subprocess.run(["onyo", "mkdir", "simple/"])
-    assert anchored_dir('simple')
+@pytest.mark.repo_dirs(*directories)
+@pytest.mark.parametrize('directory', directories)
+def test_error_dir_exists(repo: Repo, directory: str) -> None:
+    """
+    Test the correct error behavior when `onyo mkdir <path>` is called on an
+    existing directory name.
+    """
+    ret = subprocess.run(['onyo', 'mkdir', directory], capture_output=True, text=True)
+
+    # verify output
+    assert not ret.stdout
+    assert "The following paths already exist:" in ret.stderr
     assert ret.returncode == 1
 
+    assert Path(directory).is_dir()
+    assert Path(directory, ".anchor").is_file()
 
-def test_dir_exists_as_file():
-    ret = subprocess.run(["onyo", "mkdir", "simple/.anchor"])
+    # verify that the repository is clean
+    repo.fsck()
+
+
+@pytest.mark.repo_files(*directories)  # used as files to test errors
+@pytest.mark.parametrize('file', directories)
+def test_dir_exists_as_file(repo: Repo, file: str) -> None:
+    """
+    Test the correct error behavior when `onyo mkdir <file>` is called on files.
+    """
+    ret = subprocess.run(['onyo', 'mkdir', file], capture_output=True, text=True)
+
+    # verify output
+    assert not ret.stdout
+    assert "The following paths already exist:" in ret.stderr
     assert ret.returncode == 1
 
+    assert Path(file).is_file()
 
-def test_dir_protected():
-    # dir named .anchor
-    ret = subprocess.run(["onyo", "mkdir", ".anchor"], capture_output=True, text=True)
-    assert ret.returncode == 1
+    # verify that the repository is clean
+    repo.fsck()
+
+
+protected_paths = [".anchor",
+                   "simple/.git",
+                   "simple/.onyo",
+                   ".git/nope"
+                   ".onyo/nope",
+                   ]
+@pytest.mark.repo_dirs("simple")
+@pytest.mark.parametrize('protected_path', protected_paths)
+def test_dir_protected(repo: Repo, protected_path: str) -> None:
+    """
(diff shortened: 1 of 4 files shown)


tests/commands/test_new.py
```diff
@@ -36,9 +27,28 @@ def test_new(repo: Repo, directory: str) -> None:
...
+    """
+    Test that `onyo new <path>` can create an asset on the top level of the
+    repository.
+    """
+    file = 'laptop_apple_macbookpro.0'
+    ret = subprocess.run(['onyo', 'new', file], input='y', capture_output=True, text=True)
+
+    # verify correct output
+    assert "The following will be created:" in ret.stdout
+    assert "Create assets? (y/n) " in ret.stdout
```
aqw, line 51: IMO, this check should be its own test for interactive mode.

tests/commands/test_mkdir.py
```diff
@@ -1,147 +1,144 @@
...
 
-    return False
+    # verify output
+    assert not ret.stdout
+    assert not ret.stderr
+    assert ret.returncode == 0
 
+    # verify folders and anchors exist
+    d = Path(directory)
+    while not d.samefile(repo.root):
```
aqw, line 35: An alternative is to loop over `Path().parents`.

tests/commands/test_mkdir.py
```diff
@@ -1,147 +1,144 @@
...
+    ret = subprocess.run(['onyo', 'mkdir'] + directories, capture_output=True, text=True)
+
+    # verify output
+    assert not ret.stdout
+    assert not ret.stderr
     assert ret.returncode == 0
-    assert anchored_dir('simple')
 
+    # verify folders and anchors exist
+    for d in directories:
```
aqw, line 57: This will only test the deepest folder, and none of the parents.

tests/commands/test_mkdir.py
```diff
@@ -1,147 +1,147 @@
+import pytest
```
aqw, line 1: nit: this is a third party import and should be after the standard lib ones.

tests/commands/test_new.py
```diff
@@ -5,28 +5,41 @@
...
+    # verify that the new asset exists and the repository is in a clean state
+    assert len(repo.assets) == 1
+    repo.fsck()
+
+
+@pytest.mark.repo_dirs(*directories)
+@pytest.mark.parametrize('directory', directories)
+def test_new_interactive(repo: Repo, directory: str) -> None:
     """
     Test that `onyo new` can create an asset in different directories.
```
aqw, line 44: This docstring is identical to the one for `test_new()`, yet they focus on testing different things.

tests/commands/test_new.py
```diff
@@ -5,28 +5,41 @@
...
+    assert ret.returncode == 0
+    assert Path(file).exists()
+
+    # verify that the new asset exists and the repository is in a clean state
+    assert len(repo.assets) == 1
+    repo.fsck()
+
+
+@pytest.mark.repo_dirs(*directories)
+@pytest.mark.parametrize('directory', directories)
```
aqw, line 41: If I understand it correctly, the goal here is to test whether `onyo new` prompts for user interaction. If so, then I don't see a need to test the prompt across different directories, as that does not impact the logic of whether a prompt would be spawned or not.

tests/commands/test_new.py
```diff
@@ -36,9 +49,27 @@ def test_new(repo: Repo, directory: str) -> None:
...
+    assert Path(file).exists()
+
+    # verify that the new asset exists and the repository is in a clean state
+    assert len(repo.assets) == 1
+    repo.fsck()
 
-    # double-casting needed to remove ./ in beginning
-    assert str(Path(file)) in ret.stdout
+
+def test_new_top_level(repo: Repo) -> None:
```
aqw, line 62: This looks like a single-version case of `test_new()`. It should be possible to cover this special-case by adding `'./'` or `''` to the list of `directories` to test.

tests/commands/test_new.py
```diff
@@ -56,12 +87,11 @@ def test_folder_creation_with_new(repo: Repo, directory: str) -> None:
     Tests that folders are created when a sub-folder already exist, too.
```
aqw, line 87: Commenting here, because GitHub won't let me comment higher... `overlap` is unused by this test and can be removed.

```
@pytest.mark.repo_dirs('overlap')
```

aqw (commented):
I reviewed `test_mkdir.py` and `test_new.py`. Broadly, it looks good. My comments are largely nits or easy simplifications. 

I will have to review the new `set` and `unset` tests later today or tomorrow.

tests/commands/test_set.py
```diff
@@ -94,11 +114,10 @@ def test_set_with_dot_recursive(repo: Repo, set_values: list[str]) -> None:
     Test that when `onyo set KEY=VALUE .` is called from the repository root,
```
aqw, line 114: Out of scope of the current PR, but if you end up updating this PR, perhaps you can fold these in:
- I think this docstring is wrong. It should call from `cwd`, not repo root. Buuuuuuuut
- I don't think this test is necessary. `.` is a dir, just like any other. So this can be removed.

aqw (commented):
I have finished reviewing the tests for `set` and `unset`. My only comments are nits.

Once the fixes from this and the previous review land, this can be merged. :-)

tests/commands/test_new.py
TobiasKadelka: I find adding either one of these to lead to different problems, which are just solvable when doing much more casting between strings and Paths. If that is prefered, I can change this, but it will effect all tests using `directories`, and I am not sure if it actually looks better.

tests/commands/test_new.py
TobiasKadelka: It is used, because the folder `overlap` is the only case in which then a subfolder gets created, were a directory already exists, i.e. what is meant with the docstring "Tests that folders are created when a sub-folder already exist, too."

However I can add a comment to make this more obvious.

tests/commands/test_set.py
TobiasKadelka: It is logically/functionally equivalent, but just if it works. But I think, especially since we are still changing the way argparser works, it is worth testing to make sure that this gets correctly interpreted.

Also, in the past we had cases, were the onyo root was not part of the `Repo.dirs`, which should be tested else, but still was found through a test like this. So, as soon as we work on fixing the Problems with working from a subdir, I would rather argument that similar additional tests should be added.

tests/commands/test_new.py
aqw: That surprises me. Can you share the specific error you get?

tests/commands/test_set.py
aqw: > It is logically/functionally equivalent, but just if it works.

`.` is a valid dir for a POSIX filesystem. So I'm missing the situation where it wouldn't work.

> Also, in the past we had cases, were the onyo root was not part of the Repo.dirs, which should be tested else, but still was found through a test like this. 

If there isn't already a test for `Repo.dirs` that checks for this, then it should be added there.

tests/commands/test_new.py
TobiasKadelka: Yes, but it is just an Assertion-error, because "./" gets removed for onyo output, so it needs the double-casting `str(Path(asset)) in ret.stdout` again, which I hate.

For `''` I forgot the error, but it is caused by e.g. `file = f'{directory}/laptop_apple_macbookpro.0'` being then just `/laptop_apple_macbookpro.0`. Again, this can be solved by double-casting in that line, but I do not like either of these solutions.

Though in general I am all for adding it to the directories, which is why I had it like that for the longest time, and just removed the double casting in this PR, so if you know any solution that works nicely without casting, I would be super happy! :)

tests/commands/test_set.py
TobiasKadelka: Thanks, you are right.

That would then be here:
https://github.com/psyinfra/onyo/blob/main/tests/lib/test_Repo.py#L253

I will think about it and either add it to the current tests, or do it together with #253 or #251.

tests/commands/test_new.py
aqw: [Path.relative_to()](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) might be of help.

tests/commands/test_mkdir.py
TobiasKadelka: Looping over `Path().parents` starts then directly with the parent (and not the Path), which is the most important to check, and all the ways to include it does look worse in my opinion.

So I think still this is the best version to reach that functionality.
If you see a way doing it with `Path().parents` that does this better, let me know, because I did not like the ways I came up using it.

tests/commands/test_new.py
TobiasKadelka: I think it is not, because it still needs to use a Path, that then gets casted into a string again for the output check.

The general problem is that we want to use some Paths and some strings in almost all test cases, so in my opinion this generally does never really work with just one variable without some casting.

The only alternative way I could see, is like in the beginning of set: https://github.com/psyinfra/onyo/blob/main/tests/commands/test_set.py
Defining assets/directories already and then building the carthesian product, so that pre-defined strings are given into the function, and then just casting it to Path() when needed. Though I am not sure, if we want to do that in general, because I feel like this creates a lot of unneeded test-cases.

In any case I think the general Path/Test handling should be discussed at some point and a uniform way of doing might be looked for. I remember you wanting to define a matrix for paths that get tested for all commands at some point, and I think in such an effort this should be discussed.

tests/commands/test_set.py
TobiasKadelka: I added a new issue describing the need for the `tests/lib/*.py` updates and will add/solve this when I overhaul them.

tests/commands/test_mkdir.py
aqw: Makes sense. :-)

aqw:
LGTM!

PR #291, opened 2023-02-17
BF: shell-completion fixes

aqw:
This PR includes two fixes:

1) remove both forms of a flag from the suggestion list after its been used (previously, only the form used was removed)
2) prevents shell-completion from exploding with non-int nargs to flags

The fix for 2 is imperfect, but it gets the tests passing again for #282 and #290 

I have added a lengthy comment to the fix for 2. The proper solution is unclear to me, and may require posting to the ZSH mailing list.

Until then, users for query-style commands should do one of the following:
* use `--paths` before `--keys` (which has no idea of completion
* or use another flag (such as `--yes`) after `--paths`, which will then return to normal behavior

### DIFF
--- a/onyo/commands/shell_completion.py
+++ b/onyo/commands/shell_completion.py
@@ -126,9 +126,7 @@ class TabCompletion:
                 # arguments for a flag
                 cmd_tree['flags'][flag_string]['args'] = {}
 
-                # TODO: handle non-int nargs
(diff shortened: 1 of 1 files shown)


TobiasKadelka:
Looks good to me, thanks for looking into this issue and fixxing it! :)

Issue #286, opened 2023-02-02
migrate query-style commands to new, explicit syntax


aqw:
The following commands needs to be updated:
- [ ] `onyo get`
- [ ] `onyo set` (see also #116, #113)
- [ ] `onyo unset`

Niels, Tobias and I discussed the syntax ambiguity problems (#260) with commands which relate to querying fields. We came to the following conclusions:

- space separated is not a viable option. `onyo get` and `onyo unset` do not use `=`, thus it is impossible to determine what is a key vs a path
- comma separated is not an option either, as the shell swallows all quoting before passing to Python. The consequences of this make it impossible to discern between scenarios for commas, quotes, etc. 
- thus the only option is to use explicit flags

Commands will be split into two categories: query-style, and traditional-style (nix-style). Query style commands will use this new syntax. For example:

```
# query-style commands
onyo set --filter type=laptop make=apple --keys this=that thing=thong --paths warehouse retire
onyo get --filter type=laptop make=apple --keys this that --paths warehouse retire 
onyo unset --filter type=laptop make=apple --keys this that --paths warehouse retire 
```

The rest of the commands will retain their syntax:
```
# traditional-style (nix-style) commands
onyo new warehouse/laptop_apple_macbook.serial
onyo tree warehouse
onyo mkdir warehouse/shelf
onyo rm warehouse/shelf
onyo mv here there
```

TobiasKadelka:
I just assumed that `--keys` and `--paths` means also `-k` and `-p`, so I checked that these do not exist with another meaning for other commands (they don't) and started to implement them, too.

TobiasKadelka:
The syntax migration was completed with the implementation/merge of `onyo get`.
The same style of syntax was also added to the command `onyo new` so as to keep onyo consistent, because we allow setting key/values with it, too.

So I close this issue :)

PR #293, opened 2023-02-22
CLI: new `onyo new` syntax

### LINKED ISSUE #287: add `onyo new --set`
The relevant portions of #260 have been moved here:

We want `onyo new --set key=value asset` to have the optional flag `--set`, but also the possibility to not use it, while at least one (possibly multiple) asset(s) are needed, except when a `--tsv tsv.file` is specified, which disallows giving assets via command line.

It is unclear to me if this is possible without adding `--path` to `onyo new` (see #286), which I'm not thrilled about.

TobiasKadelka:
This mainly updates the argparsing for `onyo new` to use the new style of CLI syntax and flags like described in issue #286.

Updates argparsing and removes unneeded code of class `StoreKeyValuePairsAndAssetsSeparately` and uses instead `StoreKeyValuePairs` for handling key-value pairs like `onyo set`. Together with this, the tests for `onyo new` and the demo are updated, to use the new syntax everywhere.

Updates code of `onyo new` to work with the new arguments passed into it.

Close #287


onyo/main.py
```diff
@@ -376,9 +335,16 @@ def setup_parser():
         help='Open new assets to edit them before creation'
     )
     cmd_new.add_argument(
-        'asset',
+        '-k', '--keys',
+        required=False,
+        action=StoreKeyValuePairs,
+        metavar="KEYS",
+        nargs='*',
```
aqw, line 342: `--keys` should require at least one arg, thus `+` should be used.

https://docs.python.org/3/library/argparse.html#nargs

tests/commands/test_new.py
```diff
@@ -260,7 +260,9 @@ def test_new_with_flags_edit_set_template(repo: Repo, directory: str) -> None:
     set_values = "mode=set"
 
     # create asset with --edit and --template
-    ret = subprocess.run(['onyo', 'new', '--yes', '--edit', '--template', template, set_values, str(asset)], capture_output=True, text=True)
+    ret = subprocess.run(['onyo', 'new', '--yes', '--edit',
+                          '--template', template, '--keys', set_values,
```
aqw, line 264: `set_values` feels like a leftover from before, and should be renamed.

tests/commands/test_new.py
```diff
@@ -260,7 +260,9 @@ def test_new_with_flags_edit_set_template(repo: Repo, directory: str) -> None:
     set_values = "mode=set"
```
aqw, line 260: This seems like a leftover from before the rename to `--keys`.

Also, as I cannot comment higher up:
- the docstring is no longer accurate
- the function name should be changed (as you have with others)
  - I won't check all function names; please just search for `set`, etc

tests/commands/test_new.py
```diff
@@ -260,7 +260,9 @@ def test_new_with_flags_edit_set_template(repo: Repo, directory: str) -> None:
...
 
     # create asset with --edit and --template
```
aqw, line 262: And `--keys`

tests/commands/test_new.py
```diff
@@ -437,9 +440,10 @@ def test_tsv_with_flags_template_set_edit(repo: Repo) -> None:
     assert table_path.is_file()
 
     # create assets with table
-    ret = subprocess.run(['onyo', 'new', '--yes', '--edit', set_values,
-                          '--tsv', table_path, '--template', template],
-                         capture_output=True, text=True)
+    ret = subprocess.run(['onyo', 'new', '--yes', '--edit',
+                          '--keys', set_values, '--tsv', table_path,
+                          '--template', template], capture_output=True,
```
aqw, line 445: Nit: when line-splitting, I prefer that it take into account context. It would make sense if the final line was
```
```

aqw (changes requested):
Overall LGTM. There's a be missed `set`s in the tests. And `nargs` should be fixed. Otherwise, it's looking good. :-)

onyo/main.py
TobiasKadelka: Oh dang, you are right.

aqw (approved):
This PR looks good to me, and I approve of it being merged.

The only change which would be nice to see (but not a blocker) is an additional test where `--keys` overwrites a value already set in the template.

TobiasKadelka:
That sounds reasonable. I will add it and merge on Monday then, thanks!

TobiasKadelka:
I added to `tests_new.py` the test `test_new_with_keys_overwrite_template()` and will merge :)

PR #294, opened 2023-02-23
NF: Add flags, change commit messages, update tests

### LINKED ISSUE #90: Add option `-m` to commands to change default commit message
Right now, commiting onyo commands have set commit messages (e.g. `delete asset(s)` or `initialize onyo repository`).
It should be possible to add a flag -m to these commands to overwrite the default message.

### LINKED ISSUE #295: Defining commit message headers
The commit message headers should be consistent and give as much helpful information as possible for real use cases while acknowledging that we cannot always show _all information at once_. 

I focus on the commit message headers (the bodies can obviously contain all the information in the world).
Since 80 characters cannot do to much, the broader questions are "What information can we live without?", "What maximizes the value of the 80 characters?".

**1. "Full style"**

Every commit starts with the command name followed by `: `, and then in quotes the full name+path of assets modified, moved etc. This is close to the current style, but not all of these are actually implemented.

**2. "Shortened style"**

Instead of trying to present all information (which is not possible) we give "help to differentiate" between commits, to make it easier to find specifics when scrolling through (e.g. with tig).

I am currently intrigued by the idea to cut out all information that one would know or be able to guess from the rest of the commit message, and I think this "cutting out" idea will improve the visibility of information presented greatly, e.g. when changing contents, the assets are oftentimes more important than the paths, and when moving assets to a user just the `type` field is relevant.

------------

**Examples (currently):**
```
mkdir: 'ethics/Max Mustermann','ethics/Achilles Book'
new: 'warehouse/monitor_dell_PH123.86JZho','warehouse/laptop_apple_macbook.oiw629'
rm: 'admin/Karl Krebs/','ethics/Achilles Book/laptop_lenovo_thinkpad.owh8e2'
edit: 'warehouse/laptop_apple_macbook.9r5qlk','ethics/Max Mustermann/laptop_apple_macbook.9r32he'
mv: 'warehouse/laptop_apple_macbook.9r32he','repair/headphones_apple_airpods.7h8f04' -> 'ethics/Max Mustermann'
set: 'ethics/Max Mustermann/laptop_apple_macbook.9r32he','warehouse/laptop_apple_macbook.9r5qlk'
unset: 'ethics/Max Mustermann/laptop_apple_macbook.9r32he','warehouse/laptop_apple_macbook.9r5qlk'
```

**Examples (short):**
```
mkdir: 'ethics/Max Mustermann','ethics/Achilles Book'
new: 'monitor_dell_PH123.86JZho','laptop_apple_macbook.oiw629'
rm: 'Karl Krebs/','laptop_lenovo_thinkpad.owh8e2'
edit: 'laptop_apple_macbook.9r5qlk','laptop_apple_macbook.9r32he'
mv: 'laptop','headphones' -> 'Max Mustermann'
set: (RAM,USB) 'laptop_apple_macbook.9r32he','laptop_apple_macbook.9r5qlk'
unset: (RAM,USB) 'laptop_apple_macbook.9r32he','laptop_apple_macbook.9r5qlk'
```

- `mkdir`: the whole path(s)
- `new` and `edit`: just asset(s) without path
- `rm`: always the highest level (either the asset deleted, or the highest folder like user or group, without complete path)
- `mv`: just the `type` field of source(s), and the highest level of destination
- `set` and `unset`: just highest level (either assets or folders), but additionally with keys set

TobiasKadelka:
This adds flags (`-m` a.k.a. `--message` for all commands, and `--yes` and `--quiet` edit) and tests for the new flags.
The PR also normalizes the default commit messages (like discussed in #295) with the new function `Repo.generate_commit_message()` that automatically builds (and shortens, if needed) commit messages.

While I touched the older tests, I modernized them slightly. These changes are not meant to be a complete overhaul or something like it, but really just uplift them so they are easier to modify in the future (e.g. I removed a lot of hardcoded variables and asset names, and I added parametrization and `pytest.mark.repo_files(*assets)` instead).

Changes:
- adds `--message` flag to all commands except `onyo init` and `onyo config` (close #90)
  - changes are made in `main.py` and for each command
- adds `--yes` and `--quiet` to `onyo edit`
- adds tests for all newly added flags
- changes default commit messages (close #295)
- I modernized some of the tests (just the ones I was working with anyways)
  - I added pyre
  - I made the tests a bit more flexible (e.g. `pytest.mark.repo_files(*assets)`)
  - I added some special character tests, while removing redundant old tests

### DIFF
--- a/onyo/commands/edit.py
+++ b/onyo/commands/edit.py
@@ -137,6 +137,7 @@ def edit(args, opdir: str) -> None:
         print(repo._diff_changes())
         if request_user_response("Save changes? No discards all changes. (y/n) "):
             repo.commit('edit asset(s).', staged)
+            repo.commit(repo.generate_commit_message(message=args.message, cmd="edit"))
         else:
             repo.restore()
             print('No assets updated.')
--- a/onyo/commands/edit.py
+++ b/onyo/commands/edit.py
@@ -108,6 +108,12 @@ def edit(args, opdir: str) -> None:
     and return to the original state.
     """
     repo = None
+
+    # check flags for conflicts
+    if args.quiet and not args.yes:
+        print('The --quiet flag requires --yes.', file=sys.stderr)
+        sys.exit(1)
+
     try:
         repo = Repo(opdir)
         # "onyo fsck" is intentionally not run here.
@@ -129,15 +135,18 @@ def edit(args, opdir: str) -> None:
             If user wants to discard changes, restore the asset's state
             """
             repo._git(['restore', str(asset)])
-            print(f"'{asset}' not updated.")
+            if not args.quiet:
+                print(f"'{asset}' not updated.")
 
     # commit changes
     staged = sorted(repo.files_staged)
     if staged:
-        print(repo._diff_changes())
-        if request_user_response("Save changes? No discards all changes. (y/n) "):
-            repo.commit('edit asset(s).', staged)
-            repo.commit(repo.generate_commit_message(message=args.message, cmd="edit"))
+        if not args.quiet:
+            print(repo._diff_changes())
+        if args.yes or request_user_response("Save changes? No discards all changes. (y/n) "):
+            repo.commit(repo.generate_commit_message(message=args.message,
+                                                     cmd="edit"))
         else:
             repo.restore()
-            print('No assets updated.')
+            if not args.quiet:
+                print('No assets updated.')
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -92,6 +92,97 @@ class Repo:
         self._gitfiles = None
         self._templates = None
 
+    def generate_commit_message(self, message: list[str] = [], cmd: str = "",
+                                keys: list[str] = [], destination: str = "",
+                                MAX_LEN: int = 80) -> str:
+        """
+        Generate a commit subject and body suitable for use with git commit.
+
+        If `message` is given, the function uses the first element of it to
+        generate a message subject, and the following ones are joined for the
+        message body.
+
+        If no `message` is given, generate a default commit message based on
+        the other parameters given, and the `files staged` in the repository.
+        The paths for the message are shortened if needed, so that the resulting
+        message subject does not exceed `MAX_LEN`.
+
+        Adds a list of `changed files` with their path relative to the root of
+        the repository to the body of all commit messages.
+        """
+        message_subject = ""
+        message_body = ""
+        message_appendix = ""
+
+        # staged files and directories (without ".anchor") in alphabetical order
+        staged_changes = [x if not x.name == ".anchor" else x.parent
+                          for x in sorted(self.files_staged)]
+
+        if message:
+            message_subject = message[0][0]
+            message_body = self._n_join([x[0] for x in message[1:]])
+        else:
+            # get variables for the begin of the commit message `msg_dummy`
+            dest = None
+            keys_str = ""
+            if keys:
+                keys_str = f" ({','.join(str(x.split('=')[0]) for x in sorted(keys))})"
+            if destination:
+                dest = Path(self.opdir, destination).relative_to(self.root)
+            if dest and dest.name == ".anchor":
+                dest = dest.parent
+
+            # the `msg_dummy` is the way all automatically generated commit
+            # message headers (independently of later adding/shortening of
+            # information) begin, for all commands.
+            msg_dummy = f"{cmd} [{len(staged_changes)}]{keys_str}"
+            message_subject = self._generate_commit_message_subject(msg_dummy,
+                                                                    staged_changes,
+                                                                    dest, MAX_LEN)
+
+        message_appendix = self._n_join([str(x) for x in staged_changes])
+        return f"{message_subject}\n\n{message_body}\n\n{message_appendix}"
+
+    def _generate_commit_message_subject(self, msg_dummy: str,
+                                         staged_changes: list[Path],
+                                         destination: Optional[Path],
+                                         MAX_LEN: int = 80) -> str:
+        """
+        Generates "commit message subject" with the `msg_dummy` and the paths
+        from `staged_changes` and `destination`, and shortens the paths if the
+        message length exceeds the `MAX_LEN`.
+        """
+
+        # long message: full paths (relative to repo-root)
+        paths_str = ','.join([f"'{x}'" for x in staged_changes])
+        msg = f"{msg_dummy}: {paths_str}"
+        if destination:
+            msg = f"{msg} -> '{destination}'"
+
+        if len(msg) < MAX_LEN:
+            return msg
+
+        # medium message: highest level (e.g. dir or asset name)
+        paths = [x.name for x in staged_changes]
+        paths_str = ','.join(["'{}'".format(x) for x in paths])
+        msg = f"{msg_dummy}: {paths_str}"
+        if destination:
+            msg = f"{msg} -> '{destination.relative_to(destination.parent)}'"
+
+        if len(msg) < MAX_LEN:
+            return msg
+
+        # short message: "type" of devices in summary (e.g.  "laptop (2)")
+        paths = [x.name.split('_')[0] for x in staged_changes]
+        paths_str = ','.join(sorted(["'{} ({})'".format(x, paths.count(x))
+                                     for x in set(paths)]))
+        msg = f"{msg_dummy}: {paths_str}"
+        if destination:
+            msg = f"{msg} -> '{destination.relative_to(destination.parent)}'"
+
+        # return the shortest possible version of the commit message as fallback
+        return msg
+
     def restore(self) -> None:
         """
         Restore all staged files with uncommitted changes in the repository.
--- a/onyo/commands/mkdir.py
+++ b/onyo/commands/mkdir.py
@@ -1,5 +1,4 @@
 import sys
-from pathlib import Path
 
 from onyo import Repo, OnyoInvalidRepoError, OnyoProtectedPathError
 
@@ -23,6 +22,7 @@ def mkdir(args, opdir):
 
     try:
         repo.mkdir(args.directory)
-        repo.commit('mkdir: ' + ','.join(["'{}'".format(Path(opdir, x).resolve().relative_to(repo.root)) for x in args.directory]))
+        repo.commit(repo.generate_commit_message(message=args.message,
+                                                 cmd="mkdir"))
     except (FileExistsError, OnyoProtectedPathError):
         sys.exit(1)
--- a/onyo/commands/mv.py
+++ b/onyo/commands/mv.py
@@ -1,5 +1,4 @@
 import sys
-from pathlib import Path
 
 from onyo import Repo, OnyoInvalidRepoError, OnyoProtectedPathError
 
@@ -42,8 +41,7 @@ def mv(args, opdir: str) -> None:
 
     try:
         repo.mv(args.source, args.destination)
-        repo.commit('mv: ' +
-                    ','.join(["'{}'".format(Path(opdir, x).resolve().relative_to(repo.root)) for x in args.source]) +
-                    ' -> ' + str(Path(opdir, args.destination).resolve().relative_to(repo.root)))
+        repo.commit(repo.generate_commit_message(message=args.message, cmd="mv",
+                                                 destination=args.destination))
     except (FileExistsError, FileNotFoundError, OnyoProtectedPathError, ValueError):
         sys.exit(1)
--- a/onyo/commands/new.py
+++ b/onyo/commands/new.py
@@ -264,7 +264,8 @@ def new(args, opdir: str) -> None:
 
     # commit or discard changes
     if args.yes or request_user_response("Create assets? (y/n) "):
-        repo.commit("new asset(s)", changes)
+        repo.commit(repo.generate_commit_message(message=args.message,
+                                                 cmd="new"))
     else:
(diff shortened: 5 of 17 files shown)


onyo/lib/onyo.py
```diff
@@ -92,6 +92,75 @@ def clear_caches(self) -> None:
         self._gitfiles = None
         self._templates = None
 
+    def generate_commit_message(self, message: str = "", cmd: str = "",
+                                keys: list[str] = [], destination: str = "") -> str:
+        """
+        Returns either the commit message given to onyo by the user, or creates
+        a default commit message.
+        """
+        message_header = ""
```
aqw, line 102: The vocabulary used by the git project is `subject` rather than `header`.

onyo/lib/onyo.py
```diff
@@ -92,20 +92,22 @@ def clear_caches(self) -> None:
...
         message_header = ""
+        message_body = ""
         MAX_LEN = 80
+        assets = [Path(self.opdir, x).relative_to(self.root) for x in sorted(self.files_staged)]
 
-        # TODO: first construct body?
         if message:
-            message_header = message
+            message_header = "".join(message[0])
+            message_body = "\n".join(["".join(x) for x in message[1:]]) + "\n\n"
```
aqw, line 109: `_n_join()` can be useful here.

```
message_body = self._n_join(message[1:]) 
```

The final `+ "\n\n"` should be done later, when it's assembled in context.

onyo/lib/onyo.py
```diff
@@ -92,20 +92,22 @@ def clear_caches(self) -> None:
...
         """
         message_header = ""
+        message_body = ""
         MAX_LEN = 80
+        assets = [Path(self.opdir, x).relative_to(self.root) for x in sorted(self.files_staged)]
 
-        # TODO: first construct body?
         if message:
-            message_header = message
+            message_header = "".join(message[0])
```
aqw, line 108: This can be simply:
```
message_header = message[0]
```

onyo/lib/onyo.py
```diff
@@ -92,20 +92,22 @@ def clear_caches(self) -> None:
...
-        Returns either the commit message given to onyo by the user, or creates
-        a default commit message.
+        The commit message is generated based on arguments for `--message` if
+        given by the user, or a default commit message.
+        Adds a list of changed files to the body of commit messages.
         """
         message_header = ""
+        message_body = ""
         MAX_LEN = 80
+        assets = [Path(self.opdir, x).relative_to(self.root) for x in sorted(self.files_staged)]
```
aqw, line 105: `assets` is not the right name here, as non-assets (such as a config files, directories, etc) may be changed.

`sorted()` should not be necessary, as `git` sorts everything when we query for staged files.

It would be nicer UX (IMO) if the `.anchor` was stripped when directories are created/deleted. So it's simply:
- `subdir/`

onyo/lib/onyo.py
```diff
@@ -159,7 +161,10 @@ def generate_commit_message(self, message: str = "", cmd: str = "",
             if len(msg) < MAX_LEN:
                 message_header = msg
 
-        return message_header
+        # add list of changed files in alphabetical order to message body
+        message_body += "\n".join([str(x) for x in assets])
```
aqw, line 165: `_n_join()` can help here.

onyo/lib/onyo.py
```diff
@@ -159,7 +161,10 @@ def generate_commit_message(self, message: str = "", cmd: str = "",
             if len(msg) < MAX_LEN:
                 message_header = msg
 
-        return message_header
+        # add list of changed files in alphabetical order to message body
```
aqw, line 164: The "alphabetizing" is done elsewhere, so it feels out of place to specify it here.

aqw (changes requested):
Thanks for the PR. Great to see so many of these things addressed.

I've commented on various changes here and there.

The main thing I've left unreviewed is `generate_commit_message()`. It is a large chunk of code, and I am not convinced that it's the right approach. I'm unsure of what the right approach is though, and I will need to reflect on that further.

onyo/lib/onyo.py
```diff
@@ -92,6 +92,75 @@ def clear_caches(self) -> None:
         self._gitfiles = None
         self._templates = None
 
+    def generate_commit_message(self, message: str = "", cmd: str = "",
```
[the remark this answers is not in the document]
TobiasKadelka: I agree, but do not see where I diverge from that rule.
- First, if given, the function uses `message`
- Then, if not, it concatenates `cmd`, `keys`, and lastly `destination` for commands like `mv`.

onyo/commands/edit.py
```diff
@@ -129,15 +135,18 @@ def edit(args, opdir: str) -> None:
...
     # commit changes
     staged = sorted(repo.files_staged)
     if staged:
-        print(repo._diff_changes())
-        if request_user_response("Save changes? No discards all changes. (y/n) "):
-            repo.commit('edit asset(s).', staged)
-            repo.commit(repo.generate_commit_message(message=args.message, cmd="edit"))
+        if not args.quiet:
+            print(repo._diff_changes())
+        if args.yes or request_user_response("Save changes? No discards all changes. (y/n) "):
```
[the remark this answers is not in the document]
TobiasKadelka: It is not capitalized, because there is no default option. If the user response is neither y/n, then it will ask again. If the user wants to not having to answer either, `--yes` should be used.

Not having the "no" option as default, but instead asking again, is choosen so as to prevent the user from editing (or creating new assets) by hand, changing some lines, and then accidently answering `m` instead of `n` and thereby loosing the "progress" made.

This behavior was choosen at the time when we removed the temporary files. We can add a default option again if we decide to, but I want to mention that this is behaving like it does with intention and the message reflects the behavior.

onyo/lib/onyo.py
TobiasKadelka: It can't, because the way argparser does work when doing `action='append'`.
If it is done like`message_subject = message[0]`, the commit message looks like ['word'], e.g. it is still an array.
But instead of `join()`ing them, I changed it now to `message_subject = message[0][0]`, and for the body similarly.

onyo/commands/edit.py
```diff
@@ -129,15 +135,18 @@ def edit(args, opdir: str) -> None:
...
+            repo.commit(repo.generate_commit_message(message=args.message,
```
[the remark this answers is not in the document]
TobiasKadelka: That is not true, the result will either just be `message`, if `-m` is used by the user, or it will be `cmd: staged_files keys -> destination`. Maybe there might be more fitting variable names to reflect this.

onyo/lib/onyo.py
TobiasKadelka: `sorted()` is needed, because `self.files_staged` is a set, so independently of what `git` sorts, the order is not stable.

TobiasKadelka: I remove the anchors now there, and I renamed `assets` to `staged_changes`

onyo/lib/onyo.py
```diff
@@ -92,6 +92,75 @@ def clear_caches(self) -> None:
...
+                                keys: list[str] = [], destination: str = "") -> str:
+        """
+        Returns either the commit message given to onyo by the user, or creates
+        a default commit message.
+        """
+        message_header = ""
+        MAX_LEN = 80
+
+        # TODO: first construct body?
```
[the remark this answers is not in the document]
TobiasKadelka: (Yeah sorry, this was changed in the later commit.)

TobiasKadelka:
Except the points on which I commented, I implemented changes for all of your feedback. Thanks a lot!

onyo/commands/edit.py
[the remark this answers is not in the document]
aqw: Ahh yes. I was not thinking about this in the context of `edit`. I agree that the "right" thing here is unknowable, and thus there should be no default.

onyo/lib/onyo.py
```diff
@@ -92,6 +92,94 @@ def clear_caches(self) -> None:
...
+            keys_str = ""
+            if keys:
+                keys_str = f" ({','.join(str(x.split('=')[0]) for x in sorted(keys))})"
+            if destination:
+                dest = Path(self.opdir, destination).relative_to(self.root)
+            if dest and str(dest.name) == ".anchor":
+                dest = dest.parent
+
+            # build the commit message subject
+            msg_dummy = f"{cmd} [{len(staged_changes)}]{keys_str}"
```
aqw, line 135: This variable name is not intuitive to me. Why is it a "dummy"?

I think comments here would really help to explain what's going on. It seems to always reference keys, even if we're not using `set`, etc.

onyo/lib/onyo.py
```diff
@@ -92,6 +92,94 @@ def clear_caches(self) -> None:
...
+                                         destination: Optional[Path],
+                                         MAX_LEN: int) -> str:
+        """
+        Generates "commit message subject" with the `msg_dummy` and the paths
+        from `staged_changes` and `destination`, and shortens the paths if the
+        message length exceeds the `MAX_LEN`.
+        """
+
+        # long message: full paths (relative to repo-root)
+        paths_str = ','.join(["'{}'".format(x) for x in staged_changes])
```
aqw, line 154: Nit: This might be a bit more readable (and consistent) if it uses f-strings.

onyo/lib/onyo.py
```diff
@@ -92,6 +92,94 @@ def clear_caches(self) -> None:
...
+
+        # short message: "type" of devices in summary (e.g.  "laptop (2)")
+        paths = [str(x.name).split('_')[0] for x in staged_changes]
+        paths_str = ','.join(sorted(["'{} ({})'".format(x, paths.count(x))
+                                     for x in set(paths)]))
+        msg = f"{msg_dummy}: {paths_str}"
+        if destination:
+            msg = f"{msg} -> '{destination.relative_to(destination.parent)}'"
+
+        # return the shortest possible version of the commit message as default
```
aqw, line 180: I think "default" is misleading here. It's rather a "fallback".

aqw (changes requested):
I only re-reviewed the `Repo.generate_commit_message()` commit (let me know if other stuff changed).

This iteration is much easier to read. It still is a bit complicated, but overall is approachable.

Some fixes are still required. I threw a few nits in there as well.

It's getting close!

onyo/lib/onyo.py
```diff
@@ -92,6 +92,94 @@ def clear_caches(self) -> None:
...
+    def generate_commit_message(self, message: list[str] = [], cmd: str = "",
+                                keys: list[str] = [], destination: str = "") -> str:
+        """
+        If `message` is given, the function uses the first element of it to
+        generate a message subject, and the following ones for the message body.
+
+        If no `message` is given, generates a default commit message based on
+        the other parameters given, and the `files staged` in the repository.
+        The paths for the message are shortened if needed, so that the resulting
+        message subject does not exceed `MAX_LEN`.
```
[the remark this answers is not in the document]
TobiasKadelka: Or via config file. I will add it as a parameter for now, but I think we should also at some point near-ish think about how to use the config file, since there is also the faux serial number length, and probably others.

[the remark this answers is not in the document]
aqw: I agree, and had that in mind as well. :-)

onyo/lib/onyo.py
```diff
@@ -92,6 +92,94 @@ def clear_caches(self) -> None:
...
+        Adds a list of `changed files` with there path relative to the root of
+        the repository to the body of all commit messages.
+        """
+        message_subject = ""
+        message_body = ""
+        message_appendix = ""
+        MAX_LEN = 80
+
+        # staged files and directories (without ".anchor") in alphabetical order
+        staged_changes = [Path(self.opdir, x).relative_to(self.root).parent
```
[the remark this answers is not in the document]
TobiasKadelka: It removes the ".anchor" part, if it is a `<path>/.anchor`. E.g. `root_to_onyo/shelf/.anchor` would be first relative to repo, and then `.parent` removes the `.anchor`, which leads to `shelf/`

onyo/lib/onyo.py
TobiasKadelka: Thanks, I will add comments to it.

Yes, I think always refering `keys` is the better way, because it saves the `if keys`, because we need to initialize keys anyways. I find it easier to read, and especially modify, the line this way, because otherwise we had just a bigger if/else tree with many similar lines, that all set keys.

onyo/lib/onyo.py
TobiasKadelka: Maybe, but I really don't know how to do that for the given list, to come to the result of them being all quoted and comma seperated for different lenghts :(

aqw: Ahh, I didn't mean the entire thing, rather:

```
paths_str = ','.join([f"'{x}'" for x in staged_changes])
```

onyo/lib/onyo.py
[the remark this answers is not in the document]
aqw: Ahhhhh. I get it now. Silly list comprehension makes the logic backwards. Ok. I see how this works.

I do have a nit here, but it really comes down to taste, so feel free to disagree.

IMO, the dominant case should come first (i.e. the non-anchor case). So something like:
```
        staged_changes = [x.parent
                          if not x.name == '.anchor'
                          else x
                          for x in sorted(self.files_staged)]
```

The downside is that this introduces another code smell: `if not`. Which I'm not really a fan of either.

So really, choose whatever you prefer, and I'm happy.

aqw (approved):
The list comprehension parts are much easier to read now.

LGTM. Thanks!

PR #302, opened 2023-03-07
TST: run tests on tmpfs

aqw:
This should speed up the tests significantly by removing pretty much all IO overhead.


aqw:
Aaaaand this had negligible impact (merely 16s improvement locally). ;-P

Perhaps we're not doing much IO and/or runner disks are backed by some RAM anyway.

PR #303, opened 2023-03-08
TST: run tests in random order

aqw:
To ensure we don't accidentally introduce inter-test dependencies.


TobiasKadelka:
I think this is a good idea :)

aqw:
I fixed the failing test.

The new plugin has already proved its value. :-D

PR #304, opened 2023-03-08
TST: add  `repo_contents` fixture to set values in assets; use fixture for `test_unset.py`

TobiasKadelka:
I added the `repo_contents` fixture, and to demonstrate the usage and functionality, I updated `test_unset.py` with it.

Also, I reduced some of the redundant/overlapping tests (i.e. the test directory names do not need multiple dirs with spaces), like I already did for many of the other tests.

### DIFF
--- a/tests/commands/test_unset.py
+++ b/tests/commands/test_unset.py
@@ -8,32 +8,31 @@ files = ['laptop_apple_macbookpro',
          'lap top_ap ple_mac book pro']
 
 directories = ['.',
-               'simple',
                's p a c e s',
-               's p a/c e s',
                'r/e/c/u/r/s/i/v/e',
                'overlap/one',
                'overlap/two',
-               'very/very/very/deep']
+               'very/very/very/deep',
+               ]
 
 assets = [f"{d}/{f}.{i}" for f in files for i, d in enumerate(directories)]
 
+content_dict = {"one_key": "one_value",
+                "two_key": "two_value",
+                "three_key": "three_value"}
 
-@pytest.mark.repo_files(*assets)
+content_str = "\n".join([f"{elem}: {content_dict.get(elem)}"
+                         for elem in content_dict]) + "\n"
+
+contents = [[x, content_str] for x in assets]
+
+@pytest.mark.repo_contents(*contents)
 @pytest.mark.parametrize('asset', assets)
 def test_unset(repo: Repo, asset: str) -> None:
     """
     Test that `onyo unset KEY <asset>` removes keys from of assets.
     """
-    set_values = "key=value"
-    key = "key"
-
-    # TODO: find out if there is a faster way than `onyo set` for writing
-    # without leaving an unclean git tree
-    ret = subprocess.run(['onyo', 'set', '--yes', '--keys', set_values, '--path', asset], capture_output=True, text=True)
-    assert ret.returncode == 0
-    assert "key: value" in Path.read_text(Path(asset))
-
+    key = list(content_dict.keys())[0]
     ret = subprocess.run(['onyo', 'unset', '--yes', '--keys', key, '--path', asset], capture_output=True, text=True)
 
     # verify output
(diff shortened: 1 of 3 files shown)


tests/commands/test_unset.py
```diff
@@ -8,32 +8,31 @@
...
+               ]
 
 assets = [f"{d}/{f}.{i}" for f in files for i, d in enumerate(directories)]
 
+content_dict = {"key": "value",
+                "second": "input",
+                "another": "field"}
 
-@pytest.mark.repo_files(*assets)
+content_str = "\n".join([f"{elem}: {content_dict.get(elem)}"
```
aqw, line 24: IMO, this is easier to read if it's not built programmatically, and instead as a literal string.

tests/commands/test_unset.py
```diff
@@ -8,32 +8,31 @@
...
                'r/e/c/u/r/s/i/v/e',
                'overlap/one',
                'overlap/two',
-               'very/very/very/deep']
+               'very/very/very/deep',
+               ]
 
 assets = [f"{d}/{f}.{i}" for f in files for i, d in enumerate(directories)]
 
+content_dict = {"key": "value",
```
aqw, line 20: As a bias: I would like if the keys and values were predictable and iterated (e.g. `one`, `two`, `three`, etc).

It will make using in tests and debugging more obvious.

tests/commands/test_unset.py
```diff
@@ -8,32 +8,31 @@
...
+contents = [[x, content_str] for x in assets]
+
+@pytest.mark.repo_contents(*contents)
 @pytest.mark.parametrize('asset', assets)
 def test_unset(repo: Repo, asset: str) -> None:
     """
     Test that `onyo unset KEY <asset>` removes keys from of assets.
     """
-    set_values = "key=value"
     key = "key"
```
aqw, line 35: IMO, this does not need to be a variable, and can be hard-coded in context below.

(This pattern is used elsewhere, and I won't comment on those)

aqw (commented):
LGTM. Mostly nits from my point of view.

tests/commands/test_unset.py
TobiasKadelka: For the people of the future, reading through all our PRs and trying to follow our thought process, I want to mention, that we decided against hardcoding strings and other variables after all, so as to enable us in the future to easier update our tests.

tests/commands/test_unset.py
TobiasKadelka: For the people of the future, reading through all our PRs and trying to follow our thought process, I want to mention, that we decided against hardcoding strings and other variables after all, so as to enable us in the future to easier update our tests.

tests/commands/test_unset.py
TobiasKadelka: This was done and will be pushed in a minute, thanks!

tests/commands/test_unset.py
```diff
@@ -8,32 +8,31 @@
...
-    set_values = "key=value"
-    key = "key"
-
-    # TODO: find out if there is a faster way than `onyo set` for writing
-    # without leaving an unclean git tree
-    ret = subprocess.run(['onyo', 'set', '--yes', '--keys', set_values, '--path', asset], capture_output=True, text=True)
-    assert ret.returncode == 0
-    assert "key: value" in Path.read_text(Path(asset))
-
+    key = list(content_dict.keys())[0]
```
aqw, line 35: As the key is still hard-coded later, I recommend the following as more readable and explicit:
```
key = content_dict.get('one')
```

tests/commands/test_unset.py
```diff
@@ -8,32 +8,31 @@
...
                'r/e/c/u/r/s/i/v/e',
                'overlap/one',
                'overlap/two',
-               'very/very/very/deep']
+               'very/very/very/deep',
+               ]
 
 assets = [f"{d}/{f}.{i}" for f in files for i, d in enumerate(directories)]
 
+content_dict = {"one": "value",
```
aqw, line 20: I thought I left a comment on this before, but perhaps it didn't save.

I would prefer that both the keys and values here are predictable. Currently the keys are, but the values aren't. There's currently no clear, predictable relation between them.

tests/commands/test_unset.py
```diff
@@ -44,23 +43,17 @@ def test_unset(repo: Repo, asset: str) -> None:
     assert ret.returncode == 0
 
     # verify changes, and the repository clean
-    assert "key: value" not in Path.read_text(Path(asset))
+    assert "one: value" not in Path.read_text(Path(asset))
```
aqw, line 46: `Path.read_text(Path(asset))` -> `Path(asset).read_text()`

aqw (commented):
A few nits still. Getting close. :-)

tests/commands/test_unset.py
TobiasKadelka: This would get the value.

tests/commands/test_unset.py
TobiasKadelka: You did, and yes, I did it just for the keys. It will be updated with the next push.

tests/commands/test_unset.py
TobiasKadelka: Nice, I'll change it at all places, thanks.

aqw (approved):
LGTM!

PR #307, opened 2023-03-13
TST: Modernize/streamline the tests for commands

TobiasKadelka:
This PR contains a lot of small modernizations, that basically try to make all tests look/work similar, which mainly add or effect Pyre, doc-strings, parameterization, usage of the new fixture, and such things, that I found while looking through the tests for the commands. These do not add any functionality as such, but rather are meant to make the tests easier to update in the future. Also, I used valid asset names in tests (except for cases that do not test the behavior for invalid names, obviously).

Also, in some cases I noticed some of our more classic test cases missing, e.g. deletion of directories with `onyo rm`, or having at least one or two test cases for each flag of the commands. This is not meant to cover all scenarios in this PR, but I think it will help in case we want to split the code base in the future into independent packages.

When tests contained calls to subcommands, which are not tested, these were replaced by API calls, like already done for the tests of `onyo unset`. I updated the order of tests for `onyo rm`, so that now all our tests on assets start on single files, then a list of files.

### DIFF
--- a/tests/commands/test_history.py
+++ b/tests/commands/test_history.py
@@ -1,158 +1,192 @@
 import subprocess
 from pathlib import Path
+
+from onyo.lib import Repo
 import pytest
 
 
 # NOTE: the output of `onyo history` is not tested for formatting or content, as
 #       the commands called by `onyo history` are user-configurable. Instead, it
 #       is tested for whether the output of the underlying command is unaltered.
-variants = [
-    'file',
-    'dir',
-    'dir/subfile-1',
-    's p a c e s/subfile-2',
-    's p a/c e s/sub file 3',
-]
-@pytest.mark.repo_files('file', 'dir/subfile-1', 's p a c e s/subfile-2', 's p a/c e s/sub file 3')
-@pytest.mark.parametrize('variant', variants)
-def test_history_noninteractive(repo, variant):
-    ret = subprocess.run(['onyo', 'history', '-I', variant],
+
+
+files = ['laptop_apple_macbookpro',
+         'lap top_ap ple_mac book pro',
+         'spe\"c_ialch_ar\'ac.teஞrs']
+
+directories = ['.',
+               's p a c e s',
+               'very/very/very/deep',
+               ]
+
+assets = [f"{d}/{f}.{i}" for f in files for i, d in enumerate(directories)]
+
+
+@pytest.mark.repo_files(*assets)
+@pytest.mark.parametrize('asset', assets)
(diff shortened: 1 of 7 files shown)


tests/commands/test_onyo.py
```diff
@@ -10,27 +11,29 @@
 ]
 @pytest.mark.repo_dirs('just-a-dir')
 @pytest.mark.parametrize('variant', variants)
-def test_onyo_debug(repo, variant):
+def test_onyo_debug(repo: Repo, variant: str) -> None:
     ret = subprocess.run(['onyo', variant, 'mkdir', '--yes', f'flag{variant}'],
                          capture_output=True, text=True)
     assert ret.returncode == 0
     assert 'DEBUG:onyo' in ret.stderr
+    repo.fsck()
```
nhjjr, line 19: Here I would suggest using:

```
from contextlib import nullcontext as does_not_raise

with does_not_raise():
    repo.fsck()
```

You're expecting this to pass without raising any exceptions, Since the function returns `None` and only raises an exception if there are problems, we can be specific about this in the tests.

You can read more about this [in the pytest docs](https://docs.pytest.org/en/6.2.x/example/parametrize.html#parametrizing-conditional-raising) on parametrizing conditional raising.

tests/commands/test_history.py
```diff
@@ -1,137 +1,175 @@
...
-    'subdir/does-not-exist',
+no_assets = [
+    'does_not_exist.test',
+    'subdir/does_not_exist.test',
 ]
-@pytest.mark.repo_dirs('subdir')
-@pytest.mark.parametrize('variant', variants)
-def test_history_noninteractive_not_exist(repo, variant):
-    ret = subprocess.run(['onyo', 'history', '-I', variant],
+@pytest.mark.repo_dirs(*assets)
```
nhjjr, line 43: flake8 does not appear to address: `PEP 8: E302 expected 2 blank lines, found 0` on the function definition, because the closing bracket of the variable assignment is on the line before it. Furthermore, while your approach works with the way pytest evaluates a test file and executes the tests, it can easily break if the test order is randomized, and the variable is reassigned later in the file.

Rather than creating a variable and passing it to `parametrize`, I suggest writing the contents of the variable in there, e.g.,

```
@pytest.mark.parametrize('variant', ['does_not_exist.test', 'subdir/does_not_exist.test'])
```

TobiasKadelka: You are right, the definitions of variants should happen there, it just works for the static ones that are the same for all situations anyways, without being ever modified. Thanks! :)

Issue #306, opened 2023-03-13
More complete `.onyo/` support


TobiasKadelka:
The `.onyo` directory is quite important for a bunch of functionalities, and yet it feels incomplete.
This does contain a bunch of things I am repeatedly thinking about might be helpful (and which are currently uncomfortable to use).

- add `fsck` for fully populated `.onyo/`
  - replace `fully_populated_dot_onyo()` from `tests/commands/test_init.py` with a normal `repo.fsck()`
- add something like `onyo repair` command:
  - adds everything missing to `.onyo/`
  - adds `.anchor` to all directories
  - commits unsaved/uncommited changes
  - obviously all with "diff" to verify that nothing is done that should not be happening
- add tests to verify a damaged `.onyo/` does not mess up everything:
  - e.g. that `.gitconfig` is used when `.onyo/config` is missing
  - missing template folder does not crash, but instead mentions that the dir/template is missing
- enable easier template editing (remove some default fscks for `.onyo/`?)
  - `onyo new --path .onyo/templates/new_template` errors because `new_template` is no asset name
  - using an editor to create a new template forces user to `git add` and `git commit` manually
- https://github.com/psyinfra/onyo/issues/101

TobiasKadelka:
In #308, when clearing caches, I often clear `Repo._templates` and bravely declare these might be changed by the commands. This is on ocasions a lie, but it does reflect the fact that, like described in this issue above, the missing abilities of onyo to change templates is in my opinion somewhere between missing feature and bug.

TobiasKadelka:
I added the flexible name scheme issue regarding changable pseudo keys to the initial description of the issue.

aqw:
I am closing this as this issue is largely covered by #599 and then the template topic is covered in #552.

PR #308, opened 2023-03-14
NF/BF: clear caches after modifying repository

### LINKED ISSUE #248: BUG: `Repo` caches are not cleared on repo change
The following caches are used by the `Repo` class:
```
self._assets = None
self._dirs = None
self._files = None
self._gitfiles = None
 ```

Methods that modify the state of the repo (`mkdir`, `mv`, `rm`, etc) should clear those caches.

### PR DESCRIPTION
**Pull request by TobiasKadelka**
The function to re-set the caches to `None` do exist since the implementation of `onyo set`, yet they were not used in most functions, and neither in a consistent way.

This commit adds a `clear_caches()` call to the end of all functions in the API that modify the caches, and to the command `onyo new` in the function modifying the state of the repository, e.g. adding new assets.

While doing these changes, I noticed how much we should have something like API-workflow tests, or like the demo, since it is quite hard to verify that not some other functions (e.g. `restore()`, `add()` and `commit()` as git functions) also would need a clearing of caches. I think these should not require this currently, because they get mainly called by the command-functions, or in the command-files, but we should add these tests soon anyways to be sure.

Close #248

### DIFF
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -84,11 +84,31 @@ class Repo:
 
         return self._templates
 
-    def clear_caches(self) -> None:
-        self._assets = None
-        self._dirs = None
-        self._files = None
-        self._templates = None
+    def clear_caches(self, assets: bool = True, dirs: bool = True,
+                     files: bool = True, templates: bool = True) -> None:
+        """
+        Clear caches of the instance of the repository object.
+
+        Paths such as files, assets, and directories are cached, and can become
+        stale when the repository contents are modified. This function clears
+        the caches of the properties.
+
+        By default all caches are cleared, and arguments make it possible to
+        specify which caches which should remain set.
+
+        If the repository is exclusively modified via public API functions, the
+        caches of the `Repo` object are consistent. If the repository is
+        modified otherwise, this function clears the caches to ensure that the
+        caches do not contain stale information.
+        """
+        if assets:
+            self._assets = None
+        if dirs:
+            self._dirs = None
+        if files:
+            self._files = None
+        if templates:
+            self._templates = None
 
     def generate_commit_message(self, message: list[str] = [], cmd: str = "",
                                 keys: list[str] = [], destination: str = "",
@@ -187,6 +207,16 @@ class Repo:
         """
         self._git(['restore', '--source=HEAD', '--staged', '--worktree'] +
                   [str(file) for file in self.files_staged])
+        # `Repo.restore()` gets used by all most commands that might change the
+        # repository to revert changes, especially when users answer "no" to
+        # user dialogs. It might also be used by the API to reset the repository
+        # variable after doing some manual changes on files (e.g. with
+        # subprocess).
+        self.clear_caches(assets=True,  # revert e.g. `onyo set`
+                          dirs=True,  # revert e.g. `onyo mkdir`
+                          files=True,  # revert anchors of `onyo mkdir`
+                          templates=True  # revert `onyo edit` in `.onyo`
+                          )
 
     def _get_assets(self) -> set[Path]:
         """
@@ -326,6 +356,15 @@ class Repo:
                 raise FileNotFoundError(f"'{t}' does not exist.")
 
         self._git(['add'] + tgts)
+        # `Repo.add()` is used by most repo-changing commands, and it might be
+        # used often by the API to change the repository, before manually
+        # calling clear_cache() or after a file was changed with a subprocess
+        # call. To always secure the integrity of the caches, we reset them all.
+        self.clear_caches(assets=True,  # e.g. `onyo set` changes asset names
+                          dirs=True,  # e.g. `onyo mkdir` adds new dirs
+                          files=True,  # e.g. `onyo mkdir` adds new ".anchor"
+                          templates=True  # e.g. `onyo edit` in `.onyo/templates`
+                          )
 
     #
     # COMMIT
@@ -755,6 +794,11 @@ class Repo:
         log.debug('The following will be moved:\n' +
                   '\n'.join([str(x.relative_to(self.opdir)) for x in src_paths]))
 
+        self.clear_caches(assets=True,  # `onyo mv` can change the dir of assets
+                          dirs=True,  # might move directories
+                          files=True,  # might move anchors
+                          templates=True  # might move or rename templates
+                          )
         # return a list of mv-ed assets
         # TODO: is this relative to opdir or root? (should be opdir)
         return [r for r in re.findall('Renaming (.*) to (.*)', ret)]
@@ -1064,9 +1108,7 @@ class Repo:
         if name_values:
             try:
                 self._update_names(assets_to_set, name_values)
-                self.clear_caches()
             except ValueError as e:
-                self.clear_caches()
                 if self.files_staged:
                     self.restore()
                 # reset renaming needs double-restoring
@@ -1079,6 +1121,11 @@ class Repo:
         if diff and dryrun:
             self.restore()
 
+        self.clear_caches(assets=True,  # `onyo set` can rename assets
+                          dirs=False,  # `set` cannot create, move, or remove directories
+                          files=True,  # might rename asset files
+                          templates=True  # might modify templates
+                          )
         return diff
 
     def _select_assets_from_directory(self, directory: Path,
@@ -1274,6 +1321,11 @@ class Repo:
         log.debug('The following will be deleted:\n' +
                   '\n'.join([str(x.relative_to(self.opdir)) for x in paths_to_rm]))
 
+        self.clear_caches(assets=True,  # `onyo rm` can delete assets
+                          dirs=True,  # can delete directories
+                          files=True,  # if used on dir, deletes also `.anchor`
+                          templates=True  # can delete templates
+                          )
         # return a list of rm-ed assets
         # TODO: should this also list the dirs?
         # TODO: is this relative to opdir or root? (should be opdir)
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -38,6 +38,15 @@ class Repo:
 
     @property
     def assets(self) -> set[Path]:  # pyre-ignore[11]
+        """
+        A `set` containing the `Path`s relative to `Repo.root` of all assets of
+        an Onyo repository.
+
+        This property is cached, and the cache is consistent with the state of
+        the repository when only `Repo`s public functions are used. Use of
+        private functions might require a manual reset of the caches, see
+        `Repo.clear_caches()`.
+        """
         if not self._assets:
             self._assets = self._get_assets()
 
@@ -45,6 +54,15 @@ class Repo:
 
     @property
     def dirs(self) -> set[Path]:
+        """
+        A `set` containing the `Path`s relative to `Repo.root` of all directories
+        of an Onyo repository.
+
+        This property is cached, and the cache is consistent with the state of
+        the repository when only `Repo`s public functions are used. Use of
+        private functions might require a manual reset of the caches, see
+        `Repo.clear_caches()`.
+        """
         if not self._dirs:
             self._dirs = self._get_dirs()
 
@@ -52,6 +70,15 @@ class Repo:
 
     @property
     def files(self) -> set[Path]:
+        """
+        A `set` containing the `Paths` relative to `Repo.root` of all files of an
+        Onyo repository.
+
+        This property is cached, and the cache is consistent with the state of
+        the repository when only `Repo`s public functions are used. Use of
+        private functions might require a manual reset of the caches, see
+        `Repo.clear_caches()`.
+        """
         if not self._files:
             self._files = self._get_files()
 
@@ -59,26 +86,55 @@ class Repo:
 
     @property
     def files_changed(self) -> set[Path]:
+        """
+        Returns a `set` containing the `Path`s relative to `Repo.root` of all
+        changed files (according to git) of an Onyo repository.
+        """
         return self._get_files_changed()
 
     @property
     def files_staged(self) -> set[Path]:
+        """
+        Returns a `set` containing the `Path`s relative to `Repo.root` of all
+        staged files (according to git) of an Onyo repository.
+        """
         return self._get_files_staged()
 
     @property
     def files_untracked(self) -> set[Path]:
+        """
+        Returns a `set` containing the `Path`s relative to `Repo.root` of all
+        untracked files (according to git) of an Onyo repository
+        """
         return self._get_files_untracked()
 
     @property
     def opdir(self) -> Path:
+        """
+        Returns the absolute `Path` to the working directory where the
+        `Repo` object was instantiated.
+        """
         return self._opdir
 
     @property
     def root(self) -> Path:
+        """
+        Returns the absolute `Path` to the root of the `Repo` where the
+        directories `.git/` and `.onyo/` are located.
+        """
         return self._root
 
     @property
     def templates(self) -> set[Path]:
+        """
+        A `set` containing the `Path`s relative to `Repo.root` of all
+        template files of an Onyo repository.
+
+        This property is cached, and the cache is consistent with the state of
+        the repository when only `Repo`s public functions are used. Use of
+        private functions might require a manual reset of the caches, see
+        `Repo.clear_caches()`.
+        """
         if not self._templates:
             self._templates = self._get_templates()
 

### CODE REVIEW

File: onyo/commands/new.py [hunk 1]
```diff
@@ -22,6 +22,7 @@ def create_assets_in_destination(assets: dict, repo: Repo) -> None:
             asset.touch()
         repo._write_asset(asset, assets[asset])
     repo.add(list(assets.keys()))
+    repo.clear_caches()
```
aqw, line 25: IMO, this should be made part of `repo.add()`, so that it always does the right thing internally, rather than requiring consumers to maintain the state of non-public structures internal to `repo`.

Also, I think it would be better if it was more granular which caches are cleared, rather than always all.

aqw (changes requested):

onyo/commands/new.py
TobiasKadelka: I added it to `Repo.add()`, though did not modify the cache-clearing behavior for more granularity, since I think it is not clear to me, that we reliably know, if assets, directories, templates or else were changed.
This will be even more true, as soon as we add functionality to easier modify e.g. templates.

onyo/lib/onyo.py
```diff
@@ -85,12 +85,19 @@ def templates(self) -> set[Path]:
 
         return self._templates
 
-    def clear_caches(self) -> None:
-        self._assets = None
-        self._dirs = None
-        self._files = None
-        self._gitfiles = None
-        self._templates = None
+    def clear_caches(self, assets: bool = True, dirs: bool = True,
```
aqw, line 88: Just so I understand the intended use pattern: calling this function will either clear everything, or an explicit list of what /not/ to clear needs to be specified.

The listing of negatives feels backwards to me. Perhaps a list of caches, that is fully populated by default.

onyo/lib/onyo.py
```diff
@@ -85,12 +85,19 @@ def templates(self) -> set[Path]:
...
+                     files: bool = True, gitfiles: bool = True,
+                     templates: bool = True) -> None:
+        if assets and self._assets is not None:
```
aqw, line 91: IMO, there's no need to check if `is not None`. It doesn't protect anything nor saves computational effort.

onyo/lib/onyo.py
```diff
@@ -189,6 +196,7 @@ def restore(self) -> None:
         """
         self._git(['restore', '--source=HEAD', '--staged', '--worktree'] +
                   [str(file) for file in self.files_staged])
+        self.clear_caches()
```
aqw, line 199: Why does this method clear all of the caches?

TobiasKadelka: I think it gets used when the user response is "no" to reset directories of mkdir, it is used by set to restore content and file name changes, and when edit/add will be allowed to edit or add templates (I propose this in #306) it might want to be used for these, too.

onyo/lib/onyo.py
aqw: > Writting None to a variable that is already None costs runtime, I think.

I would be surprised if that is the case with `None`. I don't intend to bike-shed this, but IMO, the current code is more verbose, and not-protecting against a significant runtime cost.

If you profile that chunk of code, I would be interested in seeing the results.

onyo/lib/onyo.py
aqw: Perhaps the wrong place to have asked, but in general, this question should be applied to all uses of `clear_caches()`.

It's not clear to me why invoking `add` should invalidate the cache of `templates`.

TobiasKadelka: The onyo commands we use do just change once the state of the repository, and commit, for these the `clear_cache()` is as far as I know not immediately relevant.

I was thinking rather about somebody writing there own python script and using `restore()` and `add()`.
e.g. something like this might happen:

```
repo = Repo(".")
# here doing anything which leads to loading the template cache
template = Path(".onyo/templates/new_device")
template.write("RAM: 16GB")
repo.add(template)
repo.commit("added new template")
# here doing something which would need the new template being in the cache
```

alternatively, somebody might want to copy our existing code (especially  `edit` or `new`), and just add something to the end of it:
```
    # commit changes
    staged = sorted(repo.files_staged)
    if staged:
        if not args.quiet:
            print(repo._diff_changes())
        if args.yes or request_user_response("Save changes? No discards all changes. (y/n) "):
            repo.commit(repo.generate_commit_message(message=args.message,
                                                     cmd="edit"))
        else:
            repo.restore()
            if not args.quiet:
                print('No assets updated.')

# here some code that needs the cache, after the user responded with "no"
```

I know these cases are currently not super relevant, because we do not have any commands doing something with the caches after commiting/restoring, and we have very weak template support (I mentioned this in #306 and still think this should be worked on).
But it feels like not updating the caches of the repository, after the state of the repository does change leads the `repo` var in an inaccurate state, after the user explicitely stated that something should be added or the whole thing should be restored.

onyo/lib/onyo.py
TobiasKadelka: This is probably the main thing I was thinking about quite a lot since yesterday, and I can't say that you are wrong.

I personally feel like having booleans and being forced to explicetely state deactivation is from API usage perspective the most comfortable way to use it. I don't want to check everytime all caches the Repo class offers, just to build a list of the 4-5 of the six I want to activate. Either I want them all to be cleared, because something is changed, or I should be sure that this specific cache is not needed.

We can discuss this more deeply in a meeting or in here if you want, though my current thoughts all go back to "this is very different from the fsck functionality" and the usage like this seems most comfortable and intuitive to me.

onyo/lib/onyo.py
TobiasKadelka: After profiling, I give up the argument, since it does not change run-time. I will remove the unneeded check for `None` like proposed by you 👍

aqw: Thanks for checking! :-D 

It would have been cool to see the results. In the future, please share!

onyo/lib/onyo.py
```diff
@@ -1066,9 +1100,13 @@ def set(self, paths: Iterable[Union[Path, str]], values: dict, dryrun: bool,
         if name_values:
             try:
                 self._update_names(assets_to_set, name_values)
-                self.clear_caches()
             except ValueError as e:
-                self.clear_caches()
+                self.clear_caches(assets=True,  # `onyo set` can rename assets
+                                  dirs=False,  # no szenario with `set`
```
aqw, line 1105: "szenario" -> "scenario"

But I prefer a more explicit statement: "`set` cannot create, move, or remove directories".

onyo/lib/onyo.py
```diff
@@ -1066,9 +1100,13 @@ def set(self, paths: Iterable[Union[Path, str]], values: dict, dryrun: bool,
...
+                                  files=True,  # in theory might change other YAML
+                                  gitfiles=False,  # no szenario with `set`
```
aqw, line 1107: "szenario" -> "scenario"

onyo/lib/onyo.py
```diff
@@ -1066,9 +1100,13 @@ def set(self, paths: Iterable[Union[Path, str]], values: dict, dryrun: bool,
         if name_values:
             try:
                 self._update_names(assets_to_set, name_values)
-                self.clear_caches()
             except ValueError as e:
-                self.clear_caches()
+                self.clear_caches(assets=True,  # `onyo set` can rename assets
+                                  dirs=False,  # no szenario with `set`
+                                  files=True,  # in theory might change other YAML
```
aqw, line 1106: This is insufficiently specific. Vague "in theory" statements are not helpful.

This cache should be cleared because assets can be renamed, and assets are files.

onyo/lib/onyo.py
```diff
@@ -1081,6 +1119,12 @@ def set(self, paths: Iterable[Union[Path, str]], values: dict, dryrun: bool,
         if diff and dryrun:
             self.restore()
 
+        self.clear_caches(assets=True,  # `onyo set` can rename assets
```
aqw, line 1122: Same comments as above.

Not a fan of the duplication of this; it will require that they remain in sync. But it is unclear to me why the caches are cleared explicitly above.

onyo/lib/onyo.py
```diff
@@ -1066,9 +1100,13 @@ def set(self, paths: Iterable[Union[Path, str]], values: dict, dryrun: bool,
         if name_values:
             try:
                 self._update_names(assets_to_set, name_values)
-                self.clear_caches()
             except ValueError as e:
-                self.clear_caches()
+                self.clear_caches(assets=True,  # `onyo set` can rename assets
```
aqw, line 1104: Why are the caches cleared before doing the `restore()` (which clears the caches)?

onyo/lib/onyo.py
```diff
@@ -1066,9 +1100,13 @@ def set(self, paths: Iterable[Union[Path, str]], values: dict, dryrun: bool,
...
+                                  dirs=False,  # no szenario with `set`
+                                  files=True,  # in theory might change other YAML
+                                  gitfiles=False,  # no szenario with `set`
+                                  templates=True  # might modify templates
+                                  )
                 if self.files_staged:
                     self.restore()
                 # reset renaming needs double-restoring
```
aqw, line 1112: Why?!?

aqw (changes requested):

onyo/lib/onyo.py
```diff
@@ -326,6 +351,15 @@ def add(self, targets: Union[Iterable[Union[Path, str]], Path, str]) -> None:
                 raise FileNotFoundError(f"'{t}' does not exist.")
 
         self._git(['add'] + tgts)
+        # `Repo.add()` is used by most repo-changing commands, and it might be
+        # used often by the API to change the repository, before manually
+        # calling clear_cache() or after a file was changed with a subprocess
+        # call. To always secure the integrity of the caches, we reset them all.
+        self.clear_caches(assets=True,  # revert e.g. `onyo set`
```
aqw, line 358: Copy paste error; the comments all reference "revert"

onyo/lib/onyo.py
```diff
@@ -326,6 +351,15 @@ def add(self, targets: Union[Iterable[Union[Path, str]], Path, str]) -> None:
                 raise FileNotFoundError(f"'{t}' does not exist.")
 
         self._git(['add'] + tgts)
+        # `Repo.add()` is used by most repo-changing commands, and it might be
```
aqw, line 354: Reflecting on this further... I wonder if the cache should not be cleared with `add`. These changes have only been staged, and are not in-effect yet. `commit`, however, that indeed is a change to the onyo repo...

I think what is really exposed here is that there are no docstrings for the properties `files`, `assets`, etc. And thus it's not defined /what/ those properties are guaranteeing. And thus it is not entirely clear when those caches should be cleared.

onyo/lib/onyo.py
aqw: Ping

aqw (changes requested):
I'm not trying to blow up this PR, but I think what is really exposed by this PR is a missing dependency: There are no docstrings for the properties files, assets, etc. And thus it's not defined /what/ those properties are guaranteeing. And thus it is not entirely clear when these caches should be cleared.

We should determine if the caches are for what's on disk (which is impossible) or if it's for what's in-git (which is generally what onyo seems to be doing). But then there's staged vs committed. And that should be clarified.

onyo/lib/onyo.py
TobiasKadelka: Because, when using the command we use to restore (`git restore --source=HEAD --staged --worktree`) it does not clean up the tree completely, it just reverses the renaming, and then it does need another `git restore` to get the (through the moving) deleted file back.
```
❱ git status
On branch master
nothing to commit, working tree clean
❱ git mv repair/laptop_apple_macbookpro.dd082o recycling/new_name_same.device
❱ git status
On branch master
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	renamed:    repair/laptop_apple_macbookpro.dd082o -> recycling/new_name_same.device

tkadelka@bnbmac73 in ~/onyo on git:master+
❱ git restore --source=HEAD --staged --worktree recycling/new_name_same.device
tkadelka@bnbmac73 in ~/onyo on git:master+
❱ git status
On branch master
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	deleted:    repair/laptop_apple_macbookpro.dd082o
```

I did back in the days when working on this not find any other solution to reverse it completely with one call that also works for all other commands. If you know or come along an alternative please let me know!

onyo/commands/new.py
TobiasKadelka: I want to ping you on this one, @aqw because of what you wrote in the downstairs review about this being the most controversial part😇

I thought about it, too, and think from an API perspective it makes most sense to do this in the repo class, since "add" almost literally is saying to the repo "something changed".

onyo/lib/onyo.py
TobiasKadelka: It seems to be a vocabulary error, I changed them all from reverse. I will look up the different words and check again for which situations they are appropriate 🙈

onyo/lib/onyo.py
TobiasKadelka: I would say `commit` is a change to a `git` repository, but when you `Repo.add()` something, it is already a change to the onyo repo. I don't want to diverge too much between the too, but I shared my opinion for discussion in the chat to discuss this.

onyo/lib/onyo.py
```diff
@@ -38,47 +38,104 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
 
     @property
     def assets(self) -> set[Path]:  # pyre-ignore[11]
+        """
+        A `set` containing the `Path`s to all assets of an Onyo repository.
```
aqw, line 42: One bit of missing information: are these paths:
- absolute
- relative to `Repo.root`
- relative to `Repo.opdir`

onyo/lib/onyo.py
```diff
@@ -38,47 +38,104 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+
+        If the property is not set, load the assets of the repository and save
```
aqw, line 44: IMO, this should be removed. It's purely a private concern, and a consumer does not care.

onyo/lib/onyo.py
```diff
@@ -38,47 +38,104 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        them in `self._assets`.
+
+        When a function of `Repo` changes assets, the value will be reset to
```
aqw, line 47: This misses the point IMO.

First: "the value will be reset to `None`" is incorrect. From the user's PoV, it is /never/ None.

The /real/ point here is that the cache is guaranteed to remain consistent when `Repo`'s public APIs are used (`set()`, `add()`, etc). Use of private APIs (such as `_git()`) do not guarantee cache correctness. See also: `Repo.clear_caches()`.

onyo/lib/onyo.py
```diff
@@ -38,47 +38,104 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        """
         if not self._files:
             self._files = self._get_files()
 
         return self._files
 
     @property
     def files_changed(self) -> set[Path]:
+        """
+        Returns a `set` containing the `Path`s to all changed files (in git) of
```
aqw, line 90: nit: "in git" -> "according to git"

onyo/lib/onyo.py
```diff
@@ -38,47 +38,104 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        """
+        Returns a `set` containing the `Path`s to all untracked files (in git)
+        of an Onyo repository
+        """
         return self._get_files_untracked()
 
     @property
     def opdir(self) -> Path:
+        """
+        Returns the `Path` to the location where the `Repo` object was
```
aqw, line 114: "location" might be better as "working directory".

onyo/lib/onyo.py
```diff
@@ -38,47 +38,104 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        instantiated.
+        """
         return self._opdir
 
     @property
     def root(self) -> Path:
+        """
+        Returns the root `Path` of the `Repo` where the folders `.git/` and
```
aqw, line 122: folders -> directories

onyo/lib/onyo.py
```diff
@@ -95,6 +152,11 @@ def clear_caches(self, assets: bool = True, dirs: bool = True,
 
         By default all caches are cleared, and flags make it possible to specify
         which caches which should remain set.
+
+        If the repository is only modified via the API functions, the caches of
```
aqw, line 156: API functions -> public API functions

onyo/lib/onyo.py
```diff
@@ -95,6 +152,11 @@ def clear_caches(self, assets: bool = True, dirs: bool = True,
...
+        the `Repo` object are kept consistent. If an external process modifies
```
aqw, line 157: nit: remove external. Internally, it's /still/ the responsibility to correctly manage the caches.

onyo/lib/onyo.py
```diff
@@ -95,6 +152,11 @@ def clear_caches(self, assets: bool = True, dirs: bool = True,
...
+        the repository the caches must be cleared manually to restore the
+        integrity of the caches.
```
aqw, line 159: integrity -> sanity ?

Not sure I like that, but integrity feels wrong.

aqw (changes requested):

onyo/lib/onyo.py
TobiasKadelka: I think I will choose instead "consistency". "sanity" sounds missleading, it means kind of nothing, because who knows whats sane and what not? :P 
A lot of invalidation would be sane to expect in certain situations, I would argue.

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
 
     @property
     def assets(self) -> set[Path]:  # pyre-ignore[11]
+        """
+        A `set` containing the `Path`s relative to repo.root to all assets of an
```
aqw, line 42: Nit: I would write this a bit differently. Also use backticks for `Repo.root`, and capitalize the `R` in `Repo.root`

> A `set` containing all asset `Path`s relative to `Repo.root`.

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        Onyo repository.
+
+        The cache will be consistent with the state of the repository as long as
```
aqw, line 45: Nit:

> The cache will be consistent with the state of the repository as long as only the `Repo`s public functions are used.
to

> This property is cached, and the cache is consistent with the state of the repository when `Repo`'s public functions are used.

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        only the `Repo`s public functions are used. Use of private functions
```
aqw, line 46: "with `Repo.clear_caches()`" -> "See  `Repo.clear_caches()`."

I like referring them to the other docstring, because it will explain in more detail.

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        """
+        Returns the absolute `Path` to the root of the `Repo` where the
+        directories `.git/` and `.onyo/` are located.
+        """
         return self._root
 
     @property
     def templates(self) -> set[Path]:
+        """
+        A `set` containing the `Path`s relative to repo.root to all
```
aqw, line 127: "to all" -> "of all"

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
 
-    def clear_caches(self) -> None:
-        self._assets = None
-        self._dirs = None
-        self._files = None
-        self._templates = None
+    def clear_caches(self, assets: bool = True, dirs: bool = True,
+                     files: bool = True, templates: bool = True) -> None:
+        """
+        Clear all specified caches of the repository class.
```
aqw, line 142: "all specified" feels redundant and/or misleading here.

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+
+        The class saves paths, which can get invalidated when files or
```
aqw, line 144: invalidated -> stale

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        directories are modified. To force onyo to reload the paths, this
```
aqw, line 145: This doesn't force a reload of the properties; this clears the cache. The properties will only be reloaded on-demand.

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        function clears the caches.
+
+        By default all caches are cleared, and flags make it possible to specify
```
aqw, line 148: flags -> arguments

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        which caches which should remain set.
+
+        If the repository is only modified via the public API functions, the
+        caches of the `Repo` object are kept consistent. If the repository is
```
aqw, line 152: Remove "kept"

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        Clear all specified caches of the repository class.
+
+        The class saves paths, which can get invalidated when files or
+        directories are modified. To force onyo to reload the paths, this
+        function clears the caches.
+
+        By default all caches are cleared, and flags make it possible to specify
+        which caches which should remain set.
+
+        If the repository is only modified via the public API functions, the
```
aqw, line 151: > If the repository is only modified via the public API functions

to

> If the repository is modified via public API functions

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        caches of the `Repo` object are kept consistent. If the repository is
+        modified otherwise, the caches must be cleared manually to ensure the
```
aqw, line 153: Is "manually" using this function, or through some other way?

onyo/lib/onyo.py
```diff
@@ -38,57 +38,129 @@ def __init__(self, path: Union[Path, str], init: bool = False) -> None:
...
+        validity of the cache states.
```
aqw, line 154: Either remove this, or reword "to ensure that the cache does not contain stale information."

onyo/lib/onyo.py
aqw: We discussed this over a beer last night. I have come to the conclusion that using `Repo`'s public APIs will guarantee cache correctness. So it's about use rather than what's on disk or committed.

aqw (changes requested):

onyo/lib/onyo.py
TobiasKadelka: I think then it would not be true anymore. If you use the Repo "via public API functions", the sentence does not exclude the usage of other functions. And they must in my opinion by specifically excluded, because not all public API functions do reset the cache, so the usage of public API functions does not keep them consistent, except when *only* using them.

aqw: Yes. I agree. :-)

aqw (approved):
LGTM!

Sorry for the previous confusion. I did not see that the `clear_caches()` method was fixed in a different commit.

This reads very well and clearly. +1

I personally would like to see the `clear_caches()` docstring squashed to the previous commit. But I understand that's a pain and so is not a blocker here.

+1 from me.

TobiasKadelka:
Thanks!

Actually, the lines of the doc-strings for properties were as separate as possible from the changes in the other function, so I did a quick `git reset --soft HEAD~1` and fixxed that. Now it should all be were it belongs :)

I think as soon as the tests are finished successfully, this might then be merged :)

aqw:
Great. Thanks! :-)

PR #310, opened 2023-03-14
NF: add fsck check for pseudo-key names in the asset file

### LINKED ISSUE #283: add `fsck` check for pseudo-key names in the asset file
Pseudo Keys (`type`, `make`, `model`, `serial`) are reserved key-names and while they technically /can/ exist in the asset file, they are redundant and will be ignored by `get` and others.

There should be an `fsck` check to ensure that no asset file contains keys with those names.

TobiasKadelka:
Adds the check to `Repo.fsck()`, and adds tests.

Close #283

### DIFF
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -435,6 +435,7 @@ class Repo:
         - "asset-yaml": loads each assets and checks if it's valid YAML
         - "asset-validity": loads each asset and validates the contents against
           the validation rulesets defined in ``.onyo/validation/``.
+        - "pseudo-keys": verifies that assets do not contain pseudo-key names
         """
         all_tests = {
             "clean-tree": self._fsck_clean_tree,
@@ -442,6 +443,7 @@ class Repo:
             "asset-unique": self._fsck_unique_assets,
             "asset-yaml": self._fsck_yaml,
             "asset-validity": self._fsck_validation,
+            "pseudo-keys": self._fsck_pseudo_keys,
         }
         if tests:
             # only known tests are accepted
@@ -505,6 +507,28 @@ class Repo:
                           self._n_join(untracked))
 
             log.error('Please commit all changes or add untracked files to .gitignore')
+
+            return False
+
+        return True
+
+    def _fsck_pseudo_keys(self) -> bool:
+        """
+        Check that no asset contains any pseudo-key names.
+        """
+        assets_failed = {}
+        pseudo_keys = ["type", "make", "model", "serial"]
+
+        for asset in self.assets:
+            violation_list = [x for x in pseudo_keys if x in self._read_asset(asset)]
+            if violation_list:
+                assets_failed[asset] = violation_list
+
+        if assets_failed:
+            log.error(f"Pseudo keys ({', '.join(pseudo_keys)}) are reserved for asset names, and are not allowed in asset files. The following assets contain pseudo keys:\n" +
+                      self._n_join([f'{asset}: ' +
+                                    ', '.join([k for k in assets_failed[asset]])
+                                    for asset in assets_failed]))
 
             return False
 
--- a/tests/lib/test_Repo_fsck.py
+++ b/tests/lib/test_Repo_fsck.py
@@ -2,8 +2,7 @@ import logging
 from pathlib import Path
 
 import pytest
-from onyo import OnyoInvalidRepoError
-
+from onyo.lib import Repo, OnyoInvalidRepoError
 
 #
 # Generic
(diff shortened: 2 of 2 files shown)


onyo/lib/onyo.py
```diff
@@ -435,13 +435,15 @@ def fsck(self, tests: Optional[list[str]] = None) -> None:
         - "asset-yaml": loads each assets and checks if it's valid YAML
         - "asset-validity": loads each asset and validates the contents against
           the validation rulesets defined in ``.onyo/validation/``.
+        - "no-pseudo-keys": verifies that assets do not contain pseudo-key names
```
aqw, line 438: IMO, the `no` should be dropped. It's a check for pseudo keys.

Have "no" in the name confuses the fact that "yes" we are doing the check.

onyo/lib/onyo.py
```diff
@@ -510,6 +512,32 @@ def _fsck_clean_tree(self) -> bool:
...
+        """
+        assets_with_pseudo_keys = {}
+        pseudo_keys = ["type", "make", "model", "serial"]
+
+        for asset in self.assets:
+            contents = self._read_asset(asset)
+
+            for p in pseudo_keys:
+                if p in contents:
+                    if asset not in assets_with_pseudo_keys.keys():
```
aqw, line 527: No need for `.keys()` here. Simply: `if asset not in assets_with_pseudo_key:`

Nit: I find `assets_with_pseudo_keys.keys()` to be odd to read. Perhaps a more generic name, such as `assets_failed` or `assets_violation`?

onyo/lib/onyo.py
```diff
@@ -510,6 +512,32 @@ def _fsck_clean_tree(self) -> bool:
...
         return True
 
+    def _fsck_no_pseudo_keys(self) -> bool:
+        """
+        Check that no asset contains any pseudo-key names.
+        """
+        assets_with_pseudo_keys = {}
+        pseudo_keys = ["type", "make", "model", "serial"]
+
+        for asset in self.assets:
```
aqw, line 522: The rest of the for-loop can be simplified to:
```
violation_list = [x for x in pseudo_keys if x in self._read_asset(asset)]
if violation_list:
    assets_with_pseudo_keys[asset] = violation_list
```

onyo/lib/onyo.py
```diff
@@ -510,6 +512,32 @@ def _fsck_clean_tree(self) -> bool:
...
+            contents = self._read_asset(asset)
+
+            for p in pseudo_keys:
+                if p in contents:
+                    if asset not in assets_with_pseudo_keys.keys():
+                        assets_with_pseudo_keys[asset] = []
+                    assets_with_pseudo_keys[asset].append(p)
+
+        if assets_with_pseudo_keys:
+            log.error('The following assets contain pseudo keys:\n' +
```
aqw, line 532: This should be even more explicit. The user may not know what a pseudo key is, or that it is not allowed. Currently this message just states a fact. At minimum it should add "which are reserved for use in the asset filename."

onyo/lib/onyo.py
```diff
@@ -510,6 +512,32 @@ def _fsck_clean_tree(self) -> bool:
...
+                      self._n_join([f'{asset}: ' +
+                                    ', '.join([str(k) for k in assets_with_pseudo_keys[asset]])
```
aqw, line 534: `k` is already a string; No need to re-cast.

`assets_with_pseudo_keys[asset]` is a list not a dict.

`[str(k) for k in assets_with_pseudo_keys[asset]]` can simply be `assets_with_pseudo_keys[asset]`.

onyo/lib/onyo.py
```diff
@@ -510,6 +512,32 @@ def _fsck_clean_tree(self) -> bool:
...
+                                    for asset in list(assets_with_pseudo_keys.keys())]))
```
aqw, line 535: `.keys()` is not necessary to loop over a dict. And if it were, `.keys()` would not need to be cast into a list in order to iterate over it.

Needless casting is a common error across these PRs. Please only do so when you know that it is necessary.

tests/lib/test_Repo_fsck.py
```diff
@@ -151,6 +150,33 @@ def test_fsck_yaml_invalid(caplog, repo):
         assert i in caplog.text
 
 
+contents = ["type: value",
```
aqw, line 153: Please also add a test where the "reserved" pseudo key is in the value, and thus should not fail. For example:

```
key: serial
```
and the more perverse:
```
key: "model: value"
```

aqw (changes requested):

TobiasKadelka:
I added the feedback (except the one I commented on) and pushed again, thanks.
I especially liked the other test-group of things that should still be allowed!

onyo/lib/onyo.py
aqw: Not that this looks to be a costly part of the code, but the list comprehension also reduces the number of assignments and appends dramatically.

TobiasKadelka:
Well, I added/updated it now anyways. Not sure if the difference is relevant, but until we have performance checks for large numbers of assets, there is also no harm in optimizing a little bit. Thanks!

aqw (approved):
LGTM!

PR #311, opened 2023-03-15
BF/TST: check PATH is in repo for `onyo tree PATH`; add `test_tree.py`

### LINKED ISSUE #121: tree: checks cwd for onyo-repository, but not the path to display
If outside of an onyo repository and using `onyo tree .` it gives an error that the directory is not onyo (as wished). But if inside an onyo repository, the user can e.g. do `onyo tree ..` to pass a path into the function which is not onyo, which will be displayed then without an error.

This is low effort to change, but also not imediately important, since onyo still just uses the classic `tree` command, without doing any color highlighting or anything else onyo specific. However, since there are ideas about making commands like `onyo tree` stronger in some form, this should at some point in the future be updated.

The solutions could be:
1. check for the directory to display correctly, not just if inside an onyo repository
2. remove checks in general and allow all paths to be passed into `onyo tree`, since right now there is nothing that would get lost, instead it would just allow the command to always work and be not onyo specific.

TobiasKadelka:
`onyo tree PATH` did test that the opdir is in the repository, but did not test if the PATH given as input is in the repository, which I fixed.

To verify that the fix does not introduce new bugs, I added `test_tree.py`, because it seems we never had any tests at all for this command.

Close #121

### DIFF
--- a/onyo/commands/tree.py
+++ b/onyo/commands/tree.py
@@ -9,7 +9,7 @@ logging.basicConfig()
 log = logging.getLogger('onyo')
 
 
-def sanitize_directories(directories: list, opdir: str) -> list[str]:
+def sanitize_directories(repo: Repo, directories: list[str]) -> list[str]:
     """
     Check a list of directories. If any do not exist or are a file, and error
     will be printed.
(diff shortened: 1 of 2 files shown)


onyo/commands/tree.py
```diff
@@ -18,18 +18,26 @@ def sanitize_directories(directories: list, opdir: str) -> list[str]:
...
         else:
-            error_path.append(d)
+            no_path.append(d)
 
-    if error_path:
-        print('The following paths are not directories:\n' + '\n'.join(error_path),
-              file=sys.stderr)
+    if error_path or no_path:
+        if error_path:
+            print('The following paths are not in the Repository:\n' +
```
aqw, line 36: Nit: `Repository` should not be capitalized.

In general, I prefer when errors are explicit about what is wrong.

The issue is not "that the path is not in the repo". It's that all paths must be in the repo.

aqw (commented):
Looks good overall. I left a few nits.

I did not review the tests, as I'm unsure about how `tree` should be implemented going forward (keep as-is, vs pass through like `config`, vs replace with a python implementation). But that ambiguity is not a blocker to these initial tests.

onyo/commands/tree.py
TobiasKadelka: Thanks, I thought about a few ways of stating this more explicitly, please let me know what you think :)

TobiasKadelka:
I updated the variable names and the output of `tree.py` and updated this PR!

I am not sure about tree as well, and we discussed the idea in the past to reimplement it in python.
But I think until then there is still reason to verify that the internal call of tree does the right thing.

And I want to state that I thought about these possible changes while implementing the tests, and I think when implementing something else, the tests will provide still reasonable test-szenarios to provide helpful feedback while I am updating `onyo tree` then :)

onyo/commands/tree.py
```diff
@@ -9,27 +9,37 @@
...
     """
     Check a list of directories. If any do not exist or are a file, and error
     will be printed.
 
     Returns a string of the valid directories relative to opdir.
     """
     dirs = []
-    error_path = []
+    path_not_in_repo_error = []
+    no_directory_error = []
```
aqw, line 21: Sorry to nit. Please feel free to say if I'm descending into bike shedding.

I just see that these two vars basically do the same thing, yet the naming convention is completely different, which then (incorrectly) implies that they are quite far apart in utility.

Perhaps something similar to:
```
error_path_not_in_repo = []
error_path_not_dir = []
```

TobiasKadelka:
I nitted the new variable names into the fabric of `tree.py`, thanks for looking into all the PRs! :)

Feel free to merge, when the tests where successful.

onyo/commands/tree.py
aqw: I'm comfortable with the variables being separate. IMO, that's a matter of choice, regarding how specific the error messages should be. I have no problems with the direction you chose here.

+1 on the renaming for consistency. Thanks. :-)

aqw (approved):
LGTM!

PR #313, opened 2023-03-20
demo: check OS to not fail test on macOS

TobiasKadelka:
This is bothering me for quite a while, because when running all tests locally with pytest, it fails just the demo test, and then it does not run flake8 and pyre (when all called together with `&&`).


demo/generate_demo_repo.sh
```diff
@@ -12,7 +12,11 @@ set -e
 ############
 readonly VERSION=1.0.0
 readonly SCRIPT_NAME=${0##*/}
-readonly SCRIPT_DIR=$(dirname $(realpath -e "$0"))
+if [[ "$OSTYPE" =~ ^darwin ]]; then
+    readonly SCRIPT_DIR=$(dirname $(realpath "$0"))
+else
+    readonly SCRIPT_DIR=$(dirname $(realpath -e "$0"))
```
aqw, line 18: IMO, we can simply drop this variant. `-e` requires that all components of the path exist. Normal behavior is to enforce everything /except/ the last component.

In this case, we /know/ that all of the components exist (it's coming from `$0` after all).

aqw (changes requested):

PR #315, opened 2023-03-23
onyo.py: remove unneeded variable `Repo._gitfiles`

### LINKED ISSUE #314: Justify existence of `Repo._gitfiles`
The title of the issue is a bit of provoking clickbaiting, it might have a reason I did not see, that existed in the past, or that is planned to be added.

However, for #308 we decided to explain specifically why we clear which cache whenever we do in the repo-class.
I have no idea why this cache exists, nothing seems to ever modify it at all - I can't find it in commands, the repo class, the tests, no where. This made it hard to think about it and the `Repo.clear_cache()` might need to be updated when this variable becomes more clear.

Anyways - since it seems that this cache is always `None`, I decided to clean it always, since resetting it does not really cost runtime, and it is not loaded elsewhere anyways, a.k.a. it is also no loss on that front.

TobiasKadelka:
No need for the variable was identified.
close #314

### DIFF
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -34,7 +34,6 @@ class Repo:
         self._assets = None
         self._dirs = None
         self._files = None
-        self._gitfiles = None
         self._templates = None
 
     @property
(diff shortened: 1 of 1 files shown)


aqw (approved):
LGTM!

PR #321, opened 2023-03-31
Consistency: add typing to commands; update default values and add some typing in main.py

### LINKED ISSUE #319: FIX type inconsistency in argparse defaults
The arguments given in onyo commands are expected to be of a particular type, but the defaults do not always conform to this type. For example:

https://github.com/psyinfra/onyo/blob/5902cb16d78859e11473a339709697d0048b27b2/onyo/main.py#L528-L534

Here, `nargs='*'` ensures that if `--path` is used in `onyo set`, it returns a list of strings, however the default is a string. Considering strings are valid iterables (iterating over characters) this may lead to unexpected bugs. It is best to provide the default value as a list, enforcing type consistency.

The above quoted line is one occurrence of this, but we should thoroughly check `main.py` to avoid such inconsistencies everywhere.

TobiasKadelka:
Adds typing `argparse.Namespace` to all commands.
Adds typing to some functions in `main.py`.
Changes default values in argparsing in `main.py` to have consistent types (i.e. lists of `Path`).

Close #319

### DIFF
--- a/onyo/commands/cat.py
+++ b/onyo/commands/cat.py
@@ -1,8 +1,13 @@
+from __future__ import annotations
 import logging
 import sys
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 from onyo import Repo, OnyoInvalidRepoError
+
+if TYPE_CHECKING:
+    import argparse
 
 logging.basicConfig()
 log = logging.getLogger('onyo')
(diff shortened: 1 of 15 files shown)


onyo/commands/cat.py
```diff
@@ -1,3 +1,4 @@
+import argparse
```
nhjjr, line 1: Instead of importing `argparse`, consider the following (if `argparse` is only used for type annotations):

```
from __future__ import annotations
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import argparse

def cat(args: argparse.Namespace, opdir: str) -> None:
```

This only loads argparse when type checking.

PR #324, opened 2023-04-04
README: add section "Code Conventions"

TobiasKadelka:
This should briefly explain the conventions to follow for contributions of the Onyo project, including the code base under lib and what should be tested for, for outside contributers, but also us, currently and especially in the future.

I chose to re-organize part of the existing readme, and then add under "Develop" (which did not have contents so far) the conventions I assume to be fitting for onyo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,46 @@ cd onyo
 pip install -e ".[tests, docs]"
 ```
 
-### Tests
+### Code Conventions
+
+Onyo follows a set of development and coding conventions throughout its code
+base. For linting and type checking, Onyo uses flake8 and Pyre. The following
+sections describe conventions followed in this project, additionally to the
+[PEP 8 Style Guide](https://peps.python.org/pep-0008/).
+
+#### The Complete Code Base:
+- All classes, functions and properties (including tests) have a doc-string that
+  describes their functionality.
+- All functions have type hinting (including explicit return type for `None`).
+- The code conforms to [PEP 8](https://peps.python.org/pep-0008/).
+
+#### lib Tests (`tests/lib/`):
+- define the expected behavior for all use-cases of the different consumers
+  (i.e. CLI, demos)
+- use the API functions (no usage of `subprocess.run(["onyo", ...])`)
+- all public functions and properties of all classes are tested with all
+  possible combinations of input:
+  - single valid inputs, including different data-types (if allowed)
+  - multiple valid inputs (if allowed)
+  - without input (if allowed)
+  - one test for each possible error
+  - one test for each possible conflict with other inputs
+
+#### CLI Tests (`tests/commands/`):
+- define the expected behavior which the user can expect from Onyo
+- are created completely independent and without any consideration of the code
+  base
+- cover a large variety of use-cases to verify that Onyo works as intended,
+  even if large parts of the code would be modified
+- call `subprocess.run(["onyo", ...])` to simulate how Onyo behaves
+- verify the correct behavior for all possible combinations of arguments/flags:
+  - single valid argument
+  - multiple valid arguments (if allowed)
+  - without the flag
+  - one test per different error
+  - one test per possible conflict with other arguments
+
+### Run Tests
 Tests are run from the top-level of the repository.
 ```
 pytest -vv
--- a/README.md
+++ b/README.md
@@ -14,17 +14,6 @@ be just a thin wrapper around git commands.
 See the documentation for [installation instructions](https://onyo.readthedocs.io/en/latest/installation.html)
 and general information about Onyo.
 
-## Demo
-An [example Onyo repository](https://github.com/psyinfra/onyo-demo/) is
-available. It's easier to get a feel for how Onyo works with a populated
-repository with actual history, rather than starting from scratch. Just install
-Onyo, clone the demo repo, and start poking around!
-
-## Report Issues
-For general feedback, bug reports, and comments, please [open an issue](https://github.com/psyinfra/onyo/issues/new).
-
-## Develop
-
 ### Installation
 
 #### Non-Python Dependencies:
@@ -46,7 +35,18 @@ python -m venv ~/.venvs/onyo
 source ~/.venvs/onyo/bin/activate
 ```
 
-#### Clone the repo and install package and dependencies for tests, documentation:
+## Demo
+An [example Onyo repository](https://github.com/psyinfra/onyo-demo/) is
+available. It's easier to get a feel for how Onyo works with a populated
+repository with actual history, rather than starting from scratch. Just install
+Onyo, clone the demo repo, and start poking around!
+
+## Report Issues
+For general feedback, bug reports, and comments, please [open an issue](https://github.com/psyinfra/onyo/issues/new).
+
+## Develop
+
+### Clone the repo and install package and dependencies for tests, documentation:
 ```
 git clone https://github.com/psyinfra/onyo.git
 cd onyo


README.md
```diff
@@ -53,7 +53,42 @@ cd onyo
...
+base. For linting and type checking, the Onyo project uses flake8 and Pyre.
+
+#### the complete code base
+- All classes, functions and properties have a doc-string describing their
+  functionality.
+- All functions have type hinting (including explicit return type for `None`).
+- The code is PEP8 conform.
+
+#### lib Tests (`onyo/lib/`)
+- they follow Onyos code conventions, see `"the complete code base"
```
aqw, line 68: There's a stray backtick here.

What is meant by 'see "the complete code base"'? Do you mean that they should familiarize themselves with the surrounding code and follow its conventions?

README.md
```diff
@@ -53,7 +53,42 @@ cd onyo
...
+- The tests proof stability of the repository class and the rest of the code to
+  offer a stable base for the consumers of their code (i.e. CLI, demos)
+- all public functions of all classes are tested:
```
aqw, line 71: Public properties should also be tested. I can't think of a good word that covers all. Perhaps "Public endpoints"?

README.md
```diff
@@ -53,7 +53,42 @@ cd onyo
...
+  - for all possible existing input arguments
+    - single valid inputs used
+    - multiple valid inputs used (if allowed)
+    - valid test without input (if allowed)
+    - one test for each possible error
+    - one test for each possible conflict with other inputs
+
+#### CLI Tests (`tests/`)
+- they follow Onyos code conventions, see `"the complete code base"
```
aqw, line 80: Stray backtick

Onyos -> Onyo's

README.md
```diff
@@ -53,7 +53,42 @@ cd onyo
...
+- The tests define Onyos behavior the user can expect, and they are created
```
aqw, line 81: As a style suggestion, to reduce redundancy, each bullet point can start with a verb:

For example:

CLI tests:
- follow [...]
- define [...]
- etc

README.md
TobiasKadelka: No, I meant to refer to the section five lines above `#### the complete code base`.
I will try to link to it, thanks for highlighting that this is ambigous when reading it ;)

README.md
TobiasKadelka: Nah, I feel like trying to kill these to birds with one stone does not really bring the enhancement in clearity we are looking for. I think I will just add the words "public properties"!

README.md
```diff
@@ -53,7 +53,42 @@ cd onyo
...
+  functionality.
+- All functions have type hinting (including explicit return type for `None`).
+- The code is PEP8 conform.
+
+#### lib Tests (`onyo/lib/`)
+- they follow Onyos code conventions, see `"the complete code base"
+- The tests proof stability of the repository class and the rest of the code to
+  offer a stable base for the consumers of their code (i.e. CLI, demos)
+- all public functions of all classes are tested:
+  - for all possible existing input arguments
```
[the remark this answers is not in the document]
TobiasKadelka: I mean "a function has x different arguments in the line `def func(arg a, arg b)`, test for all to work correctly, including optional arguments, different datatypes, etc". I will try to rephrase it, though I found it a bit tricky to describe this straight forward.

README.md
aqw: :facepalm: I /swear/ I looked for a section called "The complete code base".

That said, capitalizing it would make it more obvious that it's a title.

README.md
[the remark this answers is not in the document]
aqw: The word "existing" is what threw me. Perhaps something as simple as "all arguments"

README.md
```diff
@@ -53,7 +53,47 @@ cd onyo
 pip install -e ".[tests, docs]"
 ```
 
-### Tests
+### Design
+
+Onyo follows a set of development and coding conventions throughout its code
+base. For linting and type checking, the Onyo project uses flake8 and Pyre. The
```
aqw, line 59: "the Onyo project" can be removed. Otherwise it should be said everywhere, and will be verbose.

README.md
```diff
@@ -53,7 +53,47 @@ cd onyo
...
+following sections describe additional conventions followed in this project,
```
aqw, line 60: Nit: I don't see what "additional" adds here.

README.md
```diff
@@ -53,7 +53,47 @@ cd onyo
...
+extending the existing [PEP 8 Style Guide](https://peps.python.org/pep-0008/).
```
aqw, line 61: "Extending" sounds like we've written an extension to PEP 8. I think that's a bold claim, beyond the scope of what's written here. ;-)

README.md
```diff
@@ -53,7 +53,47 @@ cd onyo
...
+
+#### The Complete Code Base:
+- All classes, functions and properties have a doc-string describing their
```
aqw, line 64: Nit-of-nits: when possible, avoid progressive. e.g. "to describe" rather than "describing"

README.md
```diff
@@ -53,7 +53,47 @@ cd onyo
...
+  functionality.
+- All functions have type hinting (including explicit return type for `None`).
+- The code conforms to [PEP 8](https://peps.python.org/pep-0008/).
+
+#### lib Tests (`tests/lib/`):
+- follow Onyo's code conventions, see [The Complete Code Base](#the-complete-code-base).
```
aqw, line 70: This feels redundant, as the above is already "The Complete Code Base". Repeating in each section that "all" means "all" distracts.

If you're concerned that people won't consider tests to be code... then those people are just beyond our help. ;-)

README.md
```diff
@@ -53,7 +53,47 @@ cd onyo
...
+- define the expected behavior of the code for all use-cases of the different
```
aqw, line 71: nit: "of the code" is a generic statement. It's ok to omit it, and the sentence is clearer as a result:

> define the expected behavior for all use-cases of the different consumers

However, I also struggle with the word "define" here. The /definition/ should be in the function's docstring. The tests should make sure that definition is well exercised.

README.md
```diff
@@ -53,7 +53,47 @@ cd onyo
...
+  consumers (i.e. CLI, demos).
+- all public functions and properties of all classes are tested with all
+  possible combinations of input:
```
aqw, line 74: "all possible combinations" is a dangerous phrase. And impossible to meet in a literal sense.

README.md
```diff
@@ -53,7 +53,47 @@ cd onyo
...
+  - single valid inputs, including different data-types (if allowed)
+  - multiple valid inputs (if allowed)
+  - without input (if allowed)
+  - one test for each possible error
+  - one test for each possible conflict with other inputs
+
+#### CLI Tests (`tests/commands/`):
+- follow Onyo's code conventions, see [The Complete Code Base](#the-complete-code-base).
+- define the expected behavior which the user can expect from Onyo.
```
aqw, line 83: nit: these are sentence fragments (like above), and thus should not have a period at the end

README.md
```diff
@@ -53,7 +53,47 @@ cd onyo
...
+- are created completely independent and without any consideration of the code
+  base.
+- cover a large variety of use-cases to verify that Onyo works as intended,
+  even if large parts of the code would be modified.
```
aqw, line 87: I see what you mean here, but I find this particular phrasing non-obvious. What I think makes it more clear is stating the objectives/values. That tests should not be brittle. they should not require updating when the functionality it tests is refactored. Tests are meant to accelerate development, not inhibit it. Etc.

README.md
```diff
@@ -53,7 +53,47 @@ cd onyo
...
+- call `subprocess.run(["onyo", ...])` to simulate how Onyo behaves.
```
aqw, line 88: This feels vague here. I think what you mean is that the Onyo API should not be used when testing the CLI, and exclusively onyo CLI commands should be used.

If so, then IMO the inverse of the rule should be added to the API tests.

README.md
TobiasKadelka: Well, that was at least my concern... :P

README.md
TobiasKadelka: While "define" might not be the perfect term, I feel like "make sure that definition from the function's doc-strings is well exercised" just does not sound that clear.

And I want to mention that the tests are implemented as independently as possible from the code, which is were the docstring is. I know that this might be different for the public functions of the repo etc, but still I feel like "define" is at least ok.

README.md
TobiasKadelka: I think in general the usage of the API should not be prohibited during the CLI tests, except for the one specific command tested, i.e. setting some config value with `Repo.config()` in `commands/test_set.py` seems a reasonable thing to do to really just test the functionality of `onyo set` and not entangle it with other commands.

Though you are right, I will add the opposite to the API tests!

README.md
TobiasKadelka: Yes, and yet I chose to write it like this anyways. :)

The truth is, currently we also do not have everywhere doc strings, type hinting, and all the other things of these categories.
I personally think it is more fitting to state the goals aimed at once clearly, instead of giving a too weak description.

aqw (approved):
LGTM. :-)

PR #327, opened 2023-04-05
(to be deleted) TST: modernize `tests/lib/`

TobiasKadelka:
I just wanted to add type hints to the functions missing it, but obviously I could not hold back, so I also added doc-strings when they were missing, and moved definition of `variants` variables into the parameterization so as to make the tests more stable when run in random order.

### DIFF
--- a/tests/lib/test_Repo.py
+++ b/tests/lib/test_Repo.py
@@ -10,12 +10,12 @@ from onyo import Repo, OnyoInvalidRepoError
 #
 # instantiation
 #
-variants = {
-    'str': 'the-repo',
-    'Path': Path('the-repo'),
-}
-@pytest.mark.parametrize('variant', variants.values(), ids=variants.keys())
-def test_Repo_instantiate_types(tmp_path, variant):
+@pytest.mark.parametrize('variant', ['the-repo', Path('the-repo')])
+def test_Repo_instantiate_types(tmp_path: str, variant: Union[Path, str]) -> None:
+    """
+    Test that the Repo class instantiates correctly for different data types
+    describing paths to existing repositories.
+    """
     os.chdir(tmp_path)
     ret = subprocess.run(['onyo', 'init', 'the-repo'])
     assert ret.returncode == 0
@@ -24,20 +24,32 @@ def test_Repo_instantiate_types(tmp_path, variant):
     Repo(variant)
 
 
-def test_Repo_instantiate_invalid_path(tmp_path):
+def test_Repo_instantiate_invalid_path(tmp_path: str) -> None:
+    """
+    Test that the Repo class raises the correct error if instantiated with a
+    path that does not exist.
+    """
     repo_path = Path(tmp_path, 'does-not-exist')
 
     with pytest.raises(FileNotFoundError):
         Repo(repo_path)
 
 
-def test_Repo_instantiate_empty_dir(tmp_path):
+def test_Repo_instantiate_empty_dir(tmp_path: str) -> None:
+    """
+    Test that the Repo class raises the correct error if instantiated with a
+    path that is not an Onyo repository.
+    """
     repo_path = Path(tmp_path)
     with pytest.raises(OnyoInvalidRepoError):
         Repo(repo_path)
 
 
-def test_Repo_instantiate_git_no_onyo(tmp_path):
+def test_Repo_instantiate_git_no_onyo(tmp_path: str) -> None:
+    """
+    Test that the Repo class raises the correct error if used on a git
+    repository that is not an Onyo repository.
+    """
     repo_path = Path(tmp_path)
     ret = subprocess.run(['git', 'init', str(repo_path)])
     assert ret.returncode == 0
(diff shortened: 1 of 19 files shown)


tests/lib/test_Repo.py
```diff
@@ -10,12 +10,12 @@
 #
 # instantiation
 #
-variants = {
-    'str': 'the-repo',
-    'Path': Path('the-repo'),
-}
-@pytest.mark.parametrize('variant', variants.values(), ids=variants.keys())
-def test_Repo_instantiate_types(tmp_path, variant):
+@pytest.mark.parametrize('variant', ['the-repo', Path('the-repo')])
```
aqw, line 13: Retaining `ids` helps the readability of the output of `pytest` when tests are failing. It doesn't know how to convert `Path('the-repo')` to a string.

tests/lib/test_Repo.py
```diff
@@ -10,12 +10,12 @@
...
+def test_Repo_instantiate_types(tmp_path: str, variant: Union[Path, str]) -> None:
+    """
+    Test that the Repo class instantiates correctly for different data types
+    describing paths to existing repositories.
```
aqw, line 17: nit: "describing" is unnecessary

tests/lib/test_Repo.py
```diff
@@ -10,12 +10,12 @@
...
-variants = {
-    'str': 'the-repo',
-    'Path': Path('the-repo'),
-}
-@pytest.mark.parametrize('variant', variants.values(), ids=variants.keys())
-def test_Repo_instantiate_types(tmp_path, variant):
+@pytest.mark.parametrize('variant', ['the-repo', Path('the-repo')])
+def test_Repo_instantiate_types(tmp_path: str, variant: Union[Path, str]) -> None:
+    """
+    Test that the Repo class instantiates correctly for different data types
```
aqw, line 16: Nit: the phrase "Test that" will be repeated for every docstring. This is unnecessary. We're in the tests. We can accept that it's testing functionality.

For example, rather than "Test that the range is higher than 10." simply state: "The range must be higher than 10."

The tests in many ways are a code version of the specification. Imagine that the tests/docstrings are written based off the spec. Let them be declarative.

tests/lib/test_Repo.py
```diff
@@ -24,20 +24,32 @@ def test_Repo_instantiate_types(tmp_path, variant):
     Repo(variant)
 
 
-def test_Repo_instantiate_invalid_path(tmp_path):
+def test_Repo_instantiate_invalid_path(tmp_path: str) -> None:
+    """
+    Test that the Repo class raises the correct error if instantiated with a
```
aqw, line 29: Going along the "spec" idea: which error?

For example:

> Repo must raise `FileNotFoundError` when instantiated with a non-existent path.

Or

> Repo must be instantiated with an existing directory. Otherwise, raise `FileNotFoundError`.

tests/lib/test_Repo_add.py
```diff
@@ -1,19 +1,18 @@
...
-    'Path': Path('single-file'),
-    'list-str': ['single-file'],
-    'list-Path': [Path('single-file')],
-    'set-str': {'single-file'},
-    'set-Path': {Path('single-file')},
-}
-@pytest.mark.parametrize('variant', variants.values(), ids=variants.keys())
-def test_add_args_single_file(repo, variant):
+@pytest.mark.parametrize('variant', [
+    'single-file', Path('single-file'), ['single-file'], [Path('single-file')],
```
aqw, line 9: IMO, this is easier to read if it's one item per line (as it was before). It is more verbose, but much easier to read.

tests/lib/test_Repo_commit.py
```diff
@@ -47,12 +51,12 @@ def test_commit_title_and_description(repo, variant):
...
-    'list-str': ['one', 'two', 'three'],
-    'list-Path': [Path('one'), Path('two'), Path('three')],
-}
-@pytest.mark.parametrize('variant', variants.values(), ids=variants.keys())
-def test_commit_list(repo, variant):
+@pytest.mark.parametrize('variant', [['one', 'two', 'three'],
+                                     [Path('one'), Path('two'), Path('three')]])
+def test_commit_list(repo: Repo, variant: List[Union[str, Path]]) -> None:
+    """
+    Test that `Repo.commit()` excepts a list as input type.
```
aqw, line 58: excepts -> accepts

tests/lib/test_Repo_commit.py
```diff
@@ -79,26 +83,26 @@ def test_commit_set(repo, variant):
...
 
-def test_commit_nothing_staged(repo):
+def test_commit_nothing_staged(repo: Repo) -> None:
+    """
+    Test that `Repo.commit()` raises the correct error when called while no
+    files are staged to commit.
+    """
     with pytest.raises(subprocess.CalledProcessError):
-        repo.commit('We believe in nothing Lebowski!')
+        repo.commit('We believe in nothing, Lebowski!')
```
aqw, line 92: LMAO

tests/lib/test_Repo_mv.py
```diff
@@ -1,21 +1,19 @@
...
 from pathlib import Path
+
 import pytest
-from onyo import commands  # noqa: F401
+from _pytest.logging import LogCaptureFixture
 from onyo import OnyoProtectedPathError
-
-
+from onyo.lib import Repo
+from typing import List
```
aqw, line 8: This import should be higher up, since it's not 3rd party.

aqw (changes requested):
I reviewed about half of this, and then the rest was repeating the same stuff, so i stopped reading.

Overall, this looks great.

My primary concerns are:
- while adding docstrings is a big improvement, I do feel that they are not as terse nor specific as they should be. Many still require to read the test code to understand the desired behavior.
- Dropping `ids` hurts readability. However, I found a nice solution to allow specifying them as the keys of a dictionary (as was common prior, but this PR removes): https://github.com/pytest-dev/pytest/issues/7568#issuecomment-665966315

Overall, a nice step forward to uplifting everything.

PR #330, opened 2023-04-13
(RFC) lib: properties/paths are absolute, git requests are made from repo.root

### LINKED ISSUE #297: Add tests for `Repo.generate_commit_message()`
Since in a future version of onyo all paths internally are handled as absolute paths, and all paths in commit messages should be given relative to the repository root, the tests for this should especially verify that `repo.root` is not in the messages (for generated messages of different lengths).

TobiasKadelka:
Changes added:
- all properties are now absolute paths
- all `git` queries (e.g. `git ls-files`) are run from `repo.root` to contain all assets/dirs etc from the repo, not just relative to `repo.opdir`, and all are absolute
- adds functions:
  - `Repo.sanitize_path()`: checks if paths are absolute or relative to the opdir and returns them as an absolute `Path`
  - `Repo.relative_to_root()`: returns for a set of absolute paths (that must be inside of the repository) a set of Pathsrelative to `repo.root`
  - `relative_to_opdir()`: returns for a set of absolute paths as set of Paths relative to the operating directory
- updates the generating of commit messages to still display paths relative from `repo.root`
- updates the tests/commands which are effected by the changes to work with the changes

Close #297

Additionally: 
- This PR currently still does not remove `Repo.opdir`, but since the helper functions work without it, and instead enable easy usage with absolute paths, I would remove it, if this is the way we want to go.

### DIFF
--- a/onyo/commands/get.py
+++ b/onyo/commands/get.py
@@ -127,14 +127,24 @@ def get(args: argparse.Namespace, opdir: str) -> None:
     except OnyoInvalidRepoError:
         sys.exit(1)
 
-    # validate arguments
-    paths = set(Path(p) for p in args.path) or {Path('.')}
-    invalid_paths = {p for p in paths if not p.exists()}
-    paths -= invalid_paths
-    if any(invalid_paths):
-        for path in invalid_paths:
-            print(
-                f"cannot access '{path}': No such directory", file=sys.stderr)
+    # validate path arguments
+    paths = set()
+    invalid_paths = set()
+    for path in args.path:
+        try:
+            onyo_path = repo.sanitize_path(path)
+            if onyo_path.exists():
+                paths.add(onyo_path)
+            else:
+                invalid_paths.add(path)
+        except ValueError:
+            invalid_paths.add(path)
+    if invalid_paths:
+        err_str = '\n'.join([str(x) for x in invalid_paths])
+        print(f"The following paths do not exist:\n{err_str}", file=sys.stderr)
+        sys.exit(1)
+    if not paths:
+        print("No assets selected.", file=sys.stderr)
         sys.exit(1)
 
     if args.depth < 0:
@@ -160,7 +170,7 @@ def get(args: argparse.Namespace, opdir: str) -> None:
         sep = '\t'  # column separator
         for asset, data in results:
             values = sep.join([str(value) for value in data.values()])
-            print(f'{values}{sep}{asset}')
+            print(f'{values}{sep}{asset.relative_to(opdir)}')
     else:
         console = Console()
         table = Table(
@@ -175,7 +185,7 @@ def get(args: argparse.Namespace, opdir: str) -> None:
         if results:
             for asset, data in results:
                 values = [str(value) for value in data.values()]
-                table.add_row(*values, str(asset))
+                table.add_row(*values, str(asset.relative_to(opdir)))
 
             console.print(table)
         else:
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -57,8 +57,7 @@ class Repo:
     @property
     def assets(self) -> set[Path]:  # pyre-ignore[11]
         """
-        A `set` containing the `Path`s relative to `Repo.root` of all assets of
-        an Onyo repository.
+        A `set` containing the absolute `Path`s of all assets of a repository.
 
         This property is cached, and the cache is consistent with the state of
         the repository when only `Repo`s public functions are used. Use of
@@ -73,8 +72,7 @@ class Repo:
     @property
     def dirs(self) -> set[Path]:
         """
-        A `set` containing the `Path`s relative to `Repo.root` of all directories
-        of an Onyo repository.
+        A `set` containing the absolute `Path`s of all directories a repository.
 
         This property is cached, and the cache is consistent with the state of
         the repository when only `Repo`s public functions are used. Use of
@@ -89,8 +87,7 @@ class Repo:
     @property
     def files(self) -> set[Path]:
         """
-        A `set` containing the `Paths` relative to `Repo.root` of all files of an
-        Onyo repository.
+        A `set` containing the absolute `Paths` of all files of a repository.
 
         This property is cached, and the cache is consistent with the state of
         the repository when only `Repo`s public functions are used. Use of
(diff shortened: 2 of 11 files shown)


onyo/lib/onyo.py
```diff
@@ -296,49 +329,56 @@ def _get_assets(self) -> set[Path]:
...
 
     def _get_files_untracked(self) -> set[Path]:
         """
-        Return a set of all untracked files in the repository.
+        Return a set of all absolute `Path`s to untracked files in the
+        repository.
         """
         log.debug('Acquiring list of untracked files')
-        untracked = {Path(x) for x in self._git(['ls-files', '-z', '--others', '--exclude-standard']).split('\0') if x}
+        untracked = {Path(self.root, x) for x in self._git(['-C', str(self.root), 'ls-files', '-z', '--others', '--exclude-standard']).split('\0') if x}
```
bpoldrack, line 375: This may be for later when `opdir` is removed, but FTR:

I'd just make `_git` always call `subprocess.run` with `cwd=self.root` rather than passing `-C str(self.root)` everywhere.

bpoldrack (commented):
Thx, @TobiasKadelka , I left a couple of suggestions.

One more comment regarding tests:
You adjusted tests that apparently want relative paths, but there seem to be no tests, that actually test your changes (that the returned paths are in fact the expected ones).
IOW: Instead of something like

`assert Path('repeat-one') in repo.relative_to_root(repo.files_staged)`

I would have done:

`assert (repo.root / 'repeat-one') in repo.files_staged`

Note, that in the former case, the test may pass while `files_staged` delivers relative paths, when CWD is the root. In the latter case, you actually test, that `files_staged` returns absolute paths as intended. (Of course you'd also want to test `repo.root` somewhere - didn't look whether that is already done somehwere)

bpoldrack:
> I propose allow to instantiate Repo(<path>) just with a <path> that exists and is an onyo root. To enable easy handling, I would split the finding of "repo.root" that is currently done into its own helper function, so that to init a Repository from a random location inside a repository, it must be instantiated like:
repo = Repo(Repo.get_root(<path>)), and <path> could for example be the opdir in the current commands.

Why not:

```

class Repo(object):

    def __init__(path)
        self.root = self.get_root(path)

```

instead of forcing the caller to use `repo = Repo(Repo.get_root(<path>))`?

onyo/lib/onyo.py
```diff
@@ -266,6 +263,41 @@ def _generate_commit_message_subject(
...
+
+        return onyo_path
+
+    def relative_to_root(
+            self, paths: Union[set[Path], list[Path]]) -> set[Path]:
+        """
+        Expects a list or set of absolute paths in the repository and returns a
+        set of paths relative to the repository root.
+        """
+        return {x.relative_to(self.root) for x in paths}
```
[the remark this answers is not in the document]
TobiasKadelka: I agree anyways, and changed it to just accept sets, because that is what we work with so far, and I think until we discuss the types we want to handle were, we should start with being more specific, as long as it is not clear that something more specific is needed!
So thanks for bringing up the set/list and str/Path issue :D

TobiasKadelka:
> instead of forcing the caller to use repo = Repo(Repo.get_root(<path>))?

This point was raised by you and Alex, and you both are right. I added it and updated it in its seperate PR https://github.com/psyinfra/onyo/pull/334!

onyo/commands/get.py
```diff
@@ -127,14 +127,23 @@ def get(args: argparse.Namespace, opdir: str) -> None:
...
+    paths = set()
+    invalid_paths = set()
+    for path in args.path:
+        try:
+            onyo_path = repo.sanitize_path(path)
+            if not onyo_path.exists():
+                raise ValueError
+            paths.add(onyo_path)
+        except ValueError:
+            invalid_paths.add(path)
```
bpoldrack, line 140: This is a strange construct. Why not like this (see below)?

```suggestion
        onyo_path = repo.sanitize_path(path)
        if not onyo_path.exists():
            invalid_paths.add(path)
        else:
            paths.add(onyo_path)
```

onyo/lib/onyo.py
```diff
@@ -260,6 +257,37 @@ def _generate_commit_message_subject(
...
+        function can be used to get paths to create new assets or move to
+        another location.
+
+        Returns an absolute `Path`.
+        """
+        onyo_path = Path(path).resolve()
+
+        # TODO: raise a more reasonable error
+        if not onyo_path.is_relative_to(self.root):
+            raise ValueError("path is not relative to root.")
```
bpoldrack, line 273: Minor suggestion, because `sanitize_paths` takes absolute paths, too. So, to error out with "not relative to ..." seems to suggest to a caller, that a relative path is required input, when really it's about not being part of the repo.

```suggestion
            raise ValueError(f"path {path} outside repository at {self.root}.")
```

bpoldrack (commented):
Overall, fine with that. Thx, @TobiasKadelka.

Slightly irritated, why this PR needs to introduce a test for generating commit messages (seems to be a different topic), but ok.

One thing is off across tests, though, and that is this kind of usage of the `/` operator for `Path` objects:

`assert Path(repo.root / asset).is_file()`

While this kinda works, it's unnecessary. `repo.root` is a `Path` object already, right?
So `assert (repo.root / asset).is_file()` it is. There's no need to call a `Path` constructor again. And if `repo.root` was in fact a string it would still be:

`assert (Path(repo.root) / asset).is_file()`.

Or better yet, to have an easy to read assertion:

```
asset_path = repo.root / asset
# alternative if repo.root was string:
# asset_path = Path(repo.root) / asset
assert asset_path.is_file()
```

onyo/commands/get.py
```diff
@@ -127,14 +127,24 @@ def get(args: argparse.Namespace, opdir: str) -> None:
...
+    for path in args.path:
+        try:
+            onyo_path = repo.sanitize_path(path)
+            if onyo_path.exists():
+                paths.add(onyo_path)
+            else:
+                invalid_paths.add(path)
+        except ValueError:
+            invalid_paths.add(path)
     if any(invalid_paths):
```
bpoldrack, line 142: I know, it's not actually part of the PR, but just stumbled upon it.
The `any` construct does nothing here.

```suggestion
    if invalid_paths:
```

bpoldrack (approved):
Thx, @TobiasKadelka.

Only left one minor suggestion, otherwise good to go, I think.

PR #331, opened 2023-04-13
Uplifting

TobiasKadelka:
This contains the "uncontroversial" changes from #327 and #328, as well as it adds the feedback from the review of #327.

It adds a class for the definition of variants.

### DIFF
--- a/tests/lib/test_Repo.py
+++ b/tests/lib/test_Repo.py
@@ -5,17 +5,22 @@ from pathlib import Path
 
 import pytest
 from onyo import Repo, OnyoInvalidRepoError
+from tests.conftest import params
 
 
 #
 # instantiation
 #
-variants = {
-    'str': 'the-repo',
-    'Path': Path('the-repo'),
-}
-@pytest.mark.parametrize('variant', variants.values(), ids=variants.keys())
-def test_Repo_instantiate_types(tmp_path, variant):
+@params({
+    "Path": {"variant": Path("the-repo")},
+    "str": {"variant": "the-repo"},
+})
+def test_Repo_instantiate_types(
+        tmp_path: str, variant: Union[str, Path]) -> None:
+    """
+    The Repo class must instantiate correctly for different data types for
+    paths to existing repositories.
+    """
     os.chdir(tmp_path)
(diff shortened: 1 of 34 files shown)


tests/lib/test_Repo.py
```diff
@@ -6,16 +6,27 @@
...
 # instantiation
 #
-variants = {
-    'str': 'the-repo',
-    'Path': Path('the-repo'),
-}
-@pytest.mark.parametrize('variant', variants.values(), ids=variants.keys())
-def test_Repo_instantiate_types(tmp_path, variant):
+@params({
+        "Path": {"variant": Path("the-repo")},
```
aqw, line 21: Nit: in later tests I notice you just indent these normally (with 4 spaces). I prefer that convention over this.

tests/lib/test_Repo.py
```diff
@@ -6,16 +6,27 @@
 import pytest
 from onyo import Repo, OnyoInvalidRepoError
 
+def params(d):
```
aqw, line 9: This should be added to `conftest.py` and registered as a fixture. That way it doesn't need to be added to each of the test files.

aqw (changes requested):
This looks /great/.

The docstrings are much better, and I /really/, /really/ like this `params()` fixture. Best of all worlds IMO.

I'd like to see `params()` moved to `conftest.py` and registered, and also that the new fixture is somehow documented so that other develops know to use it.

tests/lib/test_Repo_config.py
```diff
@@ -12,84 +14,87 @@
...
-    'default': {
+
+@params({
+    'default': {"variant": {
         'args': ('onyo.test.set-config', 'default'),
-        'file': '.onyo/config',
+        'file': '.onyo/config'},
     },
-    'local': {
+    'local': {"variant": {
```
aqw, line 23: Not so much a nit as rather a note for the future: in this case, `variant` isn't needed. Instead, `args` and `file` can be each passed to the test function.

No need to change it here. Just saying. :-)

aqw (approved):
Looks great. I agree that this is a large, beastly PR. ;-)

Happy to merge. :-)

PR #332, opened 2023-04-17
Remove Repo._n_join and refactor its use

nhjjr:
This PR addresses the use of `Repo._n_join`, replacing it with `'\n'.join(map(str, to_join))`. I've also addressed some other string formatting issues for methods I've touched with this change. There are more string formatting issues to address throughout the codebase, but addressing those here would make the PR very large.

There is one commit on refactoring `Repo._fsck_unique_assets()`, which was done to both make the code more readable, but also make the subsequent error logging more straightforward.

### DIFF
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -639,21 +639,14 @@ class Repo:
         return True
 
     def _fsck_unique_assets(self) -> bool:
-        """
-        Check if all files have unique names. Returns True or False.
-        """
-        names = {}
-        for f in self.assets:
-            try:
-                names[f.name].append(f)
-            except KeyError:
-                names[f.name] = [f]
+        """Check if all asset files have unique names."""
+        asset_names = [a.name for a in self.assets]
+        duplicates = [a for a in self.assets if asset_names.count(a.name) > 1]
+        duplicates.sort(key=lambda x: x.name)
 
-        if len(self.assets) != len(names):
-            log.error('The following file names are not unique:\n' +
-                      '\n'.join([str(y) for x in names for y in names[x]
-                                 if len(names[x]) > 1]))
-
+        if duplicates:
+            log.error('The following file names are not unique:\n{}'.format(
+                '\n'.join(map(str, duplicates))))
             return False
 
         return True
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -204,7 +204,7 @@ class Repo:
 
         if message:
             message_subject = message[0][0]
-            message_body = self._n_join([x[0] for x in message[1:]])
+            message_body = '\n'.join(map(str, [x[0] for x in message[1:]]))
         else:
             # get variables for the begin of the commit message `msg_dummy`
             dest = None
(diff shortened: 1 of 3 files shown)


onyo/lib/onyo.py
```diff
@@ -624,15 +622,18 @@ def _fsck_pseudo_keys(self) -> bool:
...
 
         if assets_failed:
-            log.error(f"Pseudo keys ({', '.join(pseudo_keys)}) are reserved for asset names, and are not allowed in asset files. The following assets contain pseudo keys:\n" +
-                      self._n_join([f'{asset}: ' +
-                                    ', '.join([k for k in assets_failed[asset]])
-                                    for asset in assets_failed]))
+            log.error(
+                'Pseudo keys {0} are reserved for asset names, and are not '
+                'allowed in asset files. The following assets contain pseudo '
+                'keys:\n{1}'.format(tuple(pseudo_keys), '\n'.join(
```
aqw, line 634: IMO, this is more readable if the second item passed to .`format()` is on the next line.

Purely a matter of taste. Feel free to disregard if you disagree.

onyo/lib/onyo.py
```diff
@@ -812,25 +815,31 @@ def _mkdir_sanitize(self, dirs: Iterable[Union[Path, str]]) -> set[Path]:
...
+                    '\n'.join(map(str, error_path_protected))))
 
         return dirs_to_create
 
     #
     # MV
     #
-    def mv(self, sources: Union[Iterable[Union[Path, str]], Path, str], destination: Union[Path, str], dryrun: bool = False) -> list[tuple[str, str]]:
+    def mv(
+            self, sources: Union[Iterable[Union[Path, str]], Path, str],
```
aqw, line 840: Why is this on a new line? This can be on the same line as `def mv(`

aqw (approved):
I like this PR. Makes a lot of stuff easier to read. +1

I especially like the refactoring of `_fsck_unique_assets()`. Much more obvious.

I've made a few nits. None are blockers though. Address or disregard as you wish.

onyo/lib/onyo.py
nhjjr: I have no problem with this, and I agree that it looks better. Changed.

onyo/lib/onyo.py
nhjjr: For a method as short as `mv` it doesn't make much of a difference, but for methods with longer names this pushes line-breaks further to the right. For example:

```
def this_has_a_really_long_name(self, foo, bar, baz,
                                lorem, ipsum, dolor,
                                sit, amet) -> bool:
    # code goes here
    return True
```
vs
```
def this_has_a_really_long_name(
        self, foo, bar, baz, lorem, ipsum, dolor, 
        sit, amet) -> bool:
    # code goes here
    return True
```

The reason to do this with `mv` is consistency.

onyo/lib/onyo.py
```diff
@@ -624,15 +622,18 @@ def _fsck_pseudo_keys(self) -> bool:
         pseudo_keys = ["type", "make", "model", "serial"]
 
         for asset in self.assets:
-            violation_list = [x for x in pseudo_keys if x in self._read_asset(asset)]
+            violation_list = [
```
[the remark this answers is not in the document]
TobiasKadelka: I feel like this is adressed to me, too, so I will consider this in the future more, too.

aqw:
LGTM!

onyo/lib/onyo.py
TobiasKadelka: I know this PR is just merged, but want to add my two cents anyways:

I think there should not be the consistent rule "break lines for functions because they are sometimes long", but rather we should aim at choosing consistently the easiest readable solution.

I myself am still on the search for when to break lines and when not to do that (see some of my last few PRs), but in this case I think it is obviously not an improvement.

PR #333, opened 2023-04-17
BF: add `tests/__init__.py` to enable importing from conftest

TobiasKadelka:
The `__init__.py` is needed to make functions and variables inside of the project explicitely import-able, otherwise the CI errors with a `ModuleNotFoundError`.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,7 @@
+from tests.conftest import params, Helpers
+
+
+__all__ = [
+    params,
+    Helpers,
+]


tests/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+from tests.conftest import params, Helpers
+
+
+logging.basicConfig(level=logging.ERROR)  # external logging level
+log: logging.Logger = logging.getLogger('onyo')  # internal logging level
+log.setLevel(level=logging.INFO)
+
+__all__ = [
+    params,
+    Helpers
```
nhjjr, line 11: Add a `,` at the end here

tests/__init__.py
```diff
@@ -0,0 +1,12 @@
+import logging
+from tests.conftest import params, Helpers
+
+
+logging.basicConfig(level=logging.ERROR)  # external logging level
```
nhjjr, line 5: I don't see the tests using logging at the moment (unless I'm missing something). We should probably not add this here unless (and until) we plan to add logging to the tests

aqw (approved):
IMO, this solves the immediate problem and unblocks the other PRs.

It buys time to discuss what the most appropriate fix is.

tests/__init__.py
TobiasKadelka: Lol, fair enough, though I want to add for the record, it does not have to literally be everywhere in the end, I was just again removing them in the end where they were already existing ;)

But I added it, so thanks for having a look!

PR #334, opened 2023-04-21
`Repo` must be instantiated with root

TobiasKadelka:
To remove ambiguity in situations were a `Repo` object is instantiated with a path that is a subdirectory inside a repository, and a function of `Repo` is called with a relative path, this PR updates the `Repo` class so as to only accept the root of an onyo repository.

- Updates `Repo` to either just allow root as path to instantiate, or to use `Repo(find_root=True)` to search the root from a subdirectory inside a repo
- Adds `Repo._find_root()` to allow the easy identification of the root of an existing repository when called from a subdirectory of it
- Applies changes to commands
- Updates existing `tests/lib/test_Repo.py` to be conform with new behavior
- Adds tests for `Repo._find_root()`

This is the first step in the direction of working only with absolute paths in Repo, and to retire `Repo.opdir`.

### DIFF
--- a/onyo/commands/cat.py
+++ b/onyo/commands/cat.py
@@ -60,7 +60,7 @@ def cat(args: argparse.Namespace, opdir: str) -> None:
     """
     repo = None
     try:
-        repo = Repo(opdir)
+        repo = Repo(opdir, find_root=True)
         repo.fsck(['asset-yaml'])
     except OnyoInvalidRepoError:
         sys.exit(1)
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -33,13 +33,15 @@ class Repo:
         path is invalid a `OnyoInvalidRepoError` is raised.
         """
         path = Path(path)
-        path = self._find_root(path) if find_root else path.resolve()
 
         if init:
+            path = path.resolve()
             self._init(path)
-        elif not Repo._is_onyo_root(path):
-            log.error(f"'{path}' is no valid Onyo Repository.")
-            raise OnyoInvalidRepoError(f"'{path}' is no valid Onyo Repository.")
+        else:
+            path = self._find_root(path) if find_root else path.resolve()
+            if not Repo._is_onyo_root(path):
+                log.error(f"'{path}' is no valid Onyo Repository.")
+                raise OnyoInvalidRepoError(f"'{path}' is no valid Onyo Repository.")
 
         self._opdir: Path = path
         self._root: Path = path
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -20,12 +20,29 @@ class OnyoProtectedPathError(Exception):
 class Repo:
     """
     """
-    def __init__(self, path: Union[Path, str], init: bool = False) -> None:
+    def __init__(self, path: Union[Path, str], find_root: bool = False,
+                 init: bool = False) -> None:
+        """
+        Instantiates a `Repo` object with `path` as the root directory.
+
+        If `find_root=True` searches the root of a repository from `path`.
+
+        If `init=True`, the `path` will be initialized as a git repo and a
+        `.onyo/` directory will be created.
+        Otherwise the validity of the onyo repository is verified, and if the
+        path is invalid a `OnyoInvalidRepoError` is raised.
+        """
+        path = Path(path)
+        path = self._find_root(path) if find_root else path.resolve()
+
         if init:
             self._init(path)
+        elif not Repo._is_onyo_root(path):
+            log.error(f"'{path}' is no valid Onyo Repository.")
+            raise OnyoInvalidRepoError(f"'{path}' is no valid Onyo Repository.")
 
-        self._opdir: Path = Path(path).resolve()
-        self._root: Path = self._get_root()
+        self._opdir: Path = path
+        self._root: Path = path
         # caches
         self._assets: Union[set[Path], None] = None
         self._dirs: Union[set[Path], None] = None
@@ -349,8 +366,7 @@ class Repo:
 
         if not Path(root, '.onyo').is_dir():
             log.error(f"'{root}' is not an Onyo repository.")
-            raise OnyoInvalidRepoError(f"'{directory}' is not an Onyo repository.")
-
+            raise OnyoInvalidRepoError(f"'{root / '.onyo'}' is missing or not a directory.")
         # TODO: check .onyo/config, etc
 
         log.debug(f"Onyo repo found at '{root}'")
@@ -573,7 +589,7 @@ class Repo:
     #
     # INIT
     #
-    def _init(self, directory: Union[Path, str]) -> None:
+    def _init(self, directory: Path) -> None:
         """
         Initialize an Onyo repository. The directory will be initialized as a
         git repository (if it is not one already), ``.onyo/`` directory created
@@ -585,6 +601,29 @@ class Repo:
         """
         from .command_utils import init_onyo
         return init_onyo(self, directory)
+
+    @staticmethod
+    def _is_onyo_root(directory: Path) -> bool:
+        """
+        Assert whether a directory is the root of a repository and has a fully
+        populated `.onyo/` directory.
+        """
+        dot_onyo = Path(directory, '.onyo')
+
+        if not dot_onyo.is_dir() or \
+                not Path(dot_onyo, "templates").is_dir() or \
+                not Path(dot_onyo, "validation").is_dir() or \
+                not Path(dot_onyo, "config").is_file() or \
+                not Path(dot_onyo, ".anchor").is_file() or \
+                not Path(dot_onyo, "templates/.anchor").is_file() or \
+                not Path(dot_onyo, "validation/.anchor").is_file():
+            return False  # noqa: E111, E117
+
+        if subprocess.run(["git", "rev-parse"], cwd=directory,
+                          stdout=subprocess.DEVNULL).returncode != 0:
+            return False
+
+        return True
 
     @staticmethod
     def _init_sanitize(directory: Union[Path, str]) -> Path:
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -28,13 +28,16 @@ class Repo:
         If `find_root=True` searches the root of a repository from `path`.
 
         If `init=True`, the `path` will be initialized as a git repo and a
-        `.onyo/` directory will be created.
+        `.onyo/` directory will be created. `find_root=True` must not be used
+        in combination with `init=True`.
         Otherwise the validity of the onyo repository is verified, and if the
         path is invalid a `OnyoInvalidRepoError` is raised.
         """
         path = Path(path)
 
         if init:
+            if find_root:
+                raise ValueError("`find_root=True` must not be used with `init=True`")
             path = path.resolve()
             self._init(path)
         else:
@@ -611,15 +614,10 @@ class Repo:
         populated `.onyo/` directory.
         """
         dot_onyo = Path(directory, '.onyo')
+        files = ['config', '.anchor', 'templates/.anchor', 'validation/.anchor']
 
-        if not dot_onyo.is_dir() or \
-                not Path(dot_onyo, "templates").is_dir() or \
-                not Path(dot_onyo, "validation").is_dir() or \
-                not Path(dot_onyo, "config").is_file() or \
-                not Path(dot_onyo, ".anchor").is_file() or \
-                not Path(dot_onyo, "templates/.anchor").is_file() or \
-                not Path(dot_onyo, "validation/.anchor").is_file():
-            return False  # noqa: E111, E117
+        if not all(x.is_file() for x in [dot_onyo / f for f in files]):
+            return False
 
         if subprocess.run(["git", "rev-parse"], cwd=directory,
                           stdout=subprocess.DEVNULL).returncode != 0:
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -327,19 +327,29 @@ class Repo:
         untracked = {Path(x) for x in self._git(['ls-files', '-z', '--others', '--exclude-standard']).split('\0') if x}
         return untracked
 
-    def _get_root(self) -> Path:
+    @staticmethod
+    def _find_root(directory: Path) -> Path:
         """
-        """
-        try:
-            root = self._git(['rev-parse', '--show-toplevel'], cwd=self._opdir).strip()
-        except subprocess.CalledProcessError:
-            log.error(f"'{self._opdir}' is not a Git repository.")
-            raise OnyoInvalidRepoError(f"'{self._opdir}' is not a Git repository.")
+        Find and return the root of an Onyo repository (containing `.git/` and
+        `.onyo/`) for a given `Path` inside of an existing repository.
 
-        root = Path(root)
+        If the `directory` is not inside an existing repository, an error is
+        raised.
+        """
+        root = None
+
+        try:
+            ret = subprocess.run(["git", "rev-parse", "--show-toplevel"],
+                                 cwd=directory, check=True,
+                                 capture_output=True, text=True)
+            root = Path(ret.stdout.strip()).resolve()
+        except (subprocess.CalledProcessError, FileNotFoundError):
+            log.error(f"'{directory}' is not a Git repository.")
+            raise OnyoInvalidRepoError(f"'{directory}' is not a Git repository.")
+
         if not Path(root, '.onyo').is_dir():
             log.error(f"'{root}' is not an Onyo repository.")
-            raise OnyoInvalidRepoError(f"'{self._opdir}' is not an Onyo repository.")
+            raise OnyoInvalidRepoError(f"'{directory}' is not an Onyo repository.")
 
         # TODO: check .onyo/config, etc
 
--- a/tests/lib/test_Repo.py
+++ b/tests/lib/test_Repo.py
@@ -465,6 +465,10 @@ def test_Repo_opdir_child(repo: Repo, tmp_path: str) -> None:
     with pytest.raises(OnyoInvalidRepoError):
         repo = Repo('.')
 
+    # test
+    repo = Repo('.', find_root=True)
+    assert Path(tmp_path).samefile(repo.root)
+
 
 #
 # Repo.root
@@ -514,25 +518,48 @@ def test_Repo_root_root(tmp_path: str) -> None:
     repo = Repo('.')
     assert Path('.').samefile(repo.root)
 
+#
+# Repo._find_root()
+#
 
-def test_Repo_root_child(tmp_path: str) -> None:
+@pytest.mark.repo_dirs('1/2/3/4/5/6')
+def test_Repo_find_root_opdir_root(repo: Repo) -> None:
     """
-    The property repo.root must be set to the real root of the repository, when
-    Repo() is instantiated in a sub-directory of a repository.
+    When the cwd is in the repo root, `Repo._find_root()` just returns the root.
+    a `Repo` can be instantiated with `Repo(<opdir>, find_root=True)`.
     """
-    os.chdir(tmp_path)
+    os.chdir(repo.root)
+    new_repo = Repo(repo.root, find_root=True)
+    assert new_repo.root.samefile(repo.root)
 
-    # setup repo
-    ret = subprocess.run(['onyo', 'init', 'root-child'])
-    assert ret.returncode == 0
-    os.chdir('root-child')
-    ret = subprocess.run(['onyo', 'mkdir', '--yes', '1/2/3/4/5/6'])
-    assert ret.returncode == 0
 
-    # test
-    os.chdir('1/2/3/4/5/6')
-    repo = Repo('.')
-    assert Path('../../../../../../').samefile(repo.root)
+@pytest.mark.repo_dirs('1/2/3/4/5/6')
+def test_Repo_find_root_opdir_child(repo: Repo) -> None:
+    """
+    When the cwd is inside an existing repo, a `Repo` can be instantiated with
+    `Repo(<opdir>, find_root=True)`.
+    """
+    os.chdir(Path(repo.root, '1/2/3/4/5/6'))
+    new_repo = Repo(Path(repo.root, '1/2/3/4/5/6'), find_root=True)
+    assert new_repo.root.samefile(repo.root)
+
+
+def test_Repo_find_root_with_no_repository_path(tmp_path: str) -> None:
+    """
+    If `Repo._find_root()` is called on an existing path that is not a
+    repository, it has to raise a `OnyoInvalidRepoError`
+    """
+    with pytest.raises(OnyoInvalidRepoError):
+        Repo._find_root(Path(tmp_path))
+
+
+def test_Repo_find_root_with_non_existing_path(tmp_path: str) -> None:
+    """
+    If `Repo._find_root()` is called on a path that does not exist, it has to
+    raise a `OnyoInvalidRepoError`
+    """
+    with pytest.raises(OnyoInvalidRepoError):
+        Repo._find_root(Path(tmp_path, "does_not_exist"))
 
 
 #
--- a/tests/lib/test_Repo.py
+++ b/tests/lib/test_Repo.py
@@ -133,6 +133,16 @@ def test_init_not_exist_dir(tmp_path: Path, variant: str) -> None:
     assert fully_populated_dot_onyo(repo_path)
 
 
+def test_init_and_find_root_true(tmp_path: Path) -> None:
+    """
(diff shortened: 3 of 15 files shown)


onyo/commands/cat.py
```diff
@@ -60,7 +60,8 @@ def cat(args: argparse.Namespace, opdir: str) -> None:
     """
     repo = None
     try:
-        repo = Repo(opdir)
+        repo_root = Repo.find_root(opdir)
```
aqw, line 63: IMO, I envision `_find_root()` as a private method for `Repo`, and the user does not interact with it directly.

Instead the syntax would be like:
```
repo = Repo(opdir, search=True)
```
Then if they want to know what the repo root is, they do it they same was as always:
```
repo.root
```

The name `search` could be better. It's just what came to mind in this moment.

onyo/lib/onyo.py
```diff
@@ -26,12 +27,32 @@ class OnyoProtectedPathError(Exception):
...
+        If `init=True`, the `path` will be initialized as a git repo and a
+        `.onyo/` directory will be created.
+        Otherwise the validity of the onyo repository is verified.
+        """
+        path = Path(path)
+
+        if find_root:
+            path = self._find_root(path)
+        else:
+            path = Path(path).resolve()
```
nhjjr, line 46: Coding style, but for brevity I'd suggest:

```python
path = self._find_root(path) if find_root else Path(path).resolve()
```

bpoldrack, line 46: `path` is already ensured above to be a `Path` object.

```suggestion
            path = path.resolve()
```

onyo/lib/onyo.py
```diff
@@ -337,19 +358,29 @@ def _get_files_untracked(self) -> set[Path]:
         untracked = {Path(x) for x in self._git(['ls-files', '-z', '--others', '--exclude-standard']).split('\0') if x}
         return untracked
 
-    def _get_root(self) -> Path:
+    @staticmethod
+    def _find_root(directory: Union[Path, str]) -> Path:
```
bpoldrack, line 362: Is there a reason, `_find_root` needs to be able to take a `str`?
I do understand, that this is how it's done everywhere, but it's relatively costly to call `Path`'s constructor over and over.

onyo/lib/onyo.py
```diff
@@ -337,19 +358,29 @@ def _get_files_untracked(self) -> set[Path]:
...
         """
+        Find and return the root of an Onyo repository (containing `.git/` and
+        `.onyo/`) for any path inside of an existing repository.
+
+        If the `directory` is not inside an existing repository, an error is
+        raised.
         """
+        directory = Path(directory).resolve()
+        root = Path()
```
bpoldrack, line 371: Why initialize `root` with `'.'`?
In case you're thinking that `Path()` is some sort of empty path - it's not:

```
In [1]: from pathlib import Path

In [2]: Path()
Out[2]: PosixPath('.')
```

So, if code below works as expected, the initial value is never relevant at all. If it isn't, we end up treating CWD as root, because it happens to have a `.onyo` subdir. Point being: This can only go wrong but doesn't provide a benefit as far as I can tell. Or is there a case where CWD is intended to be some sort of fallback?

bpoldrack (commented):
Thx, @TobiasKadelka.

Overall, I have one general remark (or may be rather a question):

You set `find_root`'s default to `False` and then change pretty much every call throughout the code base to call `Repo(find_root=True)`. This already suggests that the default should be the other way around.
Furthermore - and here is the question part - I don't know what exactly are the plans for `opdir` removal, but isn't the ultimate aim to not have it and make `find_root=True` mandatory? I imagine that eventually this parameter should not exist at all. And that would be the second reason to set the default to `True` now (in case there is a place somewhere that needs `False`) and not change the calls to `Repo(path)` - they can then stay this way all along.

nhjjr (commented):
If I understand this correctly, initializing the `Repo` class can be done in 3 different ways:

  * `init=True`: A new repository is created using `Repo.path` as the root
  * `init=False, find_root=False`: The `Repo.path` is assumed to be the root
  * `init=False, find_root=True`: The `Repo.path` is used to find the actual root

I don't see the purpose of `find_root`. If `init` is set to `False`, then we should assume that `Repo.path` is the root. I would suggest a different approach:

Initialize the `Repo` using `Repo(root: Path, init: bool = False)`. It is now clear the path we want is the root of the repository. We then have two choices:
  * Move the `find_root` function to the CLI and make it the CLI's responsibility to provide the right information to the `Repo` class
  * Be lenient, and automatically 'sanitize' the given `root`. If it is not the root, log a warning saying the root path has been derived from the given root. If it cannot be derived, raise an exception. If it is the actual root, forego the warning.

The `Repo._find_root()` method can then be part of the `Repo.root` property. You create a setter for this property that provides the given root from `__init__()` and derives the repo root from it, as described above. That is, when you do `self.root = Path('root/foo/bar')`, it returns `Path('root')`. It is then cached to always return the actual root Path and becomes immutable after initialization.

If `init` is set to `False`, you initializes the repository and set it as the new `Repo.root`. It doesn't matter that the setter for `Repo.root` checks for the actual root in the repository; it should come up with the same path. If not, then something is wrong.

I also don't understand why `Repo._find_root` is private. I suggest keeping `Repo.root` non-private, because from an API perspective it is useful information to work with (even if just as a sanity check).

aqw:
>  I don't know what exactly are the plans for opdir removal, but isn't the ultimate aim to not have it and make find_root=True mandatory

Jumping in quickly here to add my 2 cents:

`opdir` and `find_root` are related, but ultimate independent justifications.

IMO, the goal is certainly  not to have `find_root` eventually default to `True`. For me that change is about ergonomics. A user who does `Repo('repo/subdir') will expect that all operations happen from `subdir` (like on the CLI). This should throw an error and say it's not a valid Onyo Repo. `Repo('repo/')` will succeed. If the user wants Onyo to help them find repo root, then they can ack into this feature, and use `Repo('repo.subdir', find_root=True)`. They explicitly ack into the behavior, making it less of a surprise to them.

The retirement of `opdir` is related (for the obvious reasons), but it really more about an overall simplification of the pathing (see Tobias's other PR on absolte paths, etc).

onyo/lib/onyo.py
```diff
@@ -26,12 +27,32 @@ class OnyoProtectedPathError(Exception):
 class Repo:
     """
     """
-    def __init__(self, path: Union[Path, str], init: bool = False) -> None:
+    def __init__(self, path: Union[Path, str], find_root: bool = False,
```
nhjjr, line 30: What happens in the case both `find_root` and `init` are set to `True`? Since there is no use-case for it, it appears that when `init=False`, we assume `path` is the repo-root, and when `init=True` we assume `path` will become the repo root by virtue of initializing a repository. The only reason `find_root` is here, is to raise an error when it is not True and `path` is not the repo root?

bpoldrack:
@aqw 

Ok, I don't really agree with the alleged expectation, but I get the idea - so the default is fine for now, then.


@nhjjr 
> - init=False, find_root=False: The Repo.path is assumed to be the root

I think, this is not intended to be the case. `_is_onyo_root` is supposed to fail then (according to docstring). But this points to a bug, @TobiasKadelka: `is_onyo_root` doesn't do what it says it does: It doesn't care about there being a repo at all, it only checks the `.onyo` subdir. Hence, when `find_root=False` and `init=False`, we'd end up *assuming* path to be the root of a repository whenever it has a `.onyo` subdir.

onyo/lib/onyo.py
```diff
@@ -26,12 +27,32 @@ class OnyoProtectedPathError(Exception):
...
+        path = Path(path)
+
+        if find_root:
+            path = self._find_root(path)
+        else:
+            path = Path(path).resolve()
+
         if init:
             self._init(path)
+        elif not Repo._is_onyo_root(path):
```
nhjjr, line 50: Why is this calling `Repo._is_onyo_root(path)`, rather than `self._is_onyo_root(path)`? This just implies to me that this method should not be part of the `Repo` class.

bpoldrack: For the record:
While I do think, that this should eventually go elsewhere, I do think so for a different reason.
Just because it's static doesn't necessarily imply it needs to go. And do find it useful (and even better) to call a static method via the class rather than the instance (`self`). Makes it clear when reading the calling code that there's no dependence on the actual instance w/o needing to lookup the called method.

bpoldrack (approved):
Thanks, @TobiasKadelka.
Overall fine with me. Just minor comments. I'd fine with ignoring them (but then transfer them into issues instead, please.

tests/lib/test_Repo.py
```diff
@@ -521,25 +518,48 @@ def test_Repo_root_root(tmp_path: str) -> None:
...
+    assert new_repo.root.samefile(repo.root)
+
+
+def test_Repo_find_root_with_no_repository_path(tmp_path: str) -> None:
+    """
+    If `Repo._find_root()` is called on an existing path that is not a
+    repository, it has to raise a `OnyoInvalidRepoError`
+    """
+    with pytest.raises(OnyoInvalidRepoError):
+        Repo._find_root(Path(tmp_path))
```
aqw, line 553: Which makes this unnecessary to cast.

tests/lib/test_Repo.py
```diff
@@ -521,25 +518,48 @@ def test_Repo_root_root(tmp_path: str) -> None:
...
+
+
+def test_Repo_find_root_with_non_existing_path(tmp_path: str) -> None:
```
aqw, line 556: Here as well. Etc.

tests/lib/test_Repo.py
```diff
@@ -521,25 +518,48 @@ def test_Repo_root_root(tmp_path: str) -> None:
...
+    """
+    When the cwd is inside an existing repo, a `Repo` can be instantiated with
+    `Repo(<opdir>, find_root=True)`.
+    """
+    os.chdir(Path(repo.root, '1/2/3/4/5/6'))
+    new_repo = Repo(Path(repo.root, '1/2/3/4/5/6'), find_root=True)
+    assert new_repo.root.samefile(repo.root)
+
+
+def test_Repo_find_root_with_no_repository_path(tmp_path: str) -> None:
```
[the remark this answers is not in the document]
TobiasKadelka: Thanks, how annoying that Pire did not catch that. Will update.

onyo/lib/onyo.py
```diff
@@ -20,12 +20,29 @@ class OnyoProtectedPathError(Exception):
...
+        path is invalid a `OnyoInvalidRepoError` is raised.
+        """
+        path = Path(path)
+        path = self._find_root(path) if find_root else path.resolve()
+
         if init:
             self._init(path)
+        elif not Repo._is_onyo_root(path):
+            log.error(f"'{path}' is no valid Onyo Repository.")
+            raise OnyoInvalidRepoError(f"'{path}' is no valid Onyo Repository.")
```
bpoldrack, line 42: Change the order of treating `init` and `find_root`?
The issue is, that there's no point in failing on `Repo('some/path', init=True, find_root=True)`, because `find_root` doesn't find anything. That's actually fine.

```suggestion
        path = Path(path)
        
        if init:
            path = path.resolve()
            self._init(path)
        else:        
            path = self._find_root(path) if find_root else path.resolve()
            if not Repo._is_onyo_root(path):
                log.error(f"'{path}' is no valid Onyo Repository.")
                raise OnyoInvalidRepoError(f"'{path}' is no valid Onyo Repository.")
```

bpoldrack (commented):
Somehow I thought this was addressed before, but I think there's still an issue with `Repo()` being called with both `init=True` and `find_root=True`. This shouldn't fail. (And it's not tested ;-) ).

onyo/lib/onyo.py
TobiasKadelka: Thanks! I added your suggestion, am currently running the tests locally, and then will push again! :)

bpoldrack (approved):
Thx!

onyo/lib/onyo.py
nhjjr: I don't think this is resolved yet. The case where `init=True, find_root=True` is functionally equivalent to `init=True, find_root=False`. If both are `True`, it should therefore raise an exception.

Alternatively, we could use `init: Literal['from_path', 'from_root'] = 'from_path'`, because both arguments revolve around how the repository is loaded: (1) from the given path, (2) from the root found from the given path, (3) as a new repository in the given path.

onyo/lib/onyo.py
```diff
@@ -327,20 +346,29 @@ def _get_files_untracked(self) -> set[Path]:
...
+
         try:
-            root = self._git(['rev-parse', '--show-toplevel'], cwd=self._opdir).strip()
-        except subprocess.CalledProcessError:
-            log.error(f"'{self._opdir}' is not a Git repository.")
-            raise OnyoInvalidRepoError(f"'{self._opdir}' is not a Git repository.")
+            ret = subprocess.run(["git", "rev-parse", "--show-toplevel"],
+                                 cwd=directory, check=True,
+                                 capture_output=True, text=True)
+            root = Path(ret.stdout.strip()).resolve()
```
nhjjr, line 364: Minor nitpick: L364 should be outside of the try/except block, because the only thing evaluated is the subprocess call.

onyo/lib/onyo.py
```diff
@@ -576,6 +604,29 @@ def _init(self, directory: Union[Path, str]) -> None:
...
+        """
+        dot_onyo = Path(directory, '.onyo')
+
+        if not dot_onyo.is_dir() or \
+                not Path(dot_onyo, "templates").is_dir() or \
+                not Path(dot_onyo, "validation").is_dir() or \
+                not Path(dot_onyo, "config").is_file() or \
+                not Path(dot_onyo, ".anchor").is_file() or \
+                not Path(dot_onyo, "templates/.anchor").is_file() or \
+                not Path(dot_onyo, "validation/.anchor").is_file():
```
nhjjr, line 621: This can be shortened as:

```python3
dirs = ['templates', 'validation']
files = ['config', '.anchor', 'templates/.anchor', 'validation/.anchor']
has_dirs = [True if x.is_dir() else False for x in [dot_onyo / d for d in dirs]]
has_files = [True if x.is_file() else False for x in [dot_onyo / f for f in files]]

if any([*has_files, *has_dirs]):
    return False
```

onyo/lib/onyo.py
bpoldrack: I don't think so. It's not only about what is possibly raising the error, it's also about what is executed as a whole if there was no error. With that line being in the block, it's only ever executed when `ret` actually should contain something that this line can work with.
If you pull it out, you have branch twice for the same error:

1. try .. except
2. `if ret ... else`, which needs to "reconstruct" whether or not an exception was raised.

onyo/lib/onyo.py
```diff
@@ -576,6 +604,29 @@ def _init(self, directory: Union[Path, str]) -> None:
...
+            return False  # noqa: E111, E117
```
nhjjr, line 622: Why `# noqa: E111, E117`?

E111 is when indentation is not a multiple of four, and E117 is when something is over-indented. This can easily be addressed, and shouldn't even trigger on this line. If the shortened solution above is used, then this will become more obvious.

onyo/lib/onyo.py
bpoldrack: Hm. I can understand the viewpoint. But really, forcing the caller to figure out whether `find_root` makes sense, rather than just ignoring it, when it obviously doesn't, is an inconvenience. Don't just think of a caller who is literally passing `True` or `False`. This can be dynamic and relatively complex.

Ultimately, the whole problem is that the instantiation of the `Repo` class and calling `git init` is melted. Datalad started that way, too. In fact: *I* did. Seemed intuitive at first. Turned out to be an ever increasing mistake. But this is a different topic (as in: This particular PR shouldn't care).

tests/lib/test_Repo.py
```diff
@@ -453,30 +460,27 @@ def test_Repo_opdir_root(tmp_path: str) -> None:
...
+@pytest.mark.repo_dirs('1/2/3/4/5/6')
+def test_Repo_opdir_child(repo: Repo, tmp_path: Path) -> None:
     """
-    When instantiating the Repo with the cwd being inside a folder, the property
-    repo.opdir must be the cwd.
+    An existing Repo must be instantiated with the root of a repository,
+    otherwise an error is raised.
     """
-    os.chdir(tmp_path)
+    os.chdir(Path(repo.root, '1/2/3/4/5/6'))
```
nhjjr, line 469: Minor nitpick: Since `repo.root` is already a `Path` instance, you can use `repo.root / '1' / '2' / '3' / '3' / '4' / '5' / '6'. I believe Ben suggested this use with good reason.

onyo/lib/onyo.py
nhjjr: If we're going to address this topic at a later stage then I'm fine with this use-case, but I'd prefer to be explicit rather than implicit. Right now we're making the assumption that the user realizes `init` and `find_root` are mutually exclusive, and that when using both, only `init` is used and `find_root` is ignored. Such assumptions should generally be avoided.

onyo/lib/onyo.py
bpoldrack: Well, if we are at it, then:

1. `True if x.is_dir() else False` is superfluous. `is_dir()` returns a bool, so `[x.is_dir() for x ....` it is ;-)
2. Here is what I find even nicer to read:

```python3
dirs = ['templates', 'validation']
files = ['config', '.anchor', 'templates/.anchor', 'validation/.anchor']
has_dirs = all(x.is_dir() for x in [dot_onyo / d for d in dirs])
has_files = all(x.is_file() for x in [dot_onyo / f for f in files])

if not (has_files and has_dirs):
    return False
```

bpoldrack: Of course I don't care much whether it's `not (has_files and has_dirs)` or `not has_files or not has_dirs`

onyo/lib/onyo.py
nhjjr: What you're saying is case-specific. It's true for some cases, but not this one. If a `subprocess.CalledProcessError` or `FileNotFoundError` is raised, an exception is raised and the code will never reach `root`. In all other cases `root` will behave the same as it does inside the block. In this case it is better to be clear about what is being evaluated.

onyo/lib/onyo.py
aqw: A user could also think that `init` and `find_root` together means: find the root of the git repo above me and init that as an Onyo repo (Onyo allows init-ing a git repo).

I agree this discussion should happen in a different issue/PR. But the intended behavior should be outlined, and then usage made obvious.

onyo/lib/onyo.py
bpoldrack: +1

I didn't comment on it, because I haven't dived much into the linter setup yet, but I did wonder, too. If those errors are what @nhjjr is saying, then I agree.

onyo/lib/onyo.py
nhjjr: This is the kind of nitpicking I can get behind. Thanks for further improvements! 😄

onyo/lib/onyo.py
bpoldrack: What you are saying is: "Yes, this should only be executed if the previous line succeeded, but I want this line to be written in a piece of code, that doesn't know that to be the case. Instead, I - the dev - know, that at this particular moment of the development, the `except` clause raises again, thus implicitly making sure this has the same effect."

Is that assessment wrong?

onyo/lib/onyo.py
bpoldrack: I do see it the other way around, though, @aqw , @nhjjr :

With this change, the point of `find_root` is to only allow `True` and raise otherwise to "make the user consciously opt into the concept".

And now you are saying, that after being aware of that, I script a use of that, that yes - opts into always `True`. I am again screwed, because always `True` doesn't work. I must only ever call `init=True` with `find_root=False`, but I must never call `init=False` without `find_root=True`.

Instead of just declaring that `find_root` is always implied, unless `init=True` and getting rid of the entire `find_root` parameter, I as the caller not only have to understand the documentation of the function I also need to make my own code more complicated in order to reassure you guys, that I have read the doc.
The only valid and desired behavior is entirely implied by the `init` parameter. To me this feels strange.

bpoldrack: To be clear: Once I, as the caller, am clear about this. The only thing making sense in calling this dynamically - in my mind- is to say something like `Repo(path, init=initialize_repo, find_root=not intialize_repo)` to make this bearable in my usage.

nhjjr: I don't quite get your point, Ben. Let me state first that I'm with Ben and Alex on this topic being for a different PR, but I don't mind having the discussion here for now.

If I understand you correctly, the only reason `find_root` is added in this PR is as an opt-in for an approach that will become the default. If that is the case, we don't need any big discussions about it because ultimately `find_root` will go away and cause no more ambiguity. However, if as a user I provide `init=True, find_root=True` I expect something different to happen from `init=True, find_root=False` but that is not the case. At the very least I should be given a warning that there is no special case when both are True. If your code passes `init=True, find_root=True` when it only expects `init=True` anyway, then the code is confusing. It should be explicit in what it is doing, even if that requires another line of code.

If this is not the case and `find_root` is there to stay along with `init`, I suggest as a better fitting alternative to encapsulate both boolean arguments into one string literal. There are 3 approaches to instantiating `Repo`: (1) initialize a new repo in the given path, (2) use the given path as root, (3) find the root from the given path. 

There is no ambiguity with `load: Literal['from_path', 'from_root', 'init_new'] = 'from_path'` (contents of the string and name of the argument are mere examples). This way there are only 3 options for `load` and no ambiguity with a fourth option that works the same as the third option.

onyo/lib/onyo.py
nhjjr: I think that assessment is not addressing the full problem. Say we don't have that re-raise.
```python3
try:
    ret = some_code()
    root = ret.some_property
except SomeError:
   log.warning('something went wrong')

do_something(root)
```

In this case, things would crash on the last line if `some_code()` raises an exception. I would have to change things in the except block, so that we get:
```python3
try:
    ret = some_code()
    root = ret.some_property
except SomeError:
    root = None

do_something(root)
```

Now everything works fine. However, should we put `do_something(root)` also in the try/except block? It's using the contents of root, just like root is using the contents of ret, after all, and only I - the dev - know that in the except block we set it to `None` if an exception is raised. Note, though, that I think the second example is a valid use-case where `root` can be in the `try` block, because the `except` block specifically deals with it. In the current case, the try/except block doesn't deal with `root` at all (but placing it there gives the impression it does, which is the root (see what I did there 😆) of my problem).

Furthermore by this reasoning, `root = None` makes no sense, because `Path(None)` raises an exception. The only reason it's in there is to satisfy pyre, which doesn't understand that an exception is raised if `root` is not set.

onyo/lib/onyo.py
bpoldrack: > If I understand you correctly, the only reason find_root is added in this PR is as an opt-in for an approach that will become the default.

Default not planned as of now. I was referring to @aqw 's earlier clarification: https://github.com/psyinfra/onyo/pull/334#issuecomment-1525165133

> If as a user I provide init=True, find_root=True I expect something different to happen from init=True, find_root=False but that is not the case

I do understand that, but I don't see raising exceptions as part of "something different to happen" in this context.
With your suggestion there's exactly two valid ways of calling it: `init=True, find_root=False` and `init=False, find_root=True`. Everything else raises. 

> If your code passes init=True, find_root=True when it only expects init=True anyway, then the code is confusing

Well, that is because this function is confusing. It forces me to specify the inevitable. I need to set a parameter with zero degrees of freedom. If it was for `find_root` being ignored in case of `init`, then it would be less of a hassle, because I can still "acknowledge that I have seen the path handling pecularities of it" by always setting it to true.
Which is why I wanted go with that in this PR.

>  It should be explicit in what it is doing, even if that requires another line of code.

The thing is: The calling code doesn't need to explain what it's doing. It's not doing anything other than using the only valid way of instantiating a Repo from a path! Everything else is not about the calling code only about this function's insistence on being told that "Yes, I read the terms and conditions".

onyo/lib/onyo.py
bpoldrack: The problem with your version in this particular case for me is:

You have line `root = Path(ret.stdout.strip()).resolve()` whose validity entirely depends on the previous `try .. else` block, which it is not a part of. That's fine in this state of the code. The issue comes when other changes are introduced over time. Those to parts are semantically coupled, but not structurally. This can fall apart over time. At the same time: It does not provide any advantage over the existing solution. It can only make things worse in the long-run.

onyo/lib/onyo.py
bpoldrack: True.

bpoldrack:
Pushed a commit to address the review discussions. Otherwise, this is approved and passed the tests.
Hence: Merging.

Issue #340, opened 2023-04-25
Extract asset functions from `Repo`


bpoldrack:
Following up on #325, functionality in the `Repo` class, that is actually not about a Git repository, should be extracted into stand-alone functions. `lib/utils.py` seems like an obvious place to go for. The aim is, to have iterative refactoring. First, separate out what either doesn't need to be part of `Repo` or belongs to a non-repo notion (like "assets" in case of #325) and then see which of those functions are operating on the same kind of data and may form a new class. This should then help to clarify what exact notion/abstraction is actually needed.

TobiasKadelka:
I want to point to this issue: #298
It is also about extracting some functionality, that did not belong into the repository class, but should also probably not live in the commands, so it might be considered together with the current effort :)

PR #342, opened 2023-04-27
RF lib - Step 1

bpoldrack:
Ping #340

### DIFF
--- a/onyo/lib/command_utils.py
+++ b/onyo/lib/command_utils.py
@@ -10,16 +10,17 @@ from ruamel.yaml import YAML, scanner  # pyre-ignore[21]
 from onyo import OnyoInvalidRepoError, OnyoProtectedPathError
 from .onyo import Repo, log
 from .filters import Filter
+from .utils import is_protected_path, read_asset, get_assets_by_path, write_asset
 
 
 def fsck(repo: Repo, tests: Optional[list[str]] = None) -> None:
     all_tests = {
-        "clean-tree": repo._fsck_clean_tree,
-        "anchors": repo._fsck_anchors,
-        "asset-unique": repo._fsck_unique_assets,
-        "asset-yaml": repo._fsck_yaml,
-        "asset-validity": repo._fsck_validation,
-        "pseudo-keys": repo._fsck_pseudo_keys,
+        "clean-tree": fsck_clean_tree,
+        "anchors": fsck_anchors,
+        "asset-unique": fsck_unique_assets,
+        "asset-yaml": fsck_yaml,
+        "asset-validity": fsck_validation,
+        "pseudo-keys": fsck_pseudo_keys,
     }
     if tests:
         # only known tests are accepted
@@ -33,7 +34,7 @@ def fsck(repo: Repo, tests: Optional[list[str]] = None) -> None:
         # TODO: these should be INFO
         log.debug(f"'{key}' starting")
 
-        if not all_tests[key]():
+        if not all_tests[key](repo):
             log.debug(f"'{key}' failed")
             raise OnyoInvalidRepoError(f"'{repo._opdir}' failed fsck test '{key}'")
 
@@ -43,7 +44,7 @@ def fsck(repo: Repo, tests: Optional[list[str]] = None) -> None:
 def fsck_anchors(repo: Repo) -> bool:
     anchors_exist = {x for x in repo.files if x.name == '.anchor' and '.onyo' not in x.parts}
     anchors_expected = {x.joinpath('.anchor') for x in repo.dirs
-                        if not repo._is_protected_path(x)}
+                        if not is_protected_path(x)}
     difference = anchors_expected.difference(anchors_exist)
 
     if difference:
@@ -95,7 +96,7 @@ def fsck_pseudo_keys(repo: Repo) -> bool:
 
     for asset in repo.assets:
         violation_list = [
-            x for x in pseudo_keys if x in repo._read_asset(asset)]
+            x for x in pseudo_keys if x in read_asset(asset)]
         if violation_list:
             assets_failed[asset] = violation_list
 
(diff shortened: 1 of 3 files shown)


onyo/lib/utils.py
```diff
@@ -0,0 +1,76 @@
+from pathlib import Path
+from typing import Dict, Iterable, Union
+
+from ruamel.yaml import YAML, scanner
```
aqw, line 4: I see that this has removed the pyre ignore directive. Have you tested that it finally detects `ruamel` correctly?

onyo/lib/command_utils.py
```diff
@@ -0,0 +1,785 @@
...
+        return False
+
+    return True
+
+
+def set_assets(repo: Repo,
+               paths: Iterable[Union[Path, str]],
+               values: Dict[str, Union[str, int, float]],
+               dryrun: bool,
+               rename: bool, depth: Union[int]) -> str:
```
TobiasKadelka, line 568: Did you possibly overlook to add the linebreak between these two variables, or is there a reason to except the "one line per variable" rule here? :)

onyo/lib/onyo.py
```diff
@@ -357,28 +347,16 @@ def _get_root(self) -> Path:
...
-        return templates
+        from .utils import get_templates
+        return get_templates(self)
 
     #
     # HELPERS
     #
     @staticmethod
     def _is_protected_path(path: Union[Path, str]) -> bool:
-        """
```
TobiasKadelka, line 370: For `Repo._get_assets(self)` in line 280 you left the doc-string even though you did the same change with `from .utils import`

I personally think the docstrings should be in both functions until they get fully removed. They can either be full duplicates or just a shortened/summarized version, but having a function that does not have a doc-string in the repo-class might turn out to be inconvenient?

onyo/lib/onyo.py
```diff
@@ -534,165 +512,49 @@ def fsck(self, tests: Optional[list[str]] = None) -> None:
...
-            # TODO: these should be INFO
-            log.debug(f"'{key}' starting")
-
-            if not all_tests[key]():
-                log.debug(f"'{key}' failed")
-                raise OnyoInvalidRepoError(f"'{self._opdir}' failed fsck test '{key}'")
-
-            log.debug(f"'{key}' succeeded")
+        from .command_utils import fsck as ut_fsck
+        ut_fsck(self, tests)
```
TobiasKadelka, line 516: This is very nit-picky, and I know that others in the group have different opinions about this, but I want to register that I am personally not a fan of overshortened variable and function names. I feel like the function gets not that often called, that it could not be called `utils_fsck()` :)

onyo/lib/utils.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # protected paths
+    for p in full_path.parts:
+        if p in ['.anchor', '.git', '.onyo']:
+            return True
+
+    return False
+
+
+def get_assets(repo: Repo) -> set[Path]:
```
TobiasKadelka, line 24: After seeing this change, I am not fully convinced yet, that `get_assets()` and `get_templates()` should be excluded from the repo. It is not clear to me, that "just git handling" is the line that we should draw between onyo.py and utils.py.
Seeing this moved to here gives me the feeling that I know have to remember if functions do internally do a git-call or are requested or cached in some other way then for example `Repo.get_dirs()`.

I personally would prefer this to be split like the hypothetical asset class, where `Repo` contains all stuff related to handling of repo-information, independently from where and how the information is found.

That said, +1 on the `read_asset()`, `is_protected_path()` and all the `fsck()` related functios in `command_utils.py` :)

onyo/lib/utils.py
bpoldrack: @aqw Actually not intentional - import was auto-generated and haven't run pyre on it yet. So, thx - will try!

TobiasKadelka (changes requested):
Hey, thanks a lot for the PR!
In general, there were quite often discussions about how to split and devide functionality, so seeing somebody having such a concrete plan how it should look like and how it could be done in the least interruptive way is very nice, and I am in favor of most of the changes :)

My major concern after reading trough the PR and thinking about the implications of the changes is the removal of `get_assets()` and `get_templates()` into utils.py, while `Repo.get_dirs()` and other getter functions stay part of the repository.
I feel like these functions belong together and into the repository class, since they are all there to make the handling of information that is connected to the repository easy. So having to know if a function uses internally a git-call to figure out were the function is, feels awkward to me and we might want to discuss this a bit more.

That said, I am very much in favor of moving all the check/test/sanitizing functions outside of the repository! :)

onyo/lib/command_utils.py
bpoldrack: Overlooked, thx!

onyo/lib/onyo.py
bpoldrack: True - docstring should stay. Will fix. 

Not having them copied in the extracted functions for now is intentional. I want to quickly iterate over refactoring them. So, adjusting docstrings all the time would be a major slowdown. And in this state, no direct usage of the extracted functions is intended. `Repo` is to be used as before for now, hence this is what still needs the docstrings.

onyo/lib/onyo.py
TobiasKadelka: Thanks a lot :)

onyo/lib/utils.py
bpoldrack: > Seeing this moved to here gives me the feeling that I know have to remember if functions do internally do a git-call or are requested or cached in some other way then for example Repo.get_dirs().

To add to conversation in chat: No, you don't need to know. In this state, you'd still call `Repo.get_assets`/`Repo.get_dirs` as before. This is all about being able to reshape things w/o interfering in the usage of `Repo` yet.

bpoldrack:
Note, that some coverage decrease in `onyo.py` is to be expected, because functions calling each other used to do that within `Repo` whereas now, that shim remaining in `Repo` isn't actually used. That should be alright, since this PR does not intend to change `Repo`'s interface yet.

TobiasKadelka (approved):

PR #350, opened 2023-05-12
Pathing: retire `repo.opdir` and `repo.sanitize_path()`

TobiasKadelka:
This is the next step for the efforts of simplifying path handling in onyo and commands, while removing bugs and addings tests for those changes.

- change cwd to `opdir` in `main.py`
  - `opdir` is exclusively used in `main.py`  (relevant if `onyo -C <path>` is called)
- retire `repo.opdir` and `opdir` in lib, commands and tests
- remove `repo.sanitize_path()`
  - `Path(<path>).resolve()` is now used in commands, since `repo.sanitize_path()` did nothing else after retirement of `opdir` anyways

### DIFF
--- a/onyo/commands/cat.py
+++ b/onyo/commands/cat.py
@@ -13,7 +13,7 @@ logging.basicConfig()
 log: logging.Logger = logging.getLogger('onyo')
 
 
-def sanitize_paths(paths: list[str], opdir: str) -> list[Path]:
+def sanitize_paths(paths: list[str]) -> list[Path]:
     """
     Check and normalize a list of paths. If paths do not exist or are not files,
     print paths and exit with error.
@@ -23,8 +23,7 @@ def sanitize_paths(paths: list[str], opdir: str) -> list[Path]:
     error_path_not_file = []
 
     for p in paths:
-        # TODO: This is wrong when an absolute path is provided
-        full_path = Path(opdir, p).resolve()
+        full_path = Path(p).resolve()
 
         # path must exist
         if not full_path.exists():
@@ -53,19 +52,19 @@ def sanitize_paths(paths: list[str], opdir: str) -> list[Path]:
     return paths_to_cat
 
 
-def cat(args: argparse.Namespace, opdir: str) -> None:
+def cat(args: argparse.Namespace) -> None:
     """
     Print the contents of ``asset``\\(s) to the terminal without parsing or
     validating the contents.
     """
     repo = None
     try:
-        repo = Repo(opdir, find_root=True)
+        repo = Repo(Path.cwd(), find_root=True)
         repo.fsck(['asset-yaml'])
     except OnyoInvalidRepoError:
         sys.exit(1)
 
-    paths_to_cat = sanitize_paths(args.asset, opdir)
+    paths_to_cat = sanitize_paths(args.asset)
 
     # open file and print to stdout
     for path in paths_to_cat:
--- a/onyo/commands/history.py
+++ b/onyo/commands/history.py
@@ -16,17 +16,17 @@ logging.basicConfig()
 log: logging.Logger = logging.getLogger('onyo')
 
 
-def sanitize_path(path: str, opdir: str) -> Path:
+def sanitize_path(path: str) -> Path:
     """
-    Checks a path relative to opdir. If it does not exist, it will print an
-    error and exit.
+    Checks and resolves a path. If it does not exist, it will print an error and
+    exit.
 
     Returns an absolute path on success.
     """
     if not path:
         path = './'
 
-    full_path = Path(opdir, path).resolve()
+    full_path = Path(path).resolve()
 
     # check if path exists
     if not full_path.exists():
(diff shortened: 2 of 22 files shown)


onyo/commands/history.py
```diff
@@ -78,25 +78,21 @@ def history(args: argparse.Namespace, opdir: str) -> None:
...
     # get the command and path
-    path = sanitize_path(args.path, opdir)
+    path = sanitize_path(args.path, repo)
     history_cmd = get_history_cmd(args.interactive, repo)
 
     # run it
-    orig_cwd = os.getcwd()
     status = 0
     try:
-        os.chdir(opdir)
```
aqw, line 94: Removing this will cause `'history` with `onyo -C` to fail.

`git log`, `tig`, etc require that they are run from within a git repository (or use `-C` ).

TobiasKadelka: It is not removed, it is moved into `main.py`, were we now generally change the cwd to opdir, so that the commands to not have to care about opdir, thereby this does not fail, because these commands are indeed run from insode a git repository.

aqw: Cool. Thanks for explaining. :-)

onyo/commands/cat.py
```diff
@@ -13,7 +13,7 @@
 log: logging.Logger = logging.getLogger('onyo')
 
 
-def sanitize_paths(paths: list[str], opdir: str) -> list[Path]:
+def sanitize_paths(repo: Repo, paths: list[str]) -> list[Path]:
```
bpoldrack, line 16: If I didn't miss something, `repo` isn't used anymore
```suggestion
def sanitize_paths(paths: list[str]) -> list[Path]:
```

bpoldrack: I suppose, this needs double-checking across commands - remove unused parameter and adjust calls accordingly.

tests/lib/test_generate_commit_message.py
```diff
@@ -34,9 +35,8 @@ def test_Repo_generate_commit_message(repo: Repo) -> None:
     to the root of the repository.
     """
     # modify the repository with some different commands:
-    repo.mkdir(repo.sanitize_path("a/new/folder"))
-    repo.mv(repo.sanitize_path("s p a c e s"),
-            repo.sanitize_path("a/new/folder"))
+    repo.mkdir(Path("a/new/folder").resolve())
+    repo.mv(Path("s p a c e s").resolve(), Path("a/new/folder").resolve())
```
bpoldrack, line 39: For the purpose of the test we know the path exactly: `repo.root / 'a' / 'new' / 'folder'`. No need to resolve anything. Resolving also is actually wrong, since this is resolving a relative path against CWD. But CWD is irrelevant here - we don't want to hardcode assumptions about where the tests are running from. All that matters here is that we pass absolute paths to`Repo.mv`.

bpoldrack (commented):
Thx, @TobiasKadelka, looks good in principle. Left some comments. :)

onyo/commands/init.py
```diff
@@ -21,12 +21,12 @@ def init(args: argparse.Namespace, opdir: str) -> None:
...
     overwrite anything; it will exit with an error.
     """
-    target_dir = Path(opdir)
+    target_dir = Path.cwd()
     if args.directory:
         if Path(args.directory).is_absolute():
             target_dir = Path(args.directory)
         else:
-            target_dir = Path(opdir, args.directory)
+            target_dir = Path(args.directory)
```
bpoldrack, line 29: This does not make sense anymore.
1. `target_dir` gets the same assignment in both cases.
2. Now a relative one wouldn't be resolved at all

I'd replace the entire `target_dir` setting by:

```
target_dir = Path(args.directory).resolve() if args.directory else Path.cwd()
```

onyo/lib/onyo.py
```diff
@@ -339,7 +307,7 @@ def _get_dirs(self) -> set[Path]:
         (except under .git).
         """
         log.debug('Acquiring list of directories')
-        dirs = {self.sanitize_path(x) for x in Path(self.root).glob('**/')
+        dirs = {Path(x).resolve() for x in Path(self.root).glob('**/')
```
bpoldrack, line 310: This seems wrong. The paths reported by `Path(self.root).glob('**/')` are relative to root, not to CWD. So, `resolve()` is not what needs to doing here.


```suggestion
        dirs = {self.root / x for x in Path(self.root).glob('**/')
```

onyo/lib/utils.py
```diff
@@ -29,7 +29,7 @@ def get_assets(repo: Repo) -> set[Path]:
 
 
 def get_templates(repo: Repo) -> set[Path]:
-    return {repo.sanitize_path(file)
+    return {Path(file).resolve()
```
bpoldrack, line 32: Same here.

```suggestion
    return {repo.root / file
```

bpoldrack (commented):
A few oddities still, otherwise LGTM. Thx, @TobiasKadelka

bpoldrack (approved):

Issue #206, opened 2022-09-28
BUG: `-C` with a relative path results in widespread breakage


aqw:
Likely related to #170

TobiasKadelka:
I thought long about how to attack this. I want to describe what I see as the only valid solution for this, considering all other parts of onyo, not just functionaly, but also when it comes to printing:
- all `Path`s in the repo class should be **absolut** internally. We need one or multiple functions (described in issue #325) that enable functionality for all commands and the repo class, and that take the `opdir` (either cwd, or provided with `onyo -C`, then normalized/sanitized to an absolut path) and connect that to the paths that might be provided as arguments via CLI or otherwise
- afterwards, all internal functionality works with the paths build like described above
- there needs to be one or multiple functions, used whenever printing happens to make paths relative from repo root
- the behavior here described will automatically be covered by tests already existing for some cases, but we definitely need new/additional tests to cover the varieties of different path-combinations (-C with absolut or not, paths given absolute or not, etc etc etc).

This is not just true for the paths of assets, but also for all other properties of `Repo`, I think including `Repo.root` and `Repo.opdir`

TobiasKadelka:
We changed the way onyo handles paths dramatically, the issues related to pathing should be solved.
Thereby I will close this issue; if something ever comes up again, we will file a proper bug-report.

Issue #142, opened 2022-08-31
TST: verify that onyo works from different locations with relative and absolute paths


TobiasKadelka:
There are many combinations of paths that could be passed into onyo, with and without `onyo -C`, with and without absolute paths. That these work as intended needs to be implemented and tested, for API and CLI!

These are the different CLI situations Alex came up with that need to be tested for, once it is verfied that they work and are tested, they can be marked as checked:
- [x] repo_root arg_relative
- [ ] repo_root arg_absolute
- [ ] repo_subdir arg_relative
- [ ] repo_subdir arg_absolute
- [ ] repo_root -C root_relative arg_relative
- [ ] repo_root -C root_relative arg_absolute
- [ ] repo_root -C root_absolute arg_relative
- [ ] repo_root -C root_absolute arg_absolute
- [ ] repo_root -C subdir_relative arg_relative
- [ ] repo_root -C subdir_relative arg_absolute
- [ ] repo_root -C subdir_absolute arg_relative
- [ ] repo_root -C subdir_absolute arg_absolute
- [ ] repo_subdir -C root_relative arg_relative
- [ ] repo_subdir -C root_relative arg_absolute
- [ ] repo_subdir -C root_absolute arg_relative
- [ ] repo_subdir -C root_absolute arg_absolute
- [ ] repo_subdir -C subdir_relative arg_relative
- [ ] repo_subdir -C subdir_relative arg_absolute
- [ ] repo_subdir -C subdir_absolute arg_relative
- [ ] repo_subdir -C subdir_absolute arg_absolute
- [ ] outside -C root_relative arg_relative
- [ ] outside -C root_relative arg_absolute
- [ ] outside -C root_absolute arg_relative
- [ ] outside -C root_absolute arg_absolute
- [ ] outside -C subdir_relative arg_relative
- [ ] outside -C subdir_relative arg_absolute
- [ ] outside -C subdir_absolute arg_relative
- [ ] outside -C subdir_absolute arg_absolute

aqw:
In general, I think we should follow Git's lead here. We can't do anything they can't do, and it largely makes sense to match them unless we have a good reason otherwise to deviate.

1) `onyo new <absolute path>/shelf` (so everywhere on the system with absolute path?)

This should work /if/ the CWD is in the same Onyo repo. Otherwise it should fail. This is the same as Git's behavior. This is why `-C` exists.

2) `onyo new <relative path>/shelf`

[...]

PR #351, opened 2023-05-12
TST: add tests for rel/abs paths and from different locations

TobiasKadelka:
Changes:
- Adds tests for different path situations for `onyo new`:
  - test from a subdirectory of the repository with _absolute_ paths
  - test from a subdirectory of the repository with _relative_ paths

TODOs (already described in the issues listed below):
- add tests for different paths to other CLI tests (like done for `onyo new`)
- add tests with `onyo -C` to CLI tests
  
This works into the direction of #142 and #206

### DIFF
--- a/onyo/commands/tests/test_new.py
+++ b/onyo/commands/tests/test_new.py
@@ -26,16 +26,16 @@ def test_new(repo: OnyoRepo, directory: str) -> None:
     """
     Test that `onyo new` can create an asset in different directories.
     """
-    file = f'{directory}/laptop_apple_macbookpro.0'
-    ret = subprocess.run(['onyo', 'new', '--yes', '--path', file],
+    file_ = f'{directory}/laptop_apple_macbookpro.0'
+    ret = subprocess.run(['onyo', 'new', '--yes', '--path', file_],
                          capture_output=True, text=True)
 
     # verify correct output
     assert "The following will be created:" in ret.stdout
-    assert file in ret.stdout
+    assert file_ in ret.stdout
     assert not ret.stderr
     assert ret.returncode == 0
-    assert Path(file).exists()
+    assert Path(file_).exists()
 
     # verify that the new asset exists and the repository is in a clean state
     assert len(repo.asset_paths) == 1
@@ -48,17 +48,17 @@ def test_new_interactive(repo: OnyoRepo) -> None:
     Test that `onyo new` can create an asset in different directories when given
     'y' as input in the dialog, instead of using the flag '--yes'.
     """
-    file = f'{directories[0]}/laptop_apple_macbookpro.0'
-    ret = subprocess.run(['onyo', 'new', '--path', file],
+    file_ = f'{directories[0]}/laptop_apple_macbookpro.0'
+    ret = subprocess.run(['onyo', 'new', '--path', file_],
                          input='y', capture_output=True, text=True)
 
     # verify correct output
     assert "The following will be created:" in ret.stdout
     assert "Create assets? (y/n) " in ret.stdout
-    assert file in ret.stdout
+    assert file_ in ret.stdout
     assert not ret.stderr
     assert ret.returncode == 0
-    assert Path(file).exists()
+    assert Path(file_).exists()
 
     # verify that the new asset exists and the repository is in a clean state
     assert len(repo.asset_paths) == 1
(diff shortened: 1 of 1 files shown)


TobiasKadelka:
This PR was a bit older already. I stopped working on it, so as to give Ben the chance to finish his refactoring efforts. Now I took the time to rebase on top of all the changes merged over the last few weeks, and I added (and modernized) the test functions that I added originally with this PR.

The idea of those tests is to make sure that an onyo command (`new` in this case) can handle all the different cases of input paths from different locations. I would be happy if somebody would review the functions I added so far, so that we can discuss based on these how we want to actually test this for all the different commands. After review/discussion I will add functions similar to this for other commands (and other path-situations).

onyo/commands/tests/test_new.py
```diff
@@ -86,6 +86,98 @@ def test_new_top_level(repo: OnyoRepo) -> None:
...
+    for d in directories:
+        # build an absolute path for a new asset
+        new_file = Path(f"{d}/{file}").resolve()
+
+        ret = subprocess.run(['onyo', 'new', '--yes', '--path', new_file],
+                             capture_output=True, text=True)
+
+        # verify correct output
+        assert "The following will be created:" in ret.stdout
+        assert str(d) in ret.stdout
```
bpoldrack, line 106:
```suggestion
```

You assert that `newfile` is showing up already. `new_file` is an absolute path containing `d`. So, this assertion here is superfluous.

onyo/commands/tests/test_new.py
```diff
@@ -86,6 +86,98 @@ def test_new_top_level(repo: OnyoRepo) -> None:
...
+        for d in directories:
+            # build an absolute path for a new asset
+            new_file = f"{repo.git.root}/{d}/{file}"
+
+            ret = subprocess.run(['onyo', 'new', '--yes', '--path', new_file],
+                                 capture_output=True, text=True)
+
+            # verify correct output
+            assert "The following will be created:" in ret.stdout
+            assert str(d) in ret.stdout
```
bpoldrack, line 137: Same as above.

onyo/commands/tests/test_new.py
```diff
@@ -86,6 +86,98 @@ def test_new_top_level(repo: OnyoRepo) -> None:
     fsck(repo)
 
 
+@pytest.mark.repo_dirs(*directories)
+def test_new_from_root_with_absolute_paths(repo: OnyoRepo) -> None:
```
bpoldrack, line 90: To be honest, I don't think we need this test. The very next test does essentially the same thing. Why would CWD being the root of the worktree be a special case when acting upon an absolute path? The command is supposed to act upon the given absolute path only - CWD should not somehow end up influencing the interpretation of that path. These things are supposed to be covered by `test_new_sub_dir_absolute_path`.

onyo/commands/tests/test_new.py
```diff
@@ -86,6 +86,98 @@ def test_new_top_level(repo: OnyoRepo) -> None:
...
+
+@pytest.mark.repo_dirs(*directories)
+def test_new_sub_dir_absolute_path(repo: OnyoRepo) -> None:
+    """
+    Test that `onyo new --path <path>` can create an asset in sub-directories of
+    the repository while the cwd is inside a different sub-directory.
+    """
+    file = 'laptop_apple_macbookpro.faux'
+
+    for cwd in directories:
```
bpoldrack, line 124: This is expensive. Why test every single combination of directories?

Take one of the `directories` for CWD and another one for the asset. That's what you want to test (according to the docstring). What does repeating the test with different directories add, that wasn't already covered?

onyo/commands/tests/test_new.py
```diff
@@ -86,6 +86,98 @@ def test_new_top_level(repo: OnyoRepo) -> None:
...
+@pytest.mark.repo_dirs(*directories)
+def test_new_sub_dir_relative_path(repo: OnyoRepo) -> None:
+    """
+    Test that `onyo new --path <path>` can create an asset in sub-directories of
+    the repository while the cwd is inside a different sub-directory with a
+    relative path.
+    """
+    file = 'laptop_apple_macbookpro.faux'
+
+    for cwd in directories:
```
bpoldrack, line 156: Same as above: Nested loop for all combinations of directories seems unnecessarily expensive.

onyo/commands/tests/test_new.py
TobiasKadelka: I am not sure, that this is actually superfluous, though it does rather test the output. And I think, especially with the refactoring of output that we plan to centralize printing in onyo, that to have this is not necessarily a bad thing for later.

onyo/commands/tests/test_new.py
TobiasKadelka: Hm, I feel like you assume the way the paths are handled currently, and that it is done correctly everywhere (which I assume technically to be right).

But this test would IMO catch something like us at some place building paths like `Path(cwd, root_as_rel)`, that would let onyo look like it works always, but still would error when root would be an absolute path that gets entered wrong in some moment. So to test the correct passing around and handling/formatting of the path when absolute (instead of relative like normally) is here my intention.

onyo/commands/tests/test_new.py
TobiasKadelka: Woops, just saw the code, I literally check `assert str(new_file) in ret.stdout` so I do what I want to do there indeed already, you were right and I removed those lines! :)

onyo/commands/tests/test_new.py
bpoldrack: In any case: What makes `Path(cwd, root_as_rel)` special? Why does this deserve a dedicated test, that does the same thing as the next test for other directories in the worktree?

onyo/commands/tests/test_new.py
bpoldrack: Hm. While this is a little better, that's not what I meant. You have one dir for CWD and and asset in another dir. Fine. That's what you want to test. What does it add to also run `new` for other assets in even more dirs? What does that cover that is not covered by chdir into `directories[0]` and run new on an asset in `directories[1]`? How is the rest of the directories different? Dealing with all sorts of paths (several levels, unicode, etc.) has nothing to do with the stated purpose of this test, but with general path handling. That's tested elsewhere.

onyo/commands/tests/test_new.py
```diff
@@ -86,6 +86,64 @@ def test_new_top_level(repo: OnyoRepo) -> None:
...
+    """
+    # Change cwd to sub-directory
+    cwd = repo.git.root / 'r/e/c/u/r/s/i/v/e'
+    os.chdir(cwd)
+
+    # count levels until repo.git.root and build relative path with "../"
+    var = "../" * (len(cwd.relative_to(repo.git.root).parts) - 1)
+
+    # build a path for the new asset as a relative path
+    file = f"{var}just another path/laptop_apple_macbookpro.0"
```
bpoldrack, line 131: This building of a path seems a little weird. We use `pathlib`. Why sort of work around it, by manually constructing?

How about this instead?

```suggestion
    # build a path for the new asset as a relative path
    file = (repo.git.root / "just another path" / "laptop_apple_macbookpro.0").relative_to(cwd)
```

onyo/commands/tests/test_new.py
```diff
@@ -86,6 +86,64 @@ def test_new_top_level(repo: OnyoRepo) -> None:
...
+    """
+    Test that `onyo new --path <path>` can create an asset in sub-directories of
+    the repository while the cwd is inside a different sub-directory.
+    """
+    # Change cwd to sub-directory
+    cwd = repo.git.root / 'spe\"cial\\char\'acteஞrs'
+    os.chdir(cwd)
+
+    # build a path for the new asset as an absolute path
+    file = repo.git.root / "just another path/laptop_apple_macbookpro.0"
```
bpoldrack, line 100: I'd say we either use the `/` operator or not. But using it at some arbitrary spots only, seems to not serve a purpose.

```suggestion
    file = repo.git.root / "just another path" / "laptop_apple_macbookpro.0"
```

onyo/commands/tests/test_new.py
TobiasKadelka: I decided to do this way with strings, since this is a CLI test, and the CLI is called with a string.

Your example obviously makes sense too, because it is then converted into a string, during the subprocess call. But the code you share above will error out, since `relative_to(cwd)` wants one path to be a sub-path of the other (which it is not, therefor the error), and does not build a relative path with "../" in it.

TobiasKadelka: Sorry, I just posted the negative and forgot to post something productive, too!

I have now changed it to:
`file = Path(var, "just another path", "laptop_apple_macbookpro.0")`

This uses path, and keeps it relative with the `../../`. I assume you might dislike this still somewhat (because of the counting of levels in `var`), but I choose this approach because I think this works fairly well to test what the function should test.

If this is not satisfying, I am happy to discuss how this should actually look like :)

onyo/commands/tests/test_new.py
TobiasKadelka: Here I fully agree, thanks, I updated it! :)

onyo/commands/tests/test_new.py
```diff
@@ -86,6 +86,64 @@ def test_new_top_level(repo: OnyoRepo) -> None:
...
+    """
+    Test that `onyo new --path <path>` can create an asset in sub-directories of
+    the repository while the cwd is inside a different sub-directory.
+    """
+    # Change cwd to sub-directory
+    cwd = repo.git.root / 'spe\"cial\\char\'acteஞrs'
+    os.chdir(cwd)
+
+    # build a path for the new asset as an absolute path
+    file = repo.git.root / "just another path" / "laptop_apple_macbookpro.0"
```
bpoldrack, line 100: I'd rename `file` for `file_` or `path`. Generally not a fan of shadowing builtins.

bpoldrack (commented):
Just a minor renaming suggestion. Otherwise, fine. Thanks.

onyo/commands/tests/test_new.py
TobiasKadelka: Thanks a lot, I changed that for all tests of `test_new.py`!

Issue #113, opened 2022-08-09
 Add `onyo set --filter` flag 


TobiasKadelka:
This was not implemented with `onyo set`.

From the old readme (needs to be added with the implementation again):

`onyo set [--filter key=value[,key=value...], -f]`

  `--filter key=value`: Only apply changes to assets that have `key` and it
    matches `value`. Multiple restrictions can be declared by joining with a `,`
    --- which acts as a logical AND operator. There is no "OR" equivalent.

    `value` can be a string or a Python regular expression. Onyo's string
    representation of types can also be used for queries, such as "[list]",
    "[dict]", or "[unset]" (for keys that don't exist).

Issue #259, opened 2022-12-12
Add functions to get filtered (e.g. subset of) dirs/assets


TobiasKadelka:
(To minimize text, I will just describe assets, but the same is valid for other attributes like the `directories`)

Currently, the repo-class has an attribute to access `Repo.assets`, while the function `Repo._get_assets()` gets just called if the `assets` are `None`. For many places, e.g. `onyo set --filter` and `onyo get`, it is necessary to get a subset of assets (for example just from a certain directory). There are some approches how to filter a list of assets for certain names or directories.

1. Instead of using the properties, the assets should mainly be accessed via functions, the properties are just used as caches by the functions. The functions could then be called either as `Repo.assets()` (to access achive the current behavior) or with an additional parameter as `Repo.assets(directory=<subdir>)` (to get a filtered set). This would create a need to update existing code, but is in my opinion the most intuitive way to use the Repo class. To get different sets of the repos assets, the user would always use the same function call, simply with different parameters.

2. Alternatively, the current behaviour could be kept, but just additional functions can be added, that enable the filtering, e.g. `repo.filter_assets(<subdirectory>)`.

These ways can be extended with additional optional parameters as much as needed, e.g. with a parameter that filteres for names, or certain fields that must be set or even have a specific value. So the question is really how we want to access assets, directories and other attributes of the repository when using the class.

TobiasKadelka:
From a comment of the PR #243:
>>[...] as a part of this, determining how this matches with _is_protected_path() and fixing assets to filter using valid_name().

Independently from the method we solve the above filtering, this should be considered and worked into it.

TobiasKadelka:
From a comment in #257:

> The only consumer of this [`_select_assets_from_directory()`] is _set_sanitize(), yet there is a significant amount of overlap of functionality and checks between the two functions.

These to functions will be reduced when the above mentioned changes are implemented and might even be merged into one function, in case later commands (like `onyo get`) do not need the functionality separately.

PR #353, opened 2023-05-15
set/unset: add flag `--filter`

TobiasKadelka:
This PR just copies the stuff from `onyo get` to the commands unset and set, too.

Close https://github.com/psyinfra/onyo/issues/113
Close https://github.com/psyinfra/onyo/issues/285

TODO: Tests are still missing.

The function used by onyo get to filter seems to me to be quite badly written, which leads to a few awkward things I had to do. IMO the function should be split up into two parts, the first doing what https://github.com/psyinfra/onyo/issues/259 describes, and a separate one should do the filtering by keys, which is just needed by onyo get.


bpoldrack (approved):
Thanks, @TobiasKadelka.

Issue #345, opened 2023-05-08
Define usecases to test for


bpoldrack:
Based on the usecase discussion in #335, here I will try to boil it down to actually testable cases.
Opening post is intended to be edited to flesh it out. As verbal description first. Meant to be converted into actual operations (and a TODO for each case).

Usecases
========

New group/location
-----------------------------

- new inventory folder: `onyo mkdir <location>`

New member
--------------------

- add a folder: `onyo mkdir <group/user>`
- assign asset(s)
    - find something in warehouse 
    - move from warehouse  `onyo mv <asset> <group/user>`
    - add directly to user? (new purchase)  `onyo new <group/user/asset>`

Member moves
-----------------------

- first, probably query what member actually has
- assets may need to move to
  - group or room (stuff left behind)
  - move with the member
  - move to warehouse
  - move from warehouse

Member leaves
-----------------------

- What does member have assigned?
- assets may be found "lost" on this occasion?
- assets go back to warehouse
- assets may be retired
- remove member

Member/group/room changes name/"identifier"
-----------------------------------------------------------------------

- rename respective folder

Add FZJ inventory number to existing assets
----------------------------------------------------------------

- find asset in question (possibly needs interaction with member)
- add key value pair

Asset changes properties due to upgrade/repair/whatever
------------------------------------------------------------------------------------

- change value of a key
- add a key-value pair

Retire asset
-----------------

- What needs to be found?
- Move asset to "retired"

Remove an asset
-------------------------

- actual removal

New asset(s) purchased
----------------------------------

- add to warehouse

Queries
-----------

- What is available in warehouse/on shelf?
- Everything with an FZJ inventory number
- Where is an asset? Based on:
  - Hostname of a laptop
  - pseudo key(s)
  - inventory number
- All assets within a subtree (group, room, person, whatever), that also match some filter criteria
- What is the history of an asset?
- All assets not in "retired" location of a certain type, that match something like `"20160101 <= build-date <= 20171231"`
- Apple recall on battery of all 13" MacBook Pros built between 2016-2017

Validation
---------------

- unique asset names
  General note: This should probably be based on resolved paths in order to allow symlinks.
  For example: part time association with different groups and asset search point may differ based on context
- valid yaml
- clean worktree
- Configurable constraints on actual metadata?
  - criteria for keys within assets?
  - criteria for keys across assets?
  - limit such criteria to subtrees?
  
 Configuration
---------------------

That's entirely uncovered by discussion so far.

[...]

Issue #335, opened 2023-04-25
Define usecases and workflows to support


bpoldrack:
This issue is supposed to track discussion of what we think are usecases and workflows that are relevant for onyo.
From this, test cases defining them should then be extracted and implemented.

Primarily, I'd want to know from @aghisla, @j-leo and @aqw: 
- What are real-world scenarios that would make you want to interact with the inventory system?
- What information do you start with?
- What information needs to be discovered?
- What information needs to be put into it at what point in time?
- What would be typical/expected cases of some asset maybe already existing in the system, but additional information needs to be added to it or changed?

Some (partially) made-up example: My move to D&P.
I had a laptop, keyboard+mouse and a display and was assigned to a room. I took laptop, keyboard+mouse, moved to a different office and left the display in the old one.
Now, I imagine something like this: Maybe all those assets were associated with me before.
You'd now start with my name, name of group I left and name of group I joined.
So, you'd probably need to:
- move my association with groups first
- possibly after some communication move room association
- assets would need to split: keep display with former room and/or group, the rest stays (and moves) with me

^ This would need further ironing. When would such a split happen? Prior or after move of person? May be simultaneously? Is such a split expected to be a real-world usecase? (It wouldn't be necessary if the display was associated with the room or group to begin with. But then again: The same may be true for the keyboard. So, may be it's actually two moves: 1. Me (inlc. Laptop) and 2. keyboard (from group to group).

When we figure out these things, we should get to a couple of intended workflows, that can be tested and serve to set priorities for 1.0

j-leo:
OK, so, your move:
- your old screen stayed in your old office ( no change to that, since screens are bound to rooms [exception: home office screens are bound to users]))
- your new screen was brought from the warehouse to your office, so its room number changed
- your laptop is still your laptop - no change to that
- keyboards, mice, network stands will keep to not be monitored

aghisla:
Here is a list of common inventory-related actions that we want to track:

Additions to the inventory:

- add a new object to the inventory and the metadata available at the instant of delivery (metadata depending on the object, but the addition to inventory must happen at delivery, not later)
- add a new member of INM-7 (for any administrative value of "member", as long as they receive something that we inventorise)
- (to be discussed, until we get a working plan): migrate current inventory status from GLPI. Not a requirement that it will be automated (data in there are of very various levels of reliability and as we have to cherrypick data it becomes impossible to script)

Active inventory operations:

- reserve objects: they are in the ware-house but already booked for someone
[...]

PR #357, opened 2023-05-31
Implement usecase tests

bpoldrack:
Addressing the first part of #345, this patch implements a CLI workflow outlined in issue #345.
There are not a lot of assertions in this test yet. There's little use in repeating assertions from the command tests. This is supposed to:
  
  - test that desired operations are actually expressible with onyo's interface (not the currently the case for querying assets for a value within a given range and for limiting the matches to exclude certain subtrees)
  - test that an entire flow doesn't encounter barriers that are unaccounted for in isolated tests for single commands
  - test that and how information retrieved by one command can be used in the next
  - inform further refinement of interfaces, features and outcome definitions
    
Additional assertions should come from decisions made (documented in issues), that more precisely define what is the desired behavior. This is particularly true for how the history should actually look like in order to facilitate an easy assessment of what exactly happened.

As a result of writing this test, I made the issues #358, #359, #360, #361. Issue #345 and this test should be further refined as those issues get addressed.

WRT #345 this addressing the first half (CLI). Identical test for python interface will be required but I don't plan to address it in this PR.

### DIFF
--- a/onyo/tests/test_usecases.py
+++ b/onyo/tests/test_usecases.py
@@ -1,0 +1,205 @@
+import subprocess
+import pytest
+from pathlib import Path
+
+from onyo.lib.onyo import OnyoRepo
+
+
+prefilled_assets = [["warehouse/monitor_dell_PH123.86JZho",
+                     "display: 22.0\nfzj_inventory: '45FZ18'\n"],
+                    ["warehouse/monitor_dell_NoIdea.S0M3",
+                     "display: 27.0\n"],
+                    ["warehouse/laptop_apple_macbook.oiw629",
+                     "RAM: '8GB'\ndisplay: 13.3\nUSB_A: 2\nfzj_inventory: '28FZ34'\nbuild-date: '20160501'\n"],
+                    ["warehouse/laptop_apple_macbook.9r32he",
+                     "RAM: '8GB'\ndisplay: 13.3\nfzj_inventory: '28FJ34'\nbuild-date: '20180501'\n"],
+                    ["somegroup/userA/laptop_apple_macbook.9r5qlk",
+                     "RAM: '8GB'\ndisplay: 15.0\nhostname: 'first.host'\n"],
+                    ["somegroup/userB/laptop_lenovo_thinkpad.owh8e2",
+                     "RAM: '8GB'\ndisplay: 14.6\nfzj_inventory: '13BH9F'\n"],
+                    ["warehouse/laptop_lenovo_thinkpad.iu7h6d",
+                     "RAM: '8GB'\ndisplay: 13.3\nfzj_inventory: '63AH90'\n"],
+                    ["warehouse/laptop_microsoft_surface.oq782j",
+                     "RAM: '8GB'\ndisplay: 12.4\nfzj_inventory: '73CDA45'\ntouchscreen: 'yes'\n"],
+                    ]
+preset_dirs = ['retired', 'lost']
+
+
+@pytest.mark.repo_dirs(*preset_dirs)
+@pytest.mark.repo_contents(*prefilled_assets)
+def test_workflow_cli(repo: OnyoRepo) -> None:
+
+    # 1. create new group
+    workgroup = Path("newgroup")
+    cmd = ['onyo', 'mkdir', '--yes', '--quiet', str(workgroup)]
+    ret = subprocess.run(cmd, capture_output=True, text=True)
+    assert ret.returncode == 0
+
+    # 2. New member joining group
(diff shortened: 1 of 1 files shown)


onyo/tests/test_usecases.py
```diff
@@ -0,0 +1,205 @@
+import subprocess
+import pytest
+from pathlib import Path
+
+from onyo.lib.onyo import OnyoRepo
+
+
+prefilled_assets = [["warehouse/monitor_dell_PH123.86JZho",
+                     "display: '22.0'\nfzj_inventory: '45FZ18'\n"],
```
TobiasKadelka, line 9: While this is not important for the test, I still want to mention that the fzj inventory numbers (as far as I know) are meant to be represented by the serial numbers in onyo.

onyo/tests/test_usecases.py
```diff
@@ -0,0 +1,205 @@
...
+                    ["warehouse/monitor_dell_NoIdea.S0M3",
+                     "display: '27.0'\n"],
```
TobiasKadelka, line 11: Just out of curiousity, is there a specific reason, why you end all strings defining the asset contents with a new line, or is it just a style choice? :D

onyo/tests/test_usecases.py
```diff
@@ -0,0 +1,205 @@
...
+                    ["warehouse/laptop_apple_macbook.oiw629",
+                     "RAM: '8GB'\ndisplay: '13.3'\nUSB_A: '2'\nfzj_inventory: '28FZ34'\nbuild-date: '20160501'\n"],
```
TobiasKadelka, line 13: I notice that you quote all numbers (e.g. `display: '13.3'` and `USB_A: '2'`) even though YAML does know integers and floats.

The idea is around that we want to be able to sum up numbers in queries (obviously more relevant for storage then display sizes, but still). So I think these assets would come closer to real use cases when they would contain actual numbers :)

onyo/tests/test_usecases.py
```diff
@@ -0,0 +1,205 @@
...
+    # 2. New member joining group
+    member = workgroup / "Sam User"
+
+    # 2a. Create the user
+    cmd = ['onyo', 'mkdir', '--yes', '--quiet', str(member)]
+    ret = subprocess.run(cmd, capture_output=True, text=True)
+    assert ret.returncode == 0
+
+    # 2b. Check the warehouse for a display
+    cmd = ['onyo', 'get', '-p', 'warehouse', '-H', '--filter', 'type=monitor', 'display=22.0']
```
TobiasKadelka, line 47: As a follow up, I notice that here in `'display=22.0'` the number is not quoted separately, so entered as an actual number, which is different from above. :)

TobiasKadelka (commented):
LGTM, thanks for the PR!

Most points I brought up are not meant to be problems, but rather my ideas and thoughts about how to bring this closer to real use cases :)

onyo/tests/test_usecases.py
bpoldrack: I don't think, that's a good idea. Discussed this with @aqw in context of #335 and he agreed to not be able to justify it anymore ;-)

onyo/tests/test_usecases.py
bpoldrack: A choice for easy enhancement. If it's always `key: value\n` I don't need to think when appending something ;-)

onyo/tests/test_usecases.py
bpoldrack: Yes, I just took that from what I saw in other tests and the demo repo. I believe this is not yet consistently handled.

> The idea is around

Is there an issue for this idea?

onyo/tests/test_usecases.py
bpoldrack: I was going to say: "Not really. A command line argument is always coming in as a string."
However, you appear to be right, in that `main:StoreKeyValuePairs` does indeed seem to convert it.
Will fix it.

bpoldrack: Switched to floats for display size everywhere (I hope) and pushed.

onyo/tests/test_usecases.py
TobiasKadelka: Nope, but it is probably a requirement for asset folders. I will create one!

TobiasKadelka:
Thanks, I will merge :)

onyo/tests/test_usecases.py
TobiasKadelka: https://github.com/psyinfra/onyo/issues/363

PR #369, opened 2023-06-01
RF: Move direct git calls

bpoldrack:
Further consolidation of separating abstraction layers by moving uses of the git-call wrapper `GitRepo._git` into dedicated `GitRepo` methods.

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -141,7 +141,7 @@ def edit(repo: OnyoRepo,
             repo.git.add(asset)
         else:
             # If user wants to discard changes, restore the asset's state
-            repo.git._git(['restore', str(asset)])
+            repo.git.restore(asset)
             if not quiet:
                 print(f"'{asset}' not updated.")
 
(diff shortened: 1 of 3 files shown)


onyo/lib/commands.py
```diff
@@ -154,7 +154,7 @@ def edit(repo: OnyoRepo,
             repo.git.commit(repo.generate_commit_message(message=message,
                                                          cmd="edit"))
         else:
-            repo.git.restore()
+            repo.git.restore_staged()
```
TobiasKadelka, line 157: I do remember the idea to change Onyos behavior, so that files get not staged but directly commited, especially because `git diff` can be used before, too.
Here you change a bunch of calls to `git.restore_staged()`. Wouldn't all of these need to be then changed again?

bpoldrack: Yes. But: One change at a time. This can be resolved when all the widespread staging is consolidated (see #370). Not what this PR is about.

PR #372, opened 2023-06-01
Fix for mkdir + minor clean up

### LINKED ISSUE #371: Why fail on existing directories with mkdir?
The respective piece of code is currently in `lib/onyo.py:OnyoRepo.mk_inventory_dirs`.

This is failing when it is called with directory paths that already exist. However, `Path.mkdir` is called `parents=True, exist_ok=True`. Furthermore, already existing anchor files apparently are not a reason to fail.

I wouldn't know why we'd want to fail on this condition. If (some) of the dirs exist - fine. Nothing to do.
Any objections to remove that condition?

bpoldrack:
- Closes #371 (`onyo mkdir` now behaves like `mkdir -p` with respect to existing directories)
- Replace remaining, hardcoded `.anchor` strings by the central definition `OnyoRepo.ANCHOR_FILE`.

### DIFF
--- a/onyo/commands/tests/test_mkdir.py
+++ b/onyo/commands/tests/test_mkdir.py
@@ -132,17 +132,17 @@ def test_mkdir_quiet_flag(repo: OnyoRepo) -> None:
 
 @pytest.mark.repo_dirs(*directories)
 @pytest.mark.parametrize('directory', directories)
-def test_error_dir_exists(repo: OnyoRepo, directory: str) -> None:
+def test_dir_exists(repo: OnyoRepo, directory: str) -> None:
     """
-    Test the correct error behavior when `onyo mkdir <path>` is called on an
+    Test the correct behavior when `onyo mkdir <path>` is called on an
     existing directory name.
     """
     ret = subprocess.run(['onyo', 'mkdir', directory], capture_output=True, text=True)
 
     # verify output
-    assert not ret.stdout
-    assert "The following paths already exist:" in ret.stderr
-    assert ret.returncode == 1
+    assert "No assets updated." in ret.stdout
+    assert not ret.stderr
+    assert ret.returncode == 0
 
     assert Path(directory).is_dir()
     assert Path(directory, ".anchor").is_file()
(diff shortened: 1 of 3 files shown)


onyo/commands/tests/test_mkdir.py
```diff
@@ -132,17 +132,17 @@ def test_mkdir_quiet_flag(repo: OnyoRepo) -> None:
 
 @pytest.mark.repo_dirs(*directories)
 @pytest.mark.parametrize('directory', directories)
-def test_error_dir_exists(repo: OnyoRepo, directory: str) -> None:
+def test_dir_exists(repo: OnyoRepo, directory: str) -> None:
     """
     Test the correct error behavior when `onyo mkdir <path>` is called on an
```
TobiasKadelka, line 137: Nit picky, but is it still "error behavior" or rather just behavior? :)

onyo/commands/tests/test_mkdir.py
```diff
@@ -132,17 +132,17 @@ def test_mkdir_quiet_flag(repo: OnyoRepo) -> None:
...
     existing directory name.
     """
     ret = subprocess.run(['onyo', 'mkdir', directory], capture_output=True, text=True)
 
     # verify output
-    assert not ret.stdout
-    assert "The following paths already exist:" in ret.stderr
-    assert ret.returncode == 1
+    assert "No assets updated." in ret.stdout
```
TobiasKadelka, line 143: "assets" seems to be the wrong term IMO. Either directories, or even `.anchor` files are reported with `onyo mkdir`. Especially once asset dirs exist this would became missleading.

onyo/commands/tests/test_mkdir.py
bpoldrack:
```suggestion
    Test the correct behavior when `onyo mkdir <path>` is called on an
```

onyo/commands/tests/test_mkdir.py
bpoldrack: Yes. That's a separate issue, though: #374

bpoldrack:
Thx, @TobiasKadelka. Will squash the test doc and push.

TobiasKadelka:
Thanks, LGTM!

PR #373, opened 2023-06-01
Remove superfluous CWD specification for git calls

### LINKED ISSUE #348: Use `subprocess.run` with `cwd=self.root` for git calls
Currently we often do stuff like this in the repo.py class:
`self._git(['-C', str(self.root), 'ls-files', ...])`, where `self._git()` does the git call with subprocess.

Once the other pathing issues and refactoring efforts are worked on or resolved, we should change this to use the argument `cwd=self.root` for subprocess-calls to unify all behavior, and possibly expose other bugs.

bpoldrack:
`GitRepo._git` uses `self.root` by default now. Furthermore, it makes little sense to call git from elsewhere in a method that is bound to a concrete repository. Hence, the only currently valid use of the `cwd` argument is the git-init call.

Remove all other usages of `cwd` as well as `-C ...` in the command call.

Closes #348

### DIFF
--- a/onyo/lib/git.py
+++ b/onyo/lib/git.py
@@ -98,7 +98,7 @@ class GitRepo(object):
         (except under .git).
         """
         log.debug('Acquiring list of files')
-        files = {self.root / x for x in self._git(['-C', str(self.root), 'ls-files', '-z']).split('\0') if x}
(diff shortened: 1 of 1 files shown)


aqw (approved):
LGTM!

PR #378, opened 2023-06-05
docs: correct/update existing information

TobiasKadelka:
This commit is not about adding new examples or use-cases, I just correct existing information which was out of date:

- `onyo new` does not read fields interactive/separately, but expects a full asset-path
- apply new syntax (e.g. new syntax for flags `--keys` and `--path`) to the commands `onyo new` and `onyo set`
- we new expect assets that expect the name-scheme to be correct, so I renamed the incorrect `laptop_lenovo_T490s` (without serial) to the correct `laptop_lenovo_T490s.abc123`
- the example template is changed to `.onyo/templates/empty` and actually an empty file, instead of the old name `standard`, which did contain `---`

### DIFF
--- a/docs/source/concepts.rst
+++ b/docs/source/concepts.rst
@@ -95,10 +95,10 @@ repository.
     - ``onyo config history.interactive "tig --follow"``
     - ``onyo config history.non-interactive "git --no-pager log --follow"``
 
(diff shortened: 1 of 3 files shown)


bpoldrack (approved):
Thank you, @TobiasKadelka!

PR #385, opened 2023-06-06
Docs: move examples, explain the demo repository

TobiasKadelka:
The full documentation based on this PR: https://onyo--385.org.readthedocs.build/en/385/
The added section for examples of `onyo new` (which is the most complete one from all command-docs I added in this PR) is under https://onyo--385.org.readthedocs.build/en/385/cmd_new.html#example-usage

While looking into these, keep in mind they are not meant to be complete (or beautiful) yet, just provide a basis for discussion. I just want to show the way I think different flags and use-cases should in general be demonstrated for each command in the docs.

--------

CHANGES: 

The examples from the examples.rst are all very minimal, and those should be available at the doc-pages for each command, instead the examples.rst should provide information about more complex usage.

This PR moves the examples for the commands to the page of the respective command itself, furthermore I added some different ways how to use the command, e.g. for `onyo new` with a tsv table, via the CLI for more complex input, and with templates.

The demo repository is now also mentioned in the example page and should there in the future be explained (and the commands run for it should be shown).


docs/source/cmd_edit.rst
```diff
@@ -7,3 +7,13 @@ onyo edit
    :prog: onyo
    :path: edit
 
+
+Example Usage
+*************
+
+**Upgrade an asset**
```
bpoldrack, line 14: Doesn't need to be done in this PR, but for documentation/examples I'd go for a more general description. It's changing the YAML file manually. The notion of an "upgrade" is nothing that is built in or that `onyo` inherently understands.

docs/source/cmd_get.rst
```diff
@@ -6,3 +6,13 @@ onyo get
    :func: setup_parser
    :prog: onyo
    :path: get
+
+
+Example Usage
+*************
+
+**Show all display sizes of the accounting group**
```
bpoldrack, line 14: Similar to above: This is about showing the values of a particular key for all assets underneath a particular path.
Examples that are only making sense in a specific repository should somehow reference such a repository (demo in this case), but point out the general concept of the example call.

bpoldrack (approved):
LGTM.

Started commenting on the general approach to examples. Applies pretty much everywhere, but I understand this is not the aim of the PR.

PR #388, opened 2023-06-15
RFC: Tests for OnyoRepo

TobiasKadelka:
This should ideally be reviewed soon, so as to have somewhat of a basis/blueprint for lib tests, hence I am creating a PR for something that is not yet complete.


onyo/lib/tests/test_onyo.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_OnyoRepo_instantiation_non_existing(tmp_path: Path) -> None:
+    """
+    The OnyoRepo class must instantiate correctly for paths to non-existing
+    repositories.
+    """
+    new_repo = OnyoRepo(tmp_path, init=True)
+    assert new_repo.git.root.samefile(tmp_path)
+    assert (new_repo.git.root / ".onyo").exists()
```
bpoldrack, line 23: I'd add an assertion on a clean repo /fsck here as well, since so far this only tests what's written to disc - not what's committed.

bpoldrack (commented):
This looks the right direction.

I would prefer to have all the invalid calls in one test, though.

bpoldrack (approved):

Issue #341, opened 2023-04-26
API/BUG: normalize arguments and add defaults for "command functions" (e.g. `set`, `unset`)


TobiasKadelka:
These two problems are typical for multiple "command functions" across `onyo.py`:
- A function has all arguments we need for the command line, but no or (no reasonable) defaults, which makes it hard or annoying to use the API.
- Sometimes the types or existing flags along commands are not identical, even if I would assume them to be when working with the API (which is especially relevant for get/set/unset)

**Example** for set, unset and get (but I assume similar problems happen elsewhere):
- `def set(self, paths: Iterable[Union[Path, str]], values: Dict[str, Union[str, int, float]], dryrun: bool, rename: bool, depth: Union[int])`
- `def unset(self, paths: Iterable[Union[Path, str]], keys: list[str], dryrun: bool, quiet: bool, depth: Union[int])`
- `def get(self, keys: Set[str], paths: Set[Path], depth: Union[int, None] = None, filters: Union[list[Filter], None] = None)`

**Problems**:
- set uses `values` as dict, unset uses `keys` as `list[str]` and get has keys as `Set[str]`, which is round about as inconsistent as possible. I understand that values for set is a dict, but it is still awkward
- why has unset a `quiet` but not the other functions? (should be removed)
- why is it `Union[int]` instead of just `int` for `depth` in set and unset?
- set and unset have `paths: Iterable[Union[Path, str]]`, get has paths as `Set[Path]`

**Solutions**:
- All commands should have reasonable defaults, so it is easy to call them without remembering every argument.
- Some arguments should probably be removed altogether (e.g. `quiet` of unset, possibly during #255/#338).
- The variable names, types and type hints should be the same for variables that are literally the same. 
- Default values need to be added and normalized, too.

TobiasKadelka:
The default values for `onyo new` should be re-thought:

- `--template` has as default `[]`. I think you can (or at least should) only be able to specify a single template file anyways, never an array. This should probably changed to "None" or something instead.
- `--KEYS` on the other hand has no default for `onyo new` (but for `onyo get` it is `[]`)
- `--tsv` has `None` as a default, instead of the nothing that `--keys` for `onyo new` currently has. This one might make sense, but I still think we should reflect on how to declear empty/None values as defaults and do it consistently whenever reasonably possible - instead of sometimes not declearing anything, other times have empty arrays/strings, and other times using `None` explicitely.

[...]

PR #389, opened 2023-06-30
docs: change texts for arguments in argparser and doc-strings of commands

TobiasKadelka:
We have multiple places in the python code that are used to auto-generate our onyo docs: https://onyo.readthedocs.io/en/latest/index.html

The changes of this PR focus on updating these texts to have better onyo docs; this PR does not contain the discussed move of arg-parser code into the commands (I was unable to do this, and after some hours I decided to focus on the texts for now).

Some of my changes correct text, others remove incorrect (or unneeded) information, sometimes I add something, or I just normalize information, so that we use the same terms along commands.

The relevant main changes of this PR:
- change texts in (sub-)command flags and arguments
- change docstrings in command-functions used to autogenerate docs
- normalize empty/None default values in argparser (works toward #341 
  - this is the only change were I had to update some 3 lines of actual code. We had so many different ways in which we declared that default values are None (actual "None", but also empty arrays, "." and such), that were visible in the command line reference of the docs. Basically, I removed all of those None values, since None is the default anyways.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -30,11 +30,11 @@ Options
 
 ``onyo.history.interactive``
     The command used to display history when running ``onyo history``. (default:
-    "tig --follow")
+    ``tig --follow``)
 
 ``onyo.history.non-interactive``
     The command used to print history when running ``onyo history`` with
-    ``--non-interactive``.  (default: "git --no-pager log --follow")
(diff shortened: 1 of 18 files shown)


bpoldrack (commented):
I'd prefer to define the defaults in the underlying python functions where possible. Python API commands should be equivalent after all.

Issue #383, opened 2023-06-06
Create diff in memory


bpoldrack:
Related: #377

We rely on `git diff` in a lot of places, largely requiring us to write to disc before knowing whether this is actually wanted and valid. This comes in two aspects:
- use `diff` to figure out what we have done before - this is entangled with staging (#370) and dryruns (#377)
- use `diff` to show asset metadata differences to apply and ask for user confirmation before committing

The former seems superfluous. If the code is structured right, we know exactly what have done or would do. No need to query git.
For the latter, we should consider [difflib](https://docs.python.org/3/library/difflib.html). First example from playing with it:

```
In [1]: from ruamel.yaml import YAML

In [2]: from difflib import unified_diff

In [3]: from io import StringIO

In [4]: s1 = StringIO()

In [5]: s2 = StringIO()

In [6]: yaml = YAML(typ='rt', pure=True)

In [7]: a = {'a': 'value'}

In [8]: b = {'a': 'value2', 'b': 'more'}

In [9]: yaml.dump(a, stream=s1)

In [10]: yaml.dump(b, stream=s2)

In [11]: [print(line) for line in unified_diff(s1.getvalue().splitlines(keepends=True), s2.getvalue().splitlines(keepends=True))]
--- 

+++ 

@@ -1 +1,2 @@

-a: value

+a: value2

+b: more

```

Issue #377, opened 2023-06-05
Rework dryrun approach


bpoldrack:
Currently, quite a bunch of commands have a dryrun option and this is done by actually doing things and rolling-back afterwards. Apart from the superfluous staging of such to-be-rolled-back changes (see #370), most git commands actually do have a dryrun themselves, that would report what they would have done.
Utilize that!

TobiasKadelka:
I think the existing #269 might already describe this problem that dry-run modifies temporarly and rolls changes back, instead of doing an actual dry-run.

FYI: This approach is used, because while sub-optimal, it at least prints a valid/real diff, while trying to create a diff manually on our own always led to problems in certain situations. But every other approach that replaces the current behavior is wished for - and just not found in an elegant efficient way yet :)

bpoldrack:
@TobiasKadelka 
> For mv, rm and others with git-equivalent we already do that.

Depends on where. ;-)

So, `onyo mv` does use `git mv` w/ dryrun.
`onyo set`, however, also uses `git mv` but does actually call it on dryrun and then restore.

bpoldrack:
> it at least prints a valid/real diff, while trying to create a diff manually on our own always led to problems in certain situations.

Not sure, what those issues were. Generally, I think it's worth exploring [difflib](https://docs.python.org/3/library/difflib.html) with the aim to only generate the to be written content in memory, diff-ing against current file content and only write after confirmation.
I have not used difflib but at a quick glance it looks like it should be a fit for us.

Issue #370, opened 2023-06-01
Reconsider staging modifications


bpoldrack:
There are at least two aspects to this:

- Some commands accumulate changes, stage them as they happen only to then possibly roll them back, instead of first collecting what needs doing and do it (in bulk) once we know we actually want to. Even if (content) modifications are done before we are sure in order to generate a diff the user is then asked to confirm, the staging isn't needed.
- More generally the pattern stage everything and then commit what's staged isn't useful for onyo. One can commit modifications directly, saving a potentially expensive git call.

I don't think I've seen a case were staging is actually useful in the existing call paths.

bpoldrack:
Note on exploring this:

The concept is baked into the code base rather deep. One thing to take care of first is `generate_commit_message`, which is based on staged files.

bpoldrack:
Also note, that this is heavily entangled with another pattern in the code base: Doing modifications, then ask for confirmation and possibly do a rollback. In most of the cases, this should be collecting what needs doing, possibly confirm and only then do anything (no need for rollback). This is entangled with staging, because the code is using git-add with everything that **may** have been modified and then check the staged changes for what actually **was** modified. Hence, there's barely direct knowledge of what actually changed, which is required for rollback when trying to incrementally getting rid of staging (as opposed to rewrite everything in a huge PR addressing both problems everywhere at once).

PR #392, opened 2023-07-10
Fix up unset

bpoldrack:
This is a follow up on #382, addressing #370, #377, and #383 for `unset`, which now does not stage modifications in order to roll them back, but only writes to disc if actually needed. Like `set` before via #382, it uses difflib instead of `git-diff` in order to show to-be-applied changes.


TobiasKadelka:
LGTM, thanks!

PR #393, opened 2023-07-20
demo: use `onyo set --filter`

### LINKED ISSUE #266: DEMO: Update `onyo set` with filtering
Currently, the demo does:
`onyo set -y USB_A=2 */laptop_*.* */*/laptop_*.*`

This is done to demonstrate how to update a value for all laptops in the inventory.
Later (after `onyo get` exists) the laptops should either be selected with a flag of `onyo set`, or with the usage of a subshell which calls `onyo get`.

TobiasKadelka:
Before, we used a rather complicated example with wildcards to demonstrate how to set a value for all laptop assets.
Now, I updated the demo to use `--filter` instead, because since #353 `onyo set` has that functionality.

Close #266


bpoldrack:
Thanks!

Issue #339, opened 2023-04-25
Decide on a docstring style


bpoldrack:
In order to have
- consistent, easy to read docstring when actually looking at source code files, as well as
- proper rendering of developer docs

we should decide on a style that works for us. Generally, I like the [numpy](https://numpydoc.readthedocs.io/en/v1.1.0/format.html) style. With type hints being implemented already, it would be nice to figure whether this can easily be rendered and linted if we simply leave out the types in the docstrings.

Generally, actually rendering developer docs is a different task, but should probably be considered in conjunction with this issue.

- [ ] explore how to not make us write redundant type hints
- [ ] decide on a style
- [ ] document style in `README` (or a `CONTRIBUTING` doc)
- [ ] actually do it

aqw:
In the absence of a strong opinion, I like going with your experience. :-)

It looks like the numpy style is already introduced to the codebase, so I'd say it's fairly well decided to use it.

Regarding the type hints, there's a [numpy issue](https://github.com/numpy/numpydoc/issues/196) about this (of course, Yarik has already weighed in ;-). There's no resolution, but some have success with [sphinx-autodoc-typehints](https://pypi.org/project/sphinx-autodoc-typehints/).

aqw:
I have updated the issue title. If you feel that was premature, please share your objections. :-)

TobiasKadelka:
As far as I see, we decided to go redundant with the type hints after all, and begin applying it to all functions we touch.

Since we have more specific issues regarding documentation/docstrings, I close this issue.

PR #397, opened 2023-07-28
TST/DOC: add tests and use new docstring style for class `GitRepo`

TobiasKadelka:
This PR adds tests and docstrings for the `GitRepo` class, in parallel to the changes for `OnyoRepo` during #394.

Changes:
- adds tests to some functions of `GitRepo`
- adds or normalizes docstrings (based on discussion in #339)

### DIFF
--- a/onyo/lib/git.py
+++ b/onyo/lib/git.py
@@ -100,21 +100,24 @@ class GitRepo(object):
             self._files = self._get_files()
         return self._files
 
-    def clear_caches(self, files: bool = True) -> None:
+    def clear_caches(self,
+                     files: bool = True) -> None:
         """
-        Clear caches of the instance of the repository object.
+        Clear caches of the instance of the GitRepo object.
 
-        Paths such as files, assets, and directories are cached, and can become
-        stale when the repository contents are modified. This function clears
-        the caches of the properties.
-
-        By default all caches are cleared, and arguments make it possible to
-        specify which caches which should remain set.
+        Paths to files in git are cached, and can become stale when the
+        repository contents are modified. By default, this function clears the
+        cache of all properties of the GitRepo.
 
         If the repository is exclusively modified via public API functions, the
-        caches of the `Repo` object are consistent. If the repository is
+        caches of the `GitRepo` object are consistent. If the repository is
         modified otherwise, this function clears the caches to ensure that the
         caches do not contain stale information.
+
+        Parameters
+        ----------
+        files: boolean
+            Deactivate the resetting of the file cache.
         """
         if files:
             self._files = None
--- a/onyo/lib/git.py
+++ b/onyo/lib/git.py
@@ -44,7 +44,26 @@ class GitRepo(object):
 
     @staticmethod
     def find_root(path: Path) -> Path:
-        """Returns the git worktree root `path` belongs to"""
+        """
+        Returns the git worktree root `path` belongs to.
+
+        Parameters
+        ----------
+        path: Path
+            The path to identify the git worktree root for. This can be any
+            sub-directory of the repository, or the root directory itself.
+
+        Returns
+        -------
+        Path
+            An absolute path to the root of the git worktree where `.git/` is
+            located.
+
+        Raises
+        ------
+        OnyoInvalidRepoError
+            If `path` is not inside a git repository at all.
+        """
         root = None
         try:
             ret = subprocess.run(["git", "rev-parse", "--show-toplevel"],
--- a/onyo/lib/git.py
+++ b/onyo/lib/git.py
@@ -211,7 +211,12 @@ class GitRepo(object):
 
     def is_clean_worktree(self) -> bool:
         """
-        Check if the working tree for git is clean. Returns True or False.
+        Check if the working tree for git is clean.
+
+        Returns
+        -------
+        Boolean
+            True if the git worktree is clean, otherwise False.
         """
 
         changed = {str(x) for x in self.files_changed}
--- a/onyo/lib/git.py
+++ b/onyo/lib/git.py
@@ -8,6 +8,20 @@ log: logging.Logger = logging.getLogger('onyo.git')
 
 
 class GitRepo(object):
+    """
+    An object to get and set git information, and to call git functions with.
+
+    Attributes
+    ----------
+    root: Path
+        The absolute path to the directory of the git worktree root.
+
+    files: set of Paths
+        A property containing the absolute Path to all files tracked in git.
+        This property is cached and consistent when only the public functions of
+        GitRepo are called. Usage of private or external functions might
+        require a manual reset of the cache with `GitRepo.clear_caches()`.
+    """
 
     def __init__(self,
                  path: Path,
@@ -15,11 +29,18 @@ class GitRepo(object):
         """
         Instantiates a `GitRepo` object with `path` as the root directory.
 
-        If `find_root=True` searches the root of a worktree from `path`.
+        Parameters
+        ----------
+        path: Path
+            An absolute path to the root of a git repository.
+
+        find_root: bool
+            `find_root=True` allows to search the root of a git worktree from a
+            sub-directory, beginning at `path`, instead of requiring the root.
         """
         self.root = GitRepo.find_root(path) if find_root else path.resolve()
 
-        self._files: Union[set[Path], None] = None
+        self._files: Optional[set[Path]] = None
 
     @staticmethod
     def find_root(path: Path) -> Path:
--- a/onyo/lib/git.py
+++ b/onyo/lib/git.py
@@ -312,11 +312,22 @@ class GitRepo(object):
         # call. To always secure the integrity of the caches, we reset them all.
         self.clear_caches()
 
-    def commit(self, *args) -> None:
+    def commit(self,
+               *args) -> None:
         """
-        Perform a ``git commit``. The first message argument is the title; each
-        argument after is a new paragraph. Messages are converted to strings.
-        Lists are printed one item per line, also converted to a string.
+        Perform a ``git commit``.
+
+        Parameters
+        ----------
+        args: tuple
+            Components to compose the commit message from. At least one is
+            required. The first argument is the title; each argument after it is
+            a new paragraph. Lists and sets are printed one item per line.
+
+        Raises
+        ------
+        ValueError
+            If no commit message is provided.
         """
         if not args:
             raise ValueError('at least one commit message is required')
--- a/onyo/lib/git.py
+++ b/onyo/lib/git.py
@@ -141,8 +141,19 @@ class GitRepo(object):
         # subprocess).
         self.clear_caches()
 
-    def restore(self, paths: Union[list[Path], Path]) -> None:
-        """Call git-restore on `paths`.
+    def restore(self,
+                paths: Union[list[Path], Path]) -> None:
+        """
+        Call git-restore on `paths`.
+
+        This does restore files which contain changes, but it does not restore
+        changes that are already staged.
+
+        Parameters
+        ----------
+        paths: List of Paths
+            The absolute Paths to the files or directories which are to be
+            `git restore`d.
         """
         if not paths:
             log.debug("No paths passed to restore. Nothing to do.")
--- a/onyo/lib/git.py
+++ b/onyo/lib/git.py
@@ -279,11 +279,22 @@ class GitRepo(object):
         """
         return '.git' in path.parts or path.name.startswith('.git')
 
-    def add(self, targets: Union[Iterable[Path], Path, str]) -> None:
+    def add(self,
+            targets: Union[Iterable[Path], Path]) -> None:
         """
         Perform ``git add`` to stage files.
 
-        Paths are relative to ``repo.root``.
+        If called on files without changes, it does not raise an error.
+
+        Parameters
+        ----------
+        targets: List of paths
+            Paths are relative to ``repo.root``.
+
+        Raises
+        ------
+        FileNotFoundError
+            If a path in `targets` does not exist.
         """
         if isinstance(targets, (list, set)):
             tgts = [str(x) for x in targets]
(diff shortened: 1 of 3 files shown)


onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,264 @@
...
+from onyo.lib.git import GitRepo
+
+
+def test_GitRepo_instantiation_existing(repo: OnyoRepo) -> None:
+    """
+    The GitRepo class must instantiate correctly for paths to existing
+    repositories.
+    """
+    new_repo = GitRepo(repo.git.root)
+    assert new_repo.root.samefile(repo.git.root)
```
bpoldrack, line 14: General remark on that (not only this particular test): Using the `repo` fixture here, seems a bit off. `GitRepo` doesn't require an `OnyoRepo`. Ever.

Also: If the fixture was able to instantiate an `OnyoRepo` that delivers a `.git` which you can use to pass `repo.git.root`, then there already was a successful instantiation of a `GitRepo` on that path. Hence, this test does nothing.

bpoldrack: What you'd want would be a temporary directory, run `git init` on it via `subprocess` and then try and instantiate this object.

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+@pytest.mark.repo_dirs('existing/directory')
+def test_GitRepo_instantiation_find_root(repo: OnyoRepo) -> None:
+    """
+    The GitRepo class must instantiate correctly with paths to sub-directories
+    in a repository when `find_root=True`.
```
bpoldrack, line 21: I'd turn this into a test of the `find_root` parameter, not just one case for it.
So, yes, it should instantiate with `find_root=True`. But it should also fail with `find_root=False`.

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    new_repo = GitRepo(repo.git.root / 'existing/directory', find_root=True)
+    assert new_repo.root.samefile(repo.git.root)
+    assert (new_repo.root / '.git').exists()
+
+
+def test_GitRepo_find_root(tmp_path: Path) -> None:
```
bpoldrack, line 28: So, generally we are aiming at unit tests here. Meaning that any method (`GitRepo.find_root` in this case) should come up with a test, that defines its behavior. (On that note: I'd also fuse the instantiation tests into one. One test that tells us, whether `GitRepo`'s instantiation is broken or not).

Here you only have half of it. Two cases of failure. No case of success, though.

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,264 @@
...
+    If no files are staged, it should not raise an error.
+    """
+    new_git = GitRepo(repo.git.root)
+    test_file = Path('a/test/asset_for_test.0').resolve()
+    test_file2 = Path('a/test/asset_for_test.1').resolve()
+
+    # test that no error is raised, when called on a clean repository
+    new_git.restore_staged()
+
+    # change two files and add them, so they are "staged for commit"
```
bpoldrack, line 55: If you are only testing whether staged files are restored, what does the second file add?
What kind of bug would it detect? There's nothing different about the second file.
Also: This again tests only part of the behavior. You state "restore all staged files in the repository. If no files are staged, it should not raise an error." What about untracked files or modified files that are not staged? What happens? That's what you could use the second (third, etc.) file for.

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,264 @@
...
+
+    # call restore_staged() and verify that the changes are not staged anymore
+    new_git.restore_staged()
+    assert test_file not in new_git.files_staged
+    assert test_file2 not in new_git.files_staged
+
+
+@pytest.mark.repo_files('a/test/asset_for_test.0',
+                        'a/test/asset_for_test.1')
+def test_GitRepo_restore(repo: OnyoRepo) -> None:
```
bpoldrack, line 70: Similar here: You tested a modified file (and that another modified file that wasn't passed to the call is unaffected). What about passing staged changes? What about untracked files?

Also: You test "that no error is raised, when called on a clean repository". That's right. But how about checking whether the repo is still clean afterwards? Just because nothing was raised, doesn't mean nothing did happen, right?

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,264 @@
...
+
+@pytest.mark.repo_files('a/test/asset_for_test.0')
+def test_GitRepo_clear_caches(repo: OnyoRepo) -> None:
+    """
+    The function `GitRepo.clear_caches()` must allow to empty the cache of the
+    GitRepo, so that an invalid cache can be re-loaded by a new call of the
+    property.
+    """
+    new_git = GitRepo(repo.git.root)
+    file = Path('a/test/asset_for_test.0').resolve()
```
bpoldrack, line 103: This `resolve` confuses me (may be I'm wrong about how the fixtures work - I hope I'm not). Shouldn't that path be relative to the repo, when you use it with `pytest.mark.repo_files`? Now you treat it as being relative to CWD. That makes no sense. That would change the test depending on where I run the test from.

And if I'm wrong, we'd need to change that. Running the tests should never create things in my CWD!

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,264 @@
...
+    assert not file.exists()
+
+    # test GitRepo.clear_caches() fixes the cache
+    new_git.clear_caches(files=True)
+    assert file not in new_git.files
+
+
+def test_GitRepo_is_clean_worktree(repo: OnyoRepo) -> None:
+    """
+    `GitRepo.is_clean_worktree()´ must return True when the worktree is clean,
```
bpoldrack, line 122: Here you actually go through a whole bunch of cases for this function. Good. :)

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    new_git = GitRepo(repo.git.root)
+
+    # Test the examples listed above:
+    assert new_git.is_git_path(new_git.root / ".git")
+    assert new_git.is_git_path(new_git.root / ".git" / "HEAD")
+    assert new_git.is_git_path(new_git.root / ".git" / "info")
+    assert new_git.is_git_path(new_git.root / ".gitignore")
+    assert new_git.is_git_path(new_git.root / ".gitattributes")
+    assert new_git.is_git_path(new_git.root / ".gitmodules")
```
bpoldrack, line 171: Many cases, but they don't differ much. What's different between `.git/HEAD` and `.git/info`? Why only test existing files?
Also: The point of `.git*` files is, that they can be anywhere.

```suggestion
    assert new_git.is_git_path(new_git.root / ".git")
    assert new_git.is_git_path(new_git.root / ".git" / "HEAD")
    assert new_git.is_git_path(new_git.root / ".git" / "doesnotexist")
    assert new_git.is_git_path(new_git.root / ".gitignore")
    assert new_git.is_git_path(new_git.root / ".gitdoesnotexist")
    assert new_git.is_git_path(new_git.root / 'existing' / ".gitattributes")
```

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,264 @@
...
+
+    # Must return False
+    assert not new_git.is_git_path(new_git.root)
+    assert not new_git.is_git_path(new_git.root / ".onyo")
+    assert not new_git.is_git_path(new_git.root / "existing")
+    assert not new_git.is_git_path(new_git.root / "existing" / "directory")
```
bpoldrack, line 177: This doesn't add anything I can see to the previous assertion.
What's missing, however, are not existing paths.
What would also be missing are file/dir names that contain `.git` (but don't start with it) or just `git` (including the start, because it's only supposed to match `.git` as the start of the name.

```suggestion
    assert not new_git.is_git_path(new_git.root / "existing" / "git_no_.git")
```

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,264 @@
...
+    assert not new_git.is_git_path(new_git.root / "existing" / "directory" /
+                                   "test_file.txt")
+
+
+@pytest.mark.repo_files('existing/directory/test_file.txt')
+def test_GitRepo_add(repo: OnyoRepo) -> None:
```
bpoldrack, line 183: What happens if I call `GitRepo.add` with a tracked, unmodified file?

onyo/lib/tests/test_git.py
TobiasKadelka: > Hence, this test does nothing.

That is not fully true. This test-function verifies for a path/repository that already exists as git, that calling `GitRepo(root)` instantiates the GitRepo object (not the git repository itself) and sets the root for the new object correctly.

> What you'd want would be a temporary directory, run `git init` on it via `subprocess` and then try and instantiate this object.

I agree, and will add such a test.

onyo/lib/tests/test_git.py
TobiasKadelka: I will change them all to paths from the repo.root.

onyo/lib/tests/test_git.py
TobiasKadelka: Good point, I will add to the test and to the docstring of the function that it does not raise an error.

onyo/lib/tests/test_git.py
TobiasKadelka: Add-on explenation: I do not add "such a test", but change all OnyoRepos to temp_paths and initialize them with git init and subprocess like you described.

onyo/lib/tests/test_git.py
TobiasKadelka: > But it should also fail with find_root=False.

Indeed it should have been tested that it fails, because it did not. I created the issue #399 for it, and fixed it. I then proceded and added the test for it!

Thanks for the comment!

onyo/lib/tests/test_git.py
bpoldrack: > This test-function verifies for a path/repository that already exists as git, that calling GitRepo(root) instantiates the GitRepo object (not the git repository itself) and sets the root for the new object correctly.

Well not really. This needed to succeed in the fixture before the test function was even executed. It tested its own prerequirement.

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,340 @@
...
+    assert changed_file not in new_git.files_untracked
+    assert changed_file not in new_git.files_staged
+
+
+def test_GitRepo_clear_caches(tmp_path: Path) -> None:
+    """
+    The function `GitRepo.clear_caches()` must allow to empty the cache of the
+    GitRepo, so that an invalid cache can be re-loaded by a new call of the
+    property.
+    """
```
bpoldrack, line 156: No need to change the test really, but I do want to point out something:
All you want to test here, is that the cache is cleared, not how the cache is built to begin with. Meaning: There's no need to pay the cost to of writing to files and commit stuff. You could simply directly assign a set of paths (don't need to exist) to `GitRepo._files`, check that the property returns them, and then call `clear_caches` to reset it.

bpoldrack (commented):
Thanks, @TobiasKadelka. Better :)

Left some more comments. And one more general comment:
Switching to temp paths being delivered to the test functions is better. But: Now you get that path and every single test function starts with the exact same thing: `git init` on that path. I would have imagined to "simply" have a fixture, that delivers an initialized git repo (and more specific types of repos could built on it eventually).

onyo/lib/git.py
```diff
@@ -49,38 +111,43 @@ def _git(self, args: list[str], *, cwd: Optional[Path] = None, raise_error: bool
...
         If the repository is exclusively modified via public API functions, the
-        caches of the `Repo` object are consistent. If the repository is
+        caches of the `GitRepo` object are consistent. If the repository is
         modified otherwise, this function clears the caches to ensure that the
         caches do not contain stale information.
+
+        Parameters
+        ----------
+        files: boolean
+            Deactivate the resetting of the file cache.
```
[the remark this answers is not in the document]
TobiasKadelka: Thanks, will change!
While re-reading this after my holidays I directly see what you meant! ;)

onyo/lib/tests/test_git.py
TobiasKadelka: Thanks -- I will then probably not change it here, but definetely keep it in mind for future tests!

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,340 @@
...
+
+    # initialize a new git repo to test correct behavior when git exists
+    subprocess.run(['git', 'init', tmp_path])
+    subprocess.run(['mkdir', '-p', tmp_path / 'existing/directory'])
+
+    # test that `GitRepo.find_root(root)` returns the same root path
+    assert GitRepo.find_root(tmp_path).samefile(tmp_path)
+
+    # test that `GitRepo.find_root()` returns correct root for a sub-directory
+    assert GitRepo.find_root(tmp_path / 'existing/directory').samefile(tmp_path)
```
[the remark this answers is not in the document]
TobiasKadelka: Thanks, I added another test case to the function that tests the error for a non-existing sub-path under an existing root!

onyo/lib/tests/test_git.py
```diff
@@ -0,0 +1,345 @@
...
+    """
+    subprocess.run(['git', 'init', tmp_path])
+    new_git = GitRepo(tmp_path)
+    test_file = new_git.root / 'asset_for_test.0'
+    test_file.touch()
+    new_git.stage_and_commit(test_file, message="Create file for test")
+
+    # test that no error is raised, when called on a clean repository
+    new_git.restore_staged()
+
```
bpoldrack, line 83: There's no assertion on what `restore_staged` did or didn't do, just that it doesn't raise. I think minimum is an assertion that the repo is still clean. One could also test, that it's not only clean but still on the same commit.
There are similar cases in other tests, where you call the to-be-tested function and have no assertions on the result.

bpoldrack:
Thanks, @TobiasKadelka !

PR #400, opened 2023-09-14
onyo: define arguments in dedicated locations, read them generically in main.py

TobiasKadelka:
This continues the efforts to define the arguments in dedicated locations (either in the commands itself, in shared_arguments.py or in the newonyo_arguments.py) instead of in main.py; main.py just reads the other locations and creates the parsers for arguments generically.

The main changes are:
1. every command has just one argument dictionary which can be read in main.py
2. shared_arguments get imported in commando-scripts and removed from main.py
3. onyo_args are moved/defined in a seperate file, this includes:
	- yes and quiet are now onyo flags, not command flags
	- --version, -C, and --debug are also there
4. corrects tests, removes old code, etc

### DIFF
--- a/demo/generate_demo_repo.sh
+++ b/demo/generate_demo_repo.sh
@@ -87,83 +87,83 @@ cd "$DEMO_DIR"
 onyo init
 
 # setup basic directory structure
-onyo mkdir --yes warehouse
-onyo mkdir --yes recycling
-onyo mkdir --yes repair
+onyo --yes mkdir warehouse
+onyo --yes mkdir recycling
+onyo --yes mkdir repair
 
 # import some existing hardware
 # TSV files can be very useful when adding large amounts of assets
-onyo new -y --tsv "${SCRIPT_DIR}/inventory.tsv"
+onyo -y new --tsv "${SCRIPT_DIR}/inventory.tsv"
 
 # add a set of newly bought assets
-onyo new -y --keys RAM=8GB display=13.3 --path warehouse/laptop_apple_macbook.9r32he
-onyo new -y --keys RAM=8GB display=13.3 --path warehouse/laptop_apple_macbook.9r5qlk
-onyo new -y --keys RAM=8GB display=14.6 --path warehouse/laptop_lenovo_thinkpad.owh8e2
-onyo new -y --keys RAM=8GB display=14.6 --path warehouse/laptop_lenovo_thinkpad.iu7h6d
-onyo new -y --keys RAM=8GB display=12.4 touchscreen=yes --path warehouse/laptop_microsoft_surface.oq782j
+onyo -y new --keys RAM=8GB display=13.3 --path warehouse/laptop_apple_macbook.9r32he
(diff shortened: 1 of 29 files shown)


bpoldrack (commented):
Thanks, @TobiasKadelka. Overall seems to be going where I thought it should go.

There's a couple of things, though:

- Not clear to me why things like the type definitions and `StoreKeyValuePair` went into `shared_arguments.py`. Can you explain the thought?
- `--yes` and `--quiet` still exist in `shared_arguments.py` (in addition to the new place) and are used from there (by `edit` for example). Why?
- I am not too familiar with argsparse details myself, so this is just a question: Why do most definitions have a `dest` key and some don't?
- Related: Wherever there's a `dest` the dict looks like this: `'something': {'dest': 'something', ...}}`. Seems to me, `build_parser` could simply take the key and pass it to `dest` w/o `dest` being explicitly defined (repeating the string definition). Not too important, but would reduce the opportunities to make mistakes with duplicate string definitions that are supposed to match exactly.

TobiasKadelka:
1. type definitions and `StoreKeyValuePair` went into `shared_arguments.py` because it is the only place were it is ever used. When I look into what happens, I would look into that file, where it is used (type definitions and `StoreKeyValuePair` are literally shared by all the arguments). The alternative would be to keep it in `main.py`, where these things feel like they are just taking up space, while being consumed in the other file, just to be re-imported in `main.py`.

2. `--yes` and `--quiet` are now removed from `shared_arguments.py`

3. argparse does automatically decide on `dest` for most commands based on the argument, but in some cases it does not use the one we want a different one, e.g. `opdir` instead of '-C', '--onyopath'.
So we might later normalize this, but basically `dest` is added whenever it does not work automatically, and this is not an accident.

4. Yeah could be done that way, I guess

TobiasKadelka:
Update: I removed `dest` in all arguments and set it now always based on the dictionary name.

bpoldrack:
> type definitions and StoreKeyValuePair went into shared_arguments.py because it is the only place were it is ever used. When I look into what happens, I would look into that file, where it is used (type definitions and StoreKeyValuePair are literally shared by all the arguments). The alternative would be to keep it in main.py, where these things feel like they are just taking up space, while being consumed in the other file, just to be re-imported in main.py.

Well, if those are currently only used with shared arguments, that's coincidence. Nothing in the "nature" of `StoreKeyValuePair` or the type definitions is about shared arguments. We can introduce a command specific argument that needs those anytime. So, if you don't want to have it in `main` (which I understand), then I think it's appropriate to put it in a third place, like `argparse_helpers`.

> --yes and --quiet are now removed from shared_arguments.py

Thanks!

> argparse does automatically decide on dest for most commands based on the argument, but in some cases it does not use the one we want a different one, e.g. opdir instead of '-C', '--onyopath'.
So we might later normalize this, but basically dest is added whenever it does not work automatically, and this is not an accident.

Ok, I see.

> Yeah could be done that way, I guess
>  I removed dest in all arguments and set it now always based on the dictionary name.

Great!

TobiasKadelka:
I added `argparse_helpers.py`, moved the discussed code into it, and adjusted the imports everywhere accordingly. Thanks!

PR #401, opened 2023-10-13
UI: add `ui.py` and use it to replace prints, logs, etc

### LINKED ISSUE #255: DOC: logging/print overhaul
The current proposed solution:

- commands only use `print()` (no logging)
- Classes and Methods only use `log` (no printing)
- CLI logging will use a custom formatter to remove the "onyo:INFO" (etc) prefix.
- CLI logging will split INFO level and print it to stdout. (https://stackoverflow.com/questions/1383254/logging-streamhandler-and-standard-streams)
- Logging default will be bumped up to warning, as per default.
- Child loggers will be created for each public method (e.g. `onyo.fsck`). This will be achieved using decorators (e.g. `@log('fsck')`) and applied to all related private methods, so that they all inherit the same log level.
- If a CLI wants its "own" method to have greater verbosity (for example: CLI `fsck` wants `Repo.fsck()` INFO level logging), then that will be set for that specific child logger.
- CLI `quiet` modes will either set the log level to CRITICAL or pass in a null logger handle

Once this is done, then the tests for commands can be overhauled, and they can focus on the user experience (exit codes, output, etc) and path handling, rather than functionality (which is well covered by the `Repo` tests).

TobiasKadelka:
This PR will close #255

Main changes:
- The UI object is once created
- parameters (e.g. `--yes`, `--quiet`, `--debug`) are set in `main.py`
- all other parts of Onyo import the ui object
- all printing, logging, erroring, and other user interactions are handled with this object!

TODOs:
- `onyo get` is implemented quite differently to the rest; still need to unify for UI:
  - rich
  - machine_readable
  - sort ascending/descending
  - interactive
- unify `interactive`/`non-interactive` for all commands
- improve logging:
  - should everything, including assets.py use the new form ui-logging?
  - currently logging functions of `UI` just receive a string
- add comments etc
- add tests

### DIFF
--- a/onyo/__init__.py
+++ b/onyo/__init__.py
@@ -1,16 +1,17 @@
-import logging
 from onyo._version import __version__
 from onyo.lib import (
-    Filter, OnyoRepo, OnyoInvalidRepoError,
-    OnyoProtectedPathError, OnyoInvalidFilterError)
+    Filter,
+    OnyoRepo,
+    OnyoInvalidRepoError,
+    OnyoProtectedPathError,
+    OnyoInvalidFilterError,
+    UI,)
 from onyo.onyo_arguments import args_onyo
 
-logging.basicConfig(level=logging.ERROR)  # external logging level
-log: logging.Logger = logging.getLogger('onyo')  # internal logging level
-log.setLevel(level=logging.INFO)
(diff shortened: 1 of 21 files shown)


TobiasKadelka:
Closed for #402

PR #402, opened 2023-10-13
UI: add `ui.py` and use it to replace prints, logs, etc

### LINKED ISSUE #255: DOC: logging/print overhaul

TobiasKadelka:
This PR will close #255

Main changes:
- The UI object is once created
- parameters (e.g. `--yes`, `--quiet`, `--debug`) are set in `main.py`
- all other parts of Onyo import the ui object (except if it would cause circular imports)
- all printing, logging, erroring, and other user interactions are handled with this object

TODOs:
- `onyo get` is implemented quite differently to the rest; still need to unify for UI:
  - rich
  - machine_readable
  - sort ascending/descending
  - interactive
- unify `interactive`/`non-interactive` for all commands
- should everything, including assets.py use the new form ui-logging? This leads to Circular imports
- add tests

### DIFF
--- a/onyo/lib/ui.py
+++ b/onyo/lib/ui.py
@@ -1,0 +1,210 @@
+import logging
+import os
+import sys
+from pathlib import Path
+from typing import Union
+
+from traceback import format_exception
+
+logging.basicConfig()
+log: logging.Logger = logging.getLogger('onyo')
+
+
+class UI(object):
+    """
+    An object handling user interaction, including printing, errors, requests,
+    and others.
+
+    Attributes
+    ----------
+    log: Logger
+        The logger to display information with.
+    quiet: boolean
+        Activate the quiet mode (requires that `yes=True`).
+        This will suppresses all output generation.
+    yes: boolean
+        Activate the yes mode, which suppresses all interactive requests to the
+        user, and instead answers them with yes.
+    """
+
+    def __init__(self,
+                 debug: bool = False,
+                 quiet: bool = False,
+                 yes: bool = False) -> None:
+        # TODO: interactive mode with default values or autodetecting tty? And
+        # should this be unified with the whole business of rich-coloring etc?
+        """
+        Initialize the User Interface object for user communication of Onyo.
+
+        Parameters
+        ----------
+        debug: boolean
+            Activate the debug mode to display additional information via Onyo,
+            and to print the full traceback stack if errors occur.
+
+        quiet: boolean
+            Activate the quiet mode (requires that `yes=True`) to suppress all
+            output generation.
+
+        yes: boolean
+            Activate the yes mode to suppress all interactive requests to the
+            user, and instead answers them with yes.
+        """
+        # set the the attributes of the UI object
+        self.quiet = quiet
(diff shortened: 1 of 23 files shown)


onyo/lib/ui.py
```diff
@@ -0,0 +1,208 @@
...
+              error: Union[str, Exception],
+              end: str = "\n") -> None:
+        """
+        Print an error message, if the UI is not set to `quiet`.
+
+        Parameters
+        ----------
+        error: string or Exception
+            Prints the string, or the message of an error.
+            If debug mode is activated, displays the full traceback of an error.
```
bpoldrack, line 125:
```suggestion
            If debug mode is activated, displays the full traceback of an exception.
```

Just a minor clarification, since `error` is the name of the described parameter, which may be a string (hence no traceback).

onyo/lib/ui.py
```diff
@@ -0,0 +1,208 @@
...
+
+        end: str
+            Specify the string at the end of prints.
+            Per default, prints end with a line break.
+        """
+        if not self.quiet:
+            print(f"ERROR: {error}", file=sys.stderr, end=end)
+        self.logger.debug(format_exc())
```
bpoldrack, line 133: I'd use https://docs.python.org/3/library/traceback.html#traceback.format_exception and actually pass `error` (if it is an `Exception`. Otherwise you are relying on being in a context where an exception was just caught. We don't really know that here. Caller gave the exception object to consider explicitly. Let's use it.

onyo/lib/ui.py
```diff
@@ -0,0 +1,208 @@
...
+        Parameters
+        ----------
+        message: string
+            The string to display as debug information.
+        """
+        self.logger.debug(message)
+
+    def print(self,
+              message: Union[str, list, Path] = "\n",
+              end: str = "\n",
```
bpoldrack, line 161: docstring says "line break". That's `os.linesep`, not necessarily `\n`.

bpoldrack:
Ok, this headed where I want it to. Thanks, @TobiasKadelka - will merge and rebase #403

PR #404, opened 2023-10-20
TST/DOC: bump Python baseline 3.9 -> 3.11

aqw:
And add `3.12` to the test matrix.


aqw:
Number of failures is the same across both Python versions. For now, we consider that "good enough". ;-)

Merging.

Issue #349, opened 2023-05-09
TST: remove unneeded `Path()` constructor calls


TobiasKadelka:
While I have to update this behavior for my current PR #330, I think this should be applied to almost all tests, and possibly in a lot of places arround the whole code base. Here a quote from Ben describing the issue further:

> One thing is off across tests, though, and that is this kind of usage of the `/` operator for `Path` objects:
> 
> `assert Path(repo.root / asset).is_file()`
> 
> While this kinda works, it's unnecessary. `repo.root` is a `Path` object already, right?
> So `assert (repo.root / asset).is_file()` it is. There's no need to call a `Path` constructor again. And if `repo.root` was in fact a string it would still be:
> 
> `assert (Path(repo.root) / asset).is_file()`.
> 
> Or better yet, to have an easy to read assertion:
> 
> ```
> asset_path = repo.root / asset
> # alternative if repo.root was string:
> # asset_path = Path(repo.root) / asset
> assert asset_path.is_file()
> ```
> 

_Originally posted by @bpoldrack in https://github.com/psyinfra/onyo/pull/330#pullrequestreview-1412700606_

aqw:
Largely (if not entirely) addressed by #406.

PR #406, opened 2023-10-20
TST/BF: remove unneccesary `Path` invocations

aqw:
This addresses an issue identified in #349

The issue specifically identifies an anti-pattern in the tests of wrapping a `path` object with `Path()` when concatenating.

The majority of changes in this PR are indeed that, but I did not limit myself to the tests, nor just that particular unnecessary use of `Path()`. I made others simplifications when I encountered them.

I did not go through all invocations of `Path()` exhaustively, as that would be an extraordinary amount of effort. However, `git grep 'Path(' | grep ','` is now clean of this pattern, and I can say that likely the majority of these have been fixed.


bpoldrack (approved):
Thanks, @aqw !

PR #422, opened 2023-10-23
BF: Repair demo repository generation

TobiasKadelka:
The demo repository did not work anymore because of the new syntax, this PR fixes it.

Includes the newly merged changes regarding less whitespace/indentation in run-records.

This PR updates:
- the shell script generating the demo repository to the new syntax
- The git reference log to run the demo-test successfully
- the code in commands.py that generated was different for each command; this replaces the code and uses instead the function `generate_commit_message()` everywhere
- the code in `generate_commit_message()` was reduced. Before, it did generate (when `-m` was not used) a commit message subject and body. This was lately replaced with the run-records, so the commit message body generation could be completely removed from the function

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -337,7 +337,6 @@ def onyo_mv(inventory: Inventory,
                 op.operator == OPERATIONS_MAPPING['move_directories'] or
                 op.operator == OPERATIONS_MAPPING['rename_directories']]
             message = inventory.repo.generate_commit_message(
-                message="",
                 cmd=subject,
                 destination=destination,
                 modified=operation_paths)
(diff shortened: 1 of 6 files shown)


onyo/lib/commands.py
```diff
@@ -262,12 +259,10 @@ def onyo_mkdir(inventory: Inventory,
         ui.print(line)
     if ui.request_user_response("Save changes? No discards all changes. (y/n) "):
         if not message:
-            prefix = "mkdir: "
-            # TODO: This does not actually account for implicitly created dirs (and therefore duplicates).
-            #       However, not mentioning implicit ones separately may be an advantage for a message subject
-            message = prefix + f"{', '.join(p.relative_to(inventory.root).as_posix() for p in dirs)}"
-            if len(message) > 80:
-                message = prefix + f" {', '.join(p.name for p in dirs)}"
+            new_dirs = set([Path(line[1:]) for line in inventory.diff()])
```
bpoldrack, line 262: This is not a reliable approach to get the paths. They should come from `Inventory.operations`. I'll push into this PR accordingly.

TobiasKadelka: This was solved by your own push, and my adjustments to the other functions.

aqw (approved):
LGTM!

PR #446, opened 2023-10-24
Towards re-establishing CI baseline

bpoldrack:
This patch: 
- fixes `onyo cat` for the case of invalid assets files. New behavior is: Issue an error message, still print the content and return non-zero.
- fixes `onyo rm` for the case of removing a non-empty directory
- skips the tests for `unset` and `get` , since they are not currently expected to work

Should bring the number failures down to 87, 82(?) of which are actually the same issue in `set`.

### DIFF
--- a/onyo/commands/tests/test_unset.py
+++ b/onyo/commands/tests/test_unset.py
@@ -30,6 +30,9 @@ content_str: str = "\n".join([f"{elem}: {content_dict.get(elem)}"
 contents: List[List[str]] = [[x, content_str] for x in assets]
 
 
+pytest.skip("UNSET not currently implemented", allow_module_level=True)
+
+
 @pytest.mark.repo_contents(*contents)
 @pytest.mark.parametrize('asset', assets)
 def test_unset(repo: OnyoRepo, asset: str) -> None:
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -78,18 +78,21 @@ def fsck(repo: OnyoRepo, tests: Optional[list[str]] = None) -> None:
 
 
 def onyo_cat(repo: OnyoRepo, paths: Iterable[Path]) -> None:
-
+    from .assets import validate_yaml
     non_asset_paths = [str(p) for p in paths if not repo.is_asset_path(p)]
     if non_asset_paths:
         raise ValueError("The following paths are not asset files:\n%s",
                          "\n".join(non_asset_paths))
+    # TODO: "Full" asset validation. Address when fsck is reworked
+    assets_valid = validate_yaml(set(paths))
     # open file and print to stdout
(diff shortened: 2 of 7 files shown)


onyo/commands/tests/test_unset.py
```diff
@@ -30,6 +30,9 @@
 contents: List[List[str]] = [[x, content_str] for x in assets]
 
 
+pytest.skip("GET not currently implemented", allow_module_level=True)
```
TobiasKadelka, line 33: Is it right, that the message in `test_unset.py` does states "GET not currently implemented"?

onyo/lib/commands.py
```diff
@@ -78,18 +78,21 @@ def fsck(repo: OnyoRepo, tests: Optional[list[str]] = None) -> None:
...
+    # TODO: "Full" asset validation. Address when fsck is reworked
+    assets_valid = validate_yaml(set(paths))
     # open file and print to stdout
     for path in paths:
-        # TODO: Do we really not want to separate assets like print(f"{path.name}:{os.linesep}")?
         # TODO: Probably better to simply print dict_to_yaml(repo.get_asset_content(path)) - no need to distinguish
         #       asset and asset dir at this level. However, need to make sure to not print pointless empty lines.
         f = path / OnyoRepo.ASSET_DIR_FILE if repo.is_asset_dir(path) else path
         ui.print(f.read_text(), end='')
+    if not assets_valid:
```
TobiasKadelka, line 94: Why first do in line 87 `assets_valid = validate_yaml(set(paths))` and then do the check here?

The variable `assets_valid` gets not used anywhere else, so just `if not validate_yaml(set(paths)):` here would do the job.

Alternatively, instead of just `raise OnyoInvalidRepoError("Invalid assets")` it could list the invalid assets here, in case `onyo cat` gets called on multiple assets, and just a subset of them is invalid.

bpoldrack: The listing of invalid assets is done (from within `validate_yaml`).
A raise is required to exit non-zero in the CLI case and be consistent in that regard (raise in python == non-zero exit in CLI).

So, yes, we could move the `validate_yaml` call down, which would lead to showing the errors on invalid paths after instead of before the content in question. No strong opinion on that.

However, I'd not do it right now, because I'd like to get to validation behave differently and then ultimately raise a different error with a structured report on what's wrong instead of the one raise now and the error messages issued by `validate_yaml`. This should be done when reworking validation/fsck in general, I think, though.

TobiasKadelka: Cool, makes sense; just wanted to verify it ;)

TobiasKadelka:
LGTM

aqw (approved):
LGTM!

PR #463, opened 2023-10-25
Don't deduplicate lists by casting as set

### LINKED ISSUE #318: Add submodule for API defaults
Currently we use several default values, e.g., `pseudo_keys = ['type', 'make', 'model', 'serial']`, `unset_value = '<unset>'`, `dict_value = '<dict>'`, `list_value = '<list>'`, and possible future additions for `null_value = '<null'>` (interpreting YAML's `null`). With these defaults it makes sense to specify them somewhere all code can reach it.

We can either create a `_defaults.py`, or add these defaults to `onyo/lib/__init__.py` so they can be imported, used consistently, and have need only be changed in one place if the default values are set to be different. The defaults should be hard-coded, because it is the fallback for when no alternatives are present.

If we decide to allow pseudo-keys (among other defaults) to become user-configurable, these values can then be used as defaults in a `Configuration` dataclass which reads the config and sets properties accordingly.

### LINKED ISSUE #440: BUG: Demo repository proves that Operations Log is not sorted
#422 repairs the demo repository.
The commit: "inventory: sort paths in run record alphabetically" sorts the output of run-records alphabetically. If that commit gets dropped, the generated git reference log is not stable --- but it should be.
This proofs that onyo does not run the commands of the demo repository in a reliable order, which is a bug.

The command which showed this bug was the onyo new with a tsv table.

To reproduce:
1. get #422
2. drop commit: "inventory: sort paths in run record alphabetically"
3. run `REPO_ROOT=$PWD pytest -k demo -vv --cov` (multiple times)

The test run at 3. will fail every time differently, because the commands run, and the log created, is different each time.

bpoldrack:
While using sets is obviously fine in some places, don't use it for the sole purpose of deduplicating a list, since this is loosing order.
In addition introduces `lib/consts.py` to hold constants. Added to this PR to avoid circular imports.

Closes #440
Closes #318

### DIFF
--- a/onyo/lib/command_utils.py
+++ b/onyo/lib/command_utils.py
@@ -68,8 +68,8 @@ def sanitize_keys(k: Optional[list[str]],
     Remove duplicates from k while preserving key order and return default
     (pseudo) keys if k is empty
     """
-    seen = set()
-    return [x for x in k if not (x in seen or seen.add(x))] if k else defaults
+    from .utils import deduplicate
+    return deduplicate(k) if k else defaults
 
 
 def set_filters(
--- a/onyo/lib/consts.py
+++ b/onyo/lib/consts.py
@@ -1,0 +1,4 @@
+# TODO: pseudo-key generation mapping
+NEW_PSEUDO_KEYS = ['path']  # TODO: name temporary b/c RF (old idea of pseudo-keys doesn't match)
+# The following keys have functional meaning for onyo and must not be part of user-defined asset content
+RESERVED_KEYS = ['directory', 'is_asset_directory', 'template']
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -18,7 +18,7 @@ from onyo.lib.command_utils import sanitize_keys, set_filters, \
 from onyo.lib.exceptions import OnyoInvalidRepoError, NotAnAssetError
 from onyo.lib.filters import UNSET_VALUE
 from onyo.lib.onyo import OnyoRepo
-from onyo.lib.utils import edit_asset
+from onyo.lib.utils import edit_asset, deduplicate
 
 log: logging.Logger = logging.getLogger('onyo.commands')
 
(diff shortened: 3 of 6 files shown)


onyo/lib/command_utils.py
```diff
@@ -68,8 +68,8 @@ def sanitize_keys(k: Optional[list[str]],
     Remove duplicates from k while preserving key order and return default
     (pseudo) keys if k is empty
     """
-    seen = set()
-    return [x for x in k if not (x in seen or seen.add(x))] if k else defaults
+    from .utils import deduplicate
+    return deduplicate(k) if k else defaults
```
aqw, line 72: I know `get` is unconverted, so these comments are more for later reference:

- the name `sanitize_keys()` is likely a leftover from when this function did more; to me the name implies that it does more
- its location (and name) implies that it's generic, but it has one consumer (`get`), and I struggle to think of what else would need this functionality

This function has one consumer, and is a one-liner. Perhaps it is best folded directly into `get` (when that effort happens).

onyo/lib/consts.py
```diff
@@ -0,0 +1,4 @@
+# TODO: pseudo-key generation mapping
+NEW_PSEUDO_KEYS = ['path']  # TODO: name temporary b/c RF (old idea of pseudo-keys doesn't match)
```
aqw, line 2: I don't think there is a problem with this name.

IIRC, our decision was to no longer use the term "pseudo key" when referring to the keys in the filename. Since they, in fact, only contain content that is in keys.

However, we do intend "pseudo" keys that are queryable, etc. The `path` (aka filename) is among those (as well as some git related stuff, etc).

So I think `PSEUDO_KEYS` (without the `NEW_`) is an appropriate name for this const.

aqw (approved):
Overall LGTM. Neither of my comments are blockers (though the "pseudo" conversation would be nice to have a response on). Up to you whether you'd like to address my comments in this PR or in future ones.

onyo/lib/command_utils.py
bpoldrack: Agree. That should and will be part of addressing `get`, though.

onyo/lib/consts.py
bpoldrack: Agree to this as well. The `NEW_` needs to change. That's why the comment reads "temporary name", because there is a constant `PSEUDO_KEYS` already that needs to go. Just a matter of the refactoring effort. It's still used in `get` (or its helper functions), so renaming is supposed to happen when the old stuff is gone entirely.

aqw: Ahh; I did not check for that. The `NEW_` seemed like an odd "temporary" prefix.

aqw:
Thanks for responding to my comments. Everything LGTM. Merging.

PR #466, opened 2023-10-26
`set`: remove dry-run

bpoldrack:
Removes now superfluous dryrun option for `set`. Note, that dryun still exists for `unset`, because that command still needs a rewrite anyway.


aqw (approved):
Seems straightforward. LGTM.

aqw:
Only 3 tests failed on each. Looks expected (as we move towards green here). Merging!

PR #468, opened 2023-10-26
Fix interactive testing

bpoldrack:
This provides a fixture setting up `onyo.lib.ui`, that is run for every
test. Settings for `yes`, `quiet` and `debug` are read from a new test
marker 'ui' if given. Otherwise set to defaults (`False` for all of
them). The marker takes a dictionary, allowing to enhance this later on
for other future features of the `UI` object (like using the rich
package or redirecting logging or whatever else).

Note, that this is not going to work for CLI tests, since they spawn a
subprocess with its own python context. However, that is obviously
fine, since the CLI calls take explicit flags for that.

### DIFF
--- a/onyo/conftest.py
+++ b/onyo/conftest.py
@@ -156,3 +156,14 @@ class Helpers:
 @pytest.fixture
 def helpers() -> Type[Helpers]:
     return Helpers
+
+
+@pytest.fixture(scope='function', autouse=True)
+def set_ui(request):
+    """Set up onyo.lib.ui according to a dict provided by the 'ui' marker"""
+    from onyo.lib.ui import ui
+    m = request.node.get_closest_marker('ui')
+    if m:
+        ui.set_yes(m.args[0].get('yes', False))
+        ui.set_quiet(m.args[0].get('quiet', False))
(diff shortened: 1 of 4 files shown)


onyo/conftest.py
```diff
@@ -156,3 +156,14 @@ def powerset(iterable: Iterable):
 @pytest.fixture
 def helpers() -> Type[Helpers]:
     return Helpers
+
+
+@pytest.fixture(scope='function', autouse=True)
+def set_ui(request):
+    """Set up onyo.lib.ui according to a dict provided by the 'ui' marker"""
```
TobiasKadelka, line 163: Super-nit: We should talk about punctuation in doc-strings (see chat).

TobiasKadelka:
In general, this looks good to me and I will merge :)

Other than punctuation, I just want to add that I really like the ui fixture. You did just apply it to the tests that needed the fixture to be successful, not everywhere along the tests, which is fine, but since there are a lot of tests that can be removed based on the changes in the code base (discussed multiple times before), I want to note that once somebody cleans them up, we should also apply the new fixture to the other tests were it makes sense.

aqw (approved):
This is cool. Very nice. LGTM. :-)

PR #469, opened 2023-10-26
`commands.py`: normalize style and add doc-strings

TobiasKadelka:
This normalizes the stile of line breaks, whitespace, docstrings, punctunation, etc, for functions in `commands.py`. (I also do the same in `ui.py`, where I noticed some minor things while I used it as a template for the new doc-strings in `commands.py`).

The new texts probably need additional information, this is just an effort to uplift the stile and comments to a minimum quality.

There was a note in `commands.py` stating something should be an `info` instead of `debug`; I changed this, too.

Some parts are to be refactored soon, these I skipped almost completely or provided just a doc-string skeleton that has to be fleshed out later (e.g. `unset`, `--dry-run`).


onyo/lib/commands.py
```diff
@@ -65,19 +82,40 @@ def fsck(repo: OnyoRepo, tests: Optional[list[str]] = None) -> None:
...
             raise OnyoInvalidRepoError(f"'{repo.git.root}' failed fsck test '{key}'")
 
-        ui.log_debug(f"'{key}' succeeded")
+        ui.log(f"'{key}' succeeded")
 
 
-def onyo_cat(repo: OnyoRepo, paths: Iterable[Path]) -> None:
+def onyo_cat(repo: OnyoRepo,
+             paths: Iterable[Path]) -> None:
+    """Run the command `onyo cat` to display the contents of assets.
```
bpoldrack, line 99: That's a strange phrasing. This function is not "running the command" it *is* the command.
The description should state what it does.

```suggestion
    """Print the contents of assets.
```

onyo/lib/commands.py
```diff
@@ -65,19 +82,40 @@ def fsck(repo: OnyoRepo, tests: Optional[list[str]] = None) -> None:
...
+
+    Parameters
+    ----------
+    repo: OnyoRepo
+        The Onyo Repository containing the assets to `cat`.
+
+    paths: one or many `Path`
+        `Path`s to assets for which to display the contents.
```
bpoldrack, line 107: See https://numpydoc.readthedocs.io/en/v1.1.0/format.html#sections -> Parameters

```suggestion
    paths: Path or Iterable of Path
        Path(s) to assets for which to display the contents.
```

onyo/lib/commands.py
```diff
@@ -87,25 +125,44 @@ def onyo_cat(repo: OnyoRepo, paths: Iterable[Path]) -> None:
...
         ui.print(f.read_text(), end='')
     if not assets_valid:
         raise OnyoInvalidRepoError("Invalid assets")
 
 
-def onyo_config(repo: OnyoRepo, config_args: list[str]) -> None:
+def onyo_config(repo: OnyoRepo,
+                config_args: list[str]) -> None:
+    """Run the command `onyo config` to configure the onyo repository-wide
+    config options.
```
bpoldrack, line 141: Apart form the "run ...", this is not only about configuring as in setting a config. You can also get it.

```suggestion
    """Interface the configuration of an onyo repository.
```

onyo/lib/ui.py
```diff
@@ -112,8 +110,7 @@ def set_yes(self,
     def error(self,
               error: str | Exception,
               end: str = os.linesep) -> None:
-        """
-        Print an error message, if the UI is not set to `quiet`.
+        """Print an error message, if the UI is not set to `quiet`.
```
bpoldrack, line 113:
```suggestion
        """Print an error message, if the `UI` is not set to quiet mode.
```

onyo/lib/ui.py
```diff
@@ -148,34 +144,36 @@ def log(self,
...
+        args, kwargs: passed on to builtin `print`.
         """
         if not self.quiet:
             print(*args, **kwargs)
 
     def request_user_response(self,
                               question: str) -> bool:
-        """
-        Opens a dialog for the user and reads an answer from the keyboard.
+        """Opens a dialog for the user and reads an answer from the keyboard.
```
bpoldrack, line 176:
```suggestion
        """Print `question` and read a response from `stdin`.
```

bpoldrack (changes requested):

onyo/lib/commands.py
```diff
@@ -380,22 +491,42 @@ def onyo_new(inventory: Inventory,
...
+
+    message: string
+        An optional string to overwrite Onyo's default commit message.
+
+    Raises
+    ------
+    ValueError
+        If information provided (e.g. via command line call or tsv) contradicts
+        each other or information required to create a new asset is invalid or
+        missing.
```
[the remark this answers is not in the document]
TobiasKadelka: Point taken, though "If an input cannot be acted upon" I find almost a bit to generic.
Thanks, will adjust!

bpoldrack (approved):
Ok let's go with it. Thanks, @TobiasKadelka !

PR #471, opened 2023-10-27
BF: replace Union with new `|`-syntax and `Optional[]`

aqw:
This is purely a syntax uplift thing that just took me a few minutes. If it's not desired, I'm happy to walk away from it.

Now that the Python baseline is 3.11, we can use the new, more terse syntax introduced in Python 3.10.

- https://docs.python.org/3/library/stdtypes.html#types-union
- https://docs.python.org/3/library/typing.html#typing.Optional

The one controversial thing I see is the use of `Optional` (which some feel is poorly named) in place of `something | None`. This can feel odd when defining what a function returns:
```
def yope(Optional[str]) -> Optional[int]
    ...
```

If people bump on that, we can simply avoid `Optional`.

### DIFF
--- a/onyo/lib/assets.py
+++ b/onyo/lib/assets.py
@@ -3,7 +3,7 @@ import logging
 
 import re
 from pathlib import Path
-from typing import Dict, Union, Iterable, Set, Generator, Optional
+from typing import Dict, Generator, Iterable, Optional, Set
 
 from ruamel.yaml import YAML, scanner  # pyre-ignore[21]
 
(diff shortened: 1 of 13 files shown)


onyo/lib/assets.py
```diff
@@ -3,7 +3,7 @@
 
 import re
 from pathlib import Path
-from typing import Dict, Union, Iterable, Set, Generator, Optional
+from typing import Dict, Generator, Iterable, Optional, Set
```
TobiasKadelka, line 6: Thanks for sorting! :D

TobiasKadelka:
I am so happy that the new syntax is finally available; I love it, thanks!

onyo/lib/assets.py
aqw: Haha. Yeah. Most were sorted, and I just did the rest while I was there anyways.

PR #472, opened 2023-10-30
commands: normalize parameter types and names, normalize docstrings 

### LINKED ISSUE #341: API/BUG: normalize arguments and add defaults for "command functions" (e.g. `set`, `unset`)

### LINKED ISSUE #467: Normalize Variables passed into command-functions in `commands.py`
I will provide example functions in commands.py where I noticed this; but the problem is true along all functions.

1. Sometimes we pass `inventory: Inventory`, other times `repo: OnyoRepo` into the command-functions.
The variable should always be of the same type, so that using the API is easy and consistent.

2. Most command functions get one or many `Path`. The variable should be consistently named `paths`, and the type hints should for all functions be the same if possible (type hints should only be different if they are actually working different for a command!)
- onyo_rm() gets a `path: Union[Iterable[Path], Path]`
- onyo_set() gets `paths: Optional[Iterable[Path]]`
- onyo_tree() gets `paths: list[Path]`

TobiasKadelka:
In commands:
- changes in command functions the type from `OnyoRepository` ->`Inventory` (#467) 
  - to make the API more intutive, we pass now always the same variable type (except for unset and fsck, which need to be refactored still)
  - updates variable names and types/typehints in command functions in `commands.py`

This PR also adds keyword "optional" to doc-strings to follow numpydocs.

Close #467
Close #341

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -5,7 +5,6 @@ import sys
 import logging
 from typing import Dict, Iterable, Optional
 from pathlib import Path
-
 from rich.console import Console
 from rich import box
 from rich.table import Table
@@ -95,7 +94,7 @@ def fsck(repo: OnyoRepo,
 
 
 def onyo_cat(repo: OnyoRepo,
-             paths: Iterable[Path]) -> None:
+             paths: list[Path]) -> None:
     """Print the contents of assets.
 
     Parameters
@@ -103,7 +102,7 @@ def onyo_cat(repo: OnyoRepo,
     repo: OnyoRepo
         The Onyo Repository containing the assets to print.
 
-    paths: Path or Iterable of Path
+    paths: Path or list of Path
         Path(s) to assets for which to print the contents.
 
     Raises
@@ -173,7 +172,7 @@ def onyo_config(repo: OnyoRepo,
 
 
 def onyo_edit(inventory: Inventory,
-              asset_paths: Iterable[Path],
+              paths: list[Path],
               message: Optional[str]) -> None:
     """Edit the content of assets.
 
@@ -182,7 +181,7 @@ def onyo_edit(inventory: Inventory,
     inventory: Inventory
         The inventory in which to edit assets.
 
-    asset_paths: Path or Iterable of Path
+    paths: Path or list of Path
         The assets to modify.
 
     message: str, optional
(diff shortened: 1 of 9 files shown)


onyo/lib/commands.py
```diff
@@ -5,6 +5,7 @@
 import logging
 from typing import Dict, Iterable, Optional
 from pathlib import Path
+from collections.abc import Iterable as iter
```
bpoldrack, line 8: I don't understand this one.
1. `typing.Iterable` is an alias for `collection.abc.Iterable`.
2. `as iter` would shadow the builtin `iter`. That's a non-obvious thing to deal with.
3. The usage I see (that `isinstance` test), isn't really what we'd want there. (See comment there)

onyo/lib/commands.py
```diff
@@ -404,15 +406,15 @@ def onyo_mv(inventory: Inventory,
...
+    message: str, optional
         An optional string to overwrite Onyo's default commit message.
 
     Raises
     ------
     ValueError
         If multiple source paths are specified to be renamed.
     """
-    sources = [source] if not isinstance(source, list) else source
+    sources = [source] if not isinstance(source, iter) else list(source)
```
bpoldrack, line 417: I don't think so. `isinstance(source, Iterable)` is `True` for strings and dictionaries for example.

bpoldrack (commented):
I'm fine with the first and third commit - thanks!

I don't agree with the second one (yet).
Why do you want to go back to accept any iterable everywhere and what does it mean to require or not require a list? (Referring to your phrasing in #467: "... instead of list[Path] whenever no specific type is required")

IMO, whether a list is required depends on what a function implements and relies upon. First thing that's coming to mind is: Does order matter somewhere down the line and does the implementation properly account for possibly getting something not ordered (a `set`)?

bpoldrack (approved):
Thanks, @TobiasKadelka

PR #474, opened 2023-11-02
generate_commit_message: remove old code, update usage

### LINKED ISSUE #462: BUG: Default commit subject generation is outdated
`OnyoRepo.generate_commit_message` is not currently valid. While it has recently been adjusted, it still relies on outdated ideas in some cases:
 
 - any consideration of staged paths is invalid
 - filtering out anchor files is superfluous for the same reason. Incoming paths are supposed to point to assets or dirs, not to files that may need to be staged and/or committed.
 - considering `keys` doesn't work that way. What if they were coming in via template and/or tsv? What if a command was called via python interface? This piece only handles CLI strings `<key>=<value>`.
 - it still call `_generate_commit_message_subject`, which in turn falls back to considering an assets `type`. This is flawed, because neither is `type` guaranteed to exist nor is it necessarily retrievable by splitting the file name by `_`.

TobiasKadelka:
The function `generate_commit_message()` now gets a format string defining the commit message subject to be generated, so that this function just has the job of shortening given paths and insert them into the format_string.

The function `generate_commit_message()` (before this were still two functions, together with `_generate_commit_message_subject()`) contained a lot of dead code, e.g. references to modified paths of git.
Also, the function did a bunch of other things before (e.g. command specific work like adding the keys updated with `onyo set` if set was used) other than shortening paths. These are now done before calling it, and just inserted into the format_string.

Close #462

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -221,13 +221,14 @@ def onyo_edit(inventory: Inventory,
             ui.print(line)
         if ui.request_user_response("Save changes? No discards all changes. (y/n) "):
             if not message:
-                operation_paths = [
-                    op.operands[0].get("path")
+                operation_paths = sorted(deduplicate([
+                    op.operands[0].get("path").relative_to(inventory.root)
                     for op in inventory.operations
-                    if op.operator == OPERATIONS_MAPPING['modify_assets']]
+                    if op.operator == OPERATIONS_MAPPING['modify_assets']]))
                 message = inventory.repo.generate_commit_message(
-                    cmd="edit",
-                    modified=operation_paths)
+                    format_string="edit [{len}]: {operation_paths}",
+                    len=len(operation_paths),
+                    operation_paths=operation_paths)
             inventory.commit(message=message)
             return
     ui.print('No assets updated.')
@@ -352,13 +353,14 @@ def onyo_mkdir(inventory: Inventory,
             ui.print(line)
         if ui.request_user_response("Save changes? No discards all changes. (y/n) "):
             if not message:
-                operation_paths = [
-                    op.operands[0]
+                operation_paths = sorted(deduplicate([
+                    op.operands[0].relative_to(inventory.root)
                     for op in inventory.operations
-                    if op.operator == OPERATIONS_MAPPING['new_directories']]
+                    if op.operator == OPERATIONS_MAPPING['new_directories']]))
                 message = inventory.repo.generate_commit_message(
-                    cmd="mkdir",
-                    modified=operation_paths)
+                    format_string="mkdir [{len}]: {operation_paths}",
+                    len=len(operation_paths),
+                    operation_paths=sorted(operation_paths))
             inventory.commit(message=message)
             return
     ui.print('No directories created.')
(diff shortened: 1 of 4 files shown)


onyo/lib/onyo.py
```diff
@@ -203,155 +203,72 @@ def clear_caches(self,
...
-        paths_str = ','.join(sorted(["'{} ({})'".format(x, paths.count(x))
-                                     for x in set(paths)]))
-        msg = f"{msg_dummy}: {paths_str}"
-        if destination:
-            msg = f"{msg} -> '{destination.relative_to(destination.parent)}'"
+
+        shortened_kwargs = {}
+        for elem in kwargs:
+            if isinstance(kwargs[elem], list):
+                paths = [x.name.split('_')[0] for x in kwargs[elem]]
```
bpoldrack, line 261: As pointed out in #462, the file name splitting by `_` as well as assuming the first element of such a split being a meaningful grouping for a shorter depiction is not valid anymore.
Entirely depends on the `onyo.assets.filename` config.

I think, that entire block can be deleted.

onyo/lib/commands.py
```diff
@@ -812,14 +818,15 @@ def onyo_set(inventory: Inventory,
...
                     for op in inventory.operations
-                    if op.operator == OPERATIONS_MAPPING['modify_assets']]
+                    if op.operator == OPERATIONS_MAPPING['modify_assets']])
                 message = inventory.repo.generate_commit_message(
-                    cmd="set",
-                    keys=[str(k) for k in keys.keys()],
-                    modified=operation_paths)
+                    format_string="set [{len}] ({keys}): {operation_paths}",
+                    len=len(operation_paths),
+                    keys=','.join([str(k) for k in keys.keys()]),
```
bpoldrack, line 828: This already is a list of strings.

```suggestion
                    keys=','.join(keys.keys()),
```

bpoldrack: Moreover, this confuses me a bit. What is the special treatment of `list` in `generate_commit_message` for, when here you do the `', '.join(...` outside?

bpoldrack (commented):
Thanks, @TobiasKadelka.
Generally fine, left some comments, though.

One general remark: You have quite a lot of spots where you add single quotes to shortened paths (or names). Why? That's two characters per item for no reason that I can see.

Also: Would be good to mark #462 for closure with this PR ;-)

onyo/lib/onyo.py
TobiasKadelka: You are right, but I wonder if there is not a reasonable replacement for this; e.g. the first element of `onyo.assets.filename` and/or the biggest overlap in the name of all  things the list or something like that.

bpoldrack: Any such approach is making assumptions about what kind of configs would be used. We have no real-world evidence for any of that.
Ultimately, this is collapsing assets into some meaningful grouping. We have no way of assessing what would be meaningful in general, though. So, I think the only reasonable way to do that, is to have a configuration that specifies what to use in this fashion.

onyo/lib/onyo.py
```diff
@@ -203,155 +203,56 @@ def clear_caches(self,
...
+
+        **kwargs
+            Values to insert into the `format_string`. If values are paths, they will be shortened
+            to include as much user readable information as possible.
 
         Returns
         -------
         str
-            A commit message subject to use with `git commit`.
+            A message suitable as a commit message subject for use with `git commit`.
```
bpoldrack, line 229: Not a blocker for the PR, but this adds nothing:

```suggestion
            A message suitable as a commit message subject.
```

onyo/lib/onyo.py
```diff
@@ -203,155 +203,56 @@ def clear_caches(self,
...
-                                cmd: str = "",
-                                keys: Optional[list[str]] = None,
-                                destination: Optional[Path] = None,
+                                format_string: str,
                                 max_length: int = 80,
-                                modified: Optional[Iterable[Path]] = None) -> str:
-        """Generate a commit message subject and body suitable for use with
-        `git commit`.
+                                **kwargs) -> str:
+        """Generate a commit message subject suitable for use with `git commit`.
```
bpoldrack, line 209: See section "Returns":

```suggestion
        """Generate a commit message subject.
```

onyo/lib/onyo.py
```diff
@@ -203,155 +203,56 @@ def clear_caches(self,
...
-    @staticmethod
-    def _generate_commit_message_subject(msg_dummy: str,
-                                         changes: list[Path],
-                                         destination: Optional[Path],
-                                         max_length: int = 80) -> str:
-        """Generates a "commit message subject" usable with `git commit`.
-        The function lists the paths of `changes` and will shorten the resulting
-        string to try to include as much user readable information as possible.
+        The function will shorten paths in the resulting string to try to include as much
+        user readable information as possible.
```
bpoldrack, line 212: The phrasing is strange. If the goal was to maximize information, we wouldn't shorten anything.
The point is `max_length`.
As for "user readable information": What would the function do differently, if "user readability" was not a concern? Can machines extract more information from shorten paths? ;-) 


```suggestion
        The function will shorten paths in the resulting string in order to try to fit into `max_length`.
```

bpoldrack (commented):
Thanks, @TobiasKadelka. Left a few suggestions for phrasing.

PR #476, opened 2023-11-03
Add CONTRIBUTING.md

### LINKED ISSUE #391: add `CONTRIBUTING.md`
We should add a `CONTRIBUTING.md` with the readme section "Code Conventions" to describe that the CLI tests should follow a normalized/standard sheme/order as a minimal requirement, and add some further information.

Here a short overview what I think should be mentioned:
  - call with single/simple path
  - call with multiple paths
  - call from different paths, e.g. subdirectories, etc
  - call with -C in different situations
  - explicit tests with and without all flags
  - explicit tests with combinations of flags
  - explicit tests for different error szanarios

aqw:
Most of the context is pulled (and rewritten a bit) from the README. The overall structure is inspired by the mypy project's `CONTRIBUTING.md` document.

Fix #391

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,34 +83,20 @@ base. For linting and type checking, Onyo uses flake8 and Pyre. The following
 sections describe conventions followed in this project, additionally to the
 [PEP 8 Style Guide](https://peps.python.org/pep-0008/).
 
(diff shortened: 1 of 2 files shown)


bpoldrack (approved):
LGTM

PR #498, opened 2023-11-09
BF: Fix rename operation records 

### LINKED ISSUE #414: BUG: renaming assets results in incorrect path in "Inventory Operations"
When I rename an asset (aka: change the value of a pseudokey) using either `onyo edit` or `onyo set`, it records the incorrect path in the "Inventory Operations" section.It appends the original filename to the destination filename.

```
onyo set -k serial=abc123 --rename -p dp/Alex\ Waite/headphones_sennheiser_SR27.faux
```

Results in the following "Inventory Operations":
```
commit 73fef2a18f504882d65de9c8c79c2ecd5b5b3cbc (HEAD -> master)
Author: Alex Waite <alex@waite.eu>
Date:   Mon Oct 23 13:00:21 2023 +0200
 
    edit: dp/Alex Waite/headphones_sennheiser_SR27.faux
    
    --- Inventory Operations ---
    Modified assets:
            - dp/Alex Waite/headphones_sennheiser_SR27.faux
    Renamed assets:
            - dp/Alex Waite/headphones_sennheiser_SR27.faux -> dp/Alex Waite/headphones_sennheiser_SR27.abc123/headphones_sennheiser_SR27.faux
```

bpoldrack:
Inventory operation records used wrong target path when a rename
operation was recorded. The basename of the target was repeated,
indicating a directory with the same name containing the target.
Reason was that the respective recorders used the same helper function
as move recorders. The latter did, however, expect the dir to move into
as argument.
Fixed by providing a dedicated helper for rename records.

Also fixed another bug, where only the basename was treated as target
path when renaming a directory. This behavior was changed for assets
already, but directories were missing that fix.

Closes #414

### DIFF
--- a/onyo/lib/differs.py
+++ b/onyo/lib/differs.py
@@ -59,7 +59,6 @@ def differ_move_assets(repo: OnyoRepo, operands: tuple) -> Generator[str, None, 
 
 def differ_move_directories(repo: OnyoRepo, operands: tuple) -> Generator[str, None, None]:
     yield from diff_path_change(operands[0], operands[1])
-    # TODO: Fuse w/ differ_move_asset
 
 
 def differ_rename_directories(repo: OnyoRepo, operands: tuple) -> Generator[str, None, None]:
--- a/onyo/lib/recorders.py
+++ b/onyo/lib/recorders.py
@@ -19,7 +19,9 @@ from onyo.lib.onyo import OnyoRepo
 # While a recorder currently only ever returns a single snippet (line) for an operation,
 # the dict assumes a list in order to provide the option to deliver several.
 
-# TODO: Double-check we always report posix paths!
+# TODO: Most functions here account for asset being given as dict or Path.
+#       This is probably superfluous. Re-evaluate, when the `Inventory` class
+#       is reasonably stable.
 
 
 def record_item(repo: OnyoRepo, item: Path | dict) -> str:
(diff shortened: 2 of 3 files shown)


onyo/lib/differs.py
```diff
@@ -59,7 +59,6 @@ def differ_move_assets(repo: OnyoRepo, operands: tuple) -> Generator[str, None,
 
 def differ_move_directories(repo: OnyoRepo, operands: tuple) -> Generator[str, None, None]:
     yield from diff_path_change(operands[0], operands[1])
-    # TODO: Fuse w/ differ_move_asset
```
TobiasKadelka, line 62: Out of curiosity, because you removed this and other TODOs regarding fusing functions: Does this not need to happen anymore or was it done in annother way?

onyo/lib/recorders.py
```diff
@@ -28,14 +30,22 @@ def record_item(repo: OnyoRepo, item: Path | dict) -> str:
...
-    # rather than already containing src' name at the destination. This may not be consistent yet.
+    # This currently expects `dst` to be the dir to move src into,
+    # rather than already containing src' name at the destination.
     src_path = src if isinstance(src, Path) else src['path']
     dst_path = (dst / src_path.name).relative_to(repo.git.root).as_posix()
     src_path = src_path.relative_to(repo.git.root).as_posix()
     return f"- {src_path} -> {dst_path}{linesep}"
 
 
+def record_rename(repo: OnyoRepo, src: Path | dict, dst: Path) -> str:
```
TobiasKadelka, line 41: This seems to be a whole new function, maybe here a docstring, test, and/or the linebreaks for multiple arguments for readability would already make sense (I find it hard to quickly scan especially when having e.g. unions like `Path | dict`), more so because it is meant as a bug fix :)

TobiasKadelka (approved):
Looks good to me, I just left two questions, but I like it and will merge soon :) 👍

onyo/lib/recorders.py
bpoldrack: I'd agree in general.
However, in case of executors, recorders and differs, nothing has proper docstrings yet, they all have largely identical signatures, are very simple code-wise and none of that is intended for any kind of third-party usage.
That's why I didn't bother yet, apart from an introductory comment at the top of each of those three files.

TobiasKadelka: Yes, thanks a lot for the introductory comments! :)

PR #499, opened 2023-11-09
Enforce non-empty values for required keys

### LINKED ISSUE #481: Enforce non-empty values for required keys
This replaces #419, since the other aspect of it is addressed by PR #477.

Keys required for asset names as per `onyo.assets.filename` config, must not be empty. This is not currently enforced directly. Empty values would only fail to pass validation, if they lead to asset names that are not unique.

bpoldrack:
Asset validation for new_assets and modify_assets operations now fail,
when a required key has an empty value.

Closes #481

### DIFF
--- a/onyo/lib/inventory.py
+++ b/onyo/lib/inventory.py
@@ -230,6 +230,8 @@ class Inventory(object):
             raise ValueError(f"{str(path)} is not a valid asset path.")
         if name in self._get_pending_asset_names() + [p.name for p in self.repo.asset_paths]:
             raise ValueError(f"Asset name '{name}' already exists in inventory")
+        if self._required_key_empty(asset):
+            raise ValueError("Required asset keys must not have empty values.")
 
         if asset.get('is_asset_directory', False):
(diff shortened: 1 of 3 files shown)


bpoldrack:
Note, that coverage is to be addressed by #501

onyo/lib/inventory.py
```diff
@@ -547,3 +551,10 @@ def get_faux_serials(self,
                 faux_serials.add(f'faux{serial}')
 
         return faux_serials
+
+    def _required_key_empty(self, asset: dict) -> bool:
+        """Whether `asset` has an empty value for a required key
```
TobiasKadelka, line 556: I would add the dot to the end of the line here (and probably add a linebreak in the definition), though this is of course not a blocker :)

TobiasKadelka (approved):
LGTM, a very important PR :)

onyo/lib/inventory.py
bpoldrack: Done

PR #508, opened 2023-11-16
Fix `config`

bpoldrack:
This patch fixes `onyo_config` to not capture git-config's output only to dump it to stdout/stderr again.
There's also no need to test for an actual modification of `.onyo/config` - just try and let git-commit fail. Lastly, use `OnyoRepo.ONYO_CONFIG` rather than using an independent, hardcoded path for the config file.


aqw (approved):
LGTM

onyo/lib/commands.py
```diff
@@ -142,31 +142,26 @@ def onyo_config(inventory: Inventory,
...
+        # It's a write operation, and we'd want to commit
+        # if there were any changes.
+        try:
+            inventory.repo.git.stage_and_commit(inventory.repo.ONYO_CONFIG,
+                                                'config: modify repository config')
+        except subprocess.CalledProcessError as e:
+            if "no changes added to commit" in e.stdout or "nothing to commit" in e.stdout:
+                ui.print("No changes to commit.")
+                return
+            raise
```
aqw, line 164: Codecov points out that there's no test to cover this condition. That would be ideal, to help make sure this remains functional through future refactorings.

bpoldrack: Hm. Hard to test, though. Something went wrong with committing, that was not caused by "nothing to commit".
I wouldn't know how to get there ;-)
Like: file system issue? Interference by a parallel process in the same repo (killed the repo?), ...
It's really a "do not swallow anything unexpected here"-line. Not sure there's value in mocking and raising something completely made up.
Any suggestions?

PR #510, opened 2023-11-16
unset: refactor command; adjust tests

### LINKED ISSUE #484: Reimplement `unset`
`unset` is not yet adjusted to the current code base and the inventory/operations approach.
This needs to be done and tests be adjusted and reenabled.

TobiasKadelka:
Close #484

### DIFF
--- a/onyo/commands/tests/test_unset.py
+++ b/onyo/commands/tests/test_unset.py
@@ -2,372 +2,323 @@ import subprocess
 from pathlib import Path
 
 from onyo.lib import OnyoRepo
-from onyo.lib.commands import fsck
 import pytest
-from typing import List
-
-files = ['laptop_apple_macbookpro',
-         'lap top_ap ple_mac book pro']
-
-directories = ['.',
-               's p a c e s',
-               'r/e/c/u/r/s/i/v/e',
-               'overlap/one',
-               'overlap/two',
-               'very/very/very/deep',
-               ]
-
-assets: List[str] = [f"{d}/{f}.{i}" for f in files
-                     for i, d in enumerate(directories)]
-
-content_dict = {"one_key": "one_value",
-                "two_key": "two_value",
-                "three_key": "three_value"}
-
-content_str: str = "\n".join([f"{elem}: {content_dict.get(elem)}"
-                              for elem in content_dict]) + "\n"
-
-contents: List[List[str]] = [[x, content_str] for x in assets]
+from typing import Any, Generator
 
 
-pytest.skip("UNSET not currently implemented", allow_module_level=True)
+def convert_contents(
+        raw_assets: list[tuple[str, dict[str, Any]]]) -> Generator:
+    """Convert content dictionary to a plain-text string."""
+    for file, raw_contents in raw_assets:
+        contents = ''
+        for k, v in raw_contents.items():
+            if isinstance(v, str):
+                v = f"'{v}'"
+            elif isinstance(v, bool):
+                v = str(v).lower()
+            contents += f'{k}: {v}\n'
+        yield [file, contents]
 
 
-@pytest.mark.repo_contents(*contents)
-@pytest.mark.parametrize('asset', assets)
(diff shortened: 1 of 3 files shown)


onyo/commands/tests/test_unset.py
```diff
@@ -2,494 +2,383 @@
...
-    key = list(content_dict.keys())[0]
+    """Test that `onyo unset --keys KEY --path ASSET` errors correctly for non-existing assets."""
+    key = "num"
     ret = subprocess.run(['onyo', 'unset', '--keys', key, '--path', *no_assets],
                          capture_output=True, text=True)
 
     # verify output
     assert not ret.stdout
-    assert "The following paths are neither an inventory directory nor an asset:" in ret.stderr
+    assert "The following paths" in ret.stderr
```
bpoldrack, line 206: That's a bit strange. What you assert to be in the message is the exact wording of the unspecific part. What seems relevant to me here, is that the paths (or at least one) of `no_assets` shows up in that message. `The following paths` could be in any kind of unrelated error message.

onyo/commands/tests/test_unset.py
```diff
@@ -2,494 +2,383 @@
...
-@pytest.mark.repo_contents(*contents)
-@pytest.mark.parametrize('asset', assets)
-def test_unset_discard_changes_single_assets(repo: OnyoRepo, asset: str) -> None:
-    """
-    Test that `onyo unset` discards changes for assets successfully.
-    """
-    key = list(content_dict.keys())[0]
+    for asset in repo.get_asset_paths([directory]):
+        assert key not in asset.read_text()
+        assert f"-{key}" in ret.stdout
```
bpoldrack, line 255: Hm. I don't think this line does what you wanted it to do. It will always pass as long as the `-{key}` shows up once in the diff because you always look in `ret.stdout` as a whole not in something you iterate over. If you want to make sure that's the case per file, you need to count (like elsewhere).

onyo/commands/tests/test_unset.py
```diff
@@ -2,494 +2,383 @@
...
+def test_unset_depth(repo: OnyoRepo,
+                     depth: str,
+                     expected: int) -> None:
+    """Test that `onyo unset --depth x` retrieves the expected assets."""
+    key = 'num'
+    ret = subprocess.run(['onyo', '--yes', 'unset', '--keys', 'num', '--depth', depth],
+                         capture_output=True, text=True)
 
+    # Check that for each expected asset there is one line mentioning the removal in the output
+    assert ret.stdout.count(f"{key}") == expected
```
bpoldrack, line 339:
```suggestion
    assert ret.stdout.count(f"-{key}") == expected
```

onyo/lib/commands.py
```diff
@@ -925,89 +925,92 @@ def onyo_tree(inventory: Inventory,
...
+    keys: list
+        The keys that will be unset in assets.
+        If keys do not exist in an asset, a debug message is logged.
+        If keys are specified which appear in asset names an error is raised.
+    match: list of Callable, optional
+      Callables suited for use with builtin `filter`. They are
+      passed an asset dictionary and expected to return a `bool`,
+      where `True` indicates a match. The result of the query
+      consists of all assets that are matched by all callables in
+      this list.
```
bpoldrack, line 949: Copy-Pasta ;-)

```suggestion
      where `True` indicates a match. `keys` will be removed from
      all assets that are matched by all callables in this list.
```

bpoldrack (commented):
Overall looks good. Left a few suggestions, though.

A general remark re tests:
You are introducing a hardcoded `key="num"` everywhere, which means if things change this has to be replaced everywhere.
You have access to the dictionaries defining the assets. Just get a key from these dicts dynamically (same concept as before). If you feel that's complicated with the current state of affairs, then at least make `key` a parameter to the test functions.  The we have the `repo_contents` and asset selection and the `key` all in the decorators. That should make it easier to refactor eventually.

bpoldrack:
Thanks, @TobiasKadelka

Issue #426, opened 2023-10-23
BUG: `onyo mv` does no allow renaming and moving a dir in one call


aqw:
Moving a directory implicitly is ok:
```
❱ onyo mv dp/Alex\ Waite ./
The following will be moved:
/tmp/yope/dp/Alex Waite -> /tmp/yope
Save changes? No discards all changes. (y/n) 
```

Moving a directory explicitly is ok:
```
❱ onyo mv dp/Alex\ Waite Alex\ Waite
The following will be moved:
/tmp/yope/dp/Alex Waite -> /tmp/yope
Save changes? No discards all changes. (y/n) n
Nothing was moved.
```

Renaming a directory is ok (at any depth): 
```
❱ onyo mv dp/Alex\ Waite/ dp/Alex\ Hurr
The following will be moved:
/tmp/yope/dp/Alex Waite -> Alex Hurr
Save changes? No discards all changes. (y/n) n
Nothing was moved.
```

But renaming a folder while moving it is not allowed:
```
❱ onyo mv dp/Alex\ Waite exists/Alex\ Hurr
ERROR: Cannot rename /tmp/yope/dp/Alex Waite -> /tmp/yope/exists/Alex Hurr. Consider moving instead.
```

bpoldrack:
Note: Either we disallow it (as is the case right now), but clarify the message or we settle on fixed order (aqw suggests move, rename) to do both.

PR #514, opened 2023-11-23
BF: print correct error when calling mv to non-existing destination directory

### LINKED ISSUE #420: BUG: `onyo mv`ing an asset into a non-existent dir results in incorrect message
The source file exists and the destination dir doesn't. Yet the error message says that the source is not an inventory directory.

```
(onyo) aqw@hrududu in /tmp/yope on git:master
❱ onyo mv dp/Alex\ Waite/keyboard_wasd_cool.faux does_not_exist/
ERROR: Not an inventory directory: /tmp/yope/dp/Alex Waite/keyboard_wasd_cool.faux
```

### LINKED ISSUE #421: BUG: `onyo mv`ing a dir into a nonexistent dir results in (odd) error
The source directory exists, but none of the destination directories exist. I do think that `onyo` should error here, as the parents do not exist. Onyo does error, but the message printed is incorrect/doesn't make sense.

```
(onyo) aqw@hrududu in /tmp/yope on git:master
❱ onyo mv dp/Alex\ Waite does/not/exist/
ERROR: Cannot rename /tmp/yope/dp/Alex Waite -> /tmp/yope/does/not/exist. Consider moving instead.
```

vs

```
❱ onyo mv dp/Alex\ Waite/ does/not/exist/Alex\ Waite/
ERROR: Destination is not an inventory directory: /tmp/yope/does/not/exist
```

TobiasKadelka:
close #420
close #421

The command did fail (which is the correct behavior), but printed the wrong error messages.

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -537,13 +537,12 @@ def onyo_mv(inventory: Inventory,
         subject = "mv"
         for s in sources:
             move_asset_or_dir(inventory, s, destination)
-    else:
-        # RENAME
+    elif len(sources) == 1 and sources[0].is_dir() and destination.parent.is_dir():
+        # RENAME directory
         subject = "ren"
-        if len(sources) != 1:
-            raise ValueError("Cannot rename multiple sources.")
-        else:
-            inventory.rename_directory(sources[0], destination)
+        inventory.rename_directory(sources[0], destination)
+    else:
+        raise ValueError("Can only move into an existing directory or rename a single directory.")
 
     if inventory.operations_pending():
         ui.print("The following will be {}:".format("moved" if subject == "mv" else "renamed"))
--- a/onyo/commands/tests/test_mv.py
+++ b/onyo/commands/tests/test_mv.py
@@ -36,6 +36,32 @@ def test_mv_interactive_missing_y(repo: OnyoRepo) -> None:
 
 
 @pytest.mark.repo_files('subdir/laptop_apple_macbook.abc123')
+def test_mv_errors_non_existing_destination(repo: OnyoRepo) -> None:
+    """Moving an existing asset or directory into a non-existing destination must error."""
+    # Verify error for asset:
+    ret = subprocess.run(
+        ['onyo', 'mv', 'subdir/laptop_apple_macbook.abc123', 'non/existing/directory'],
+        capture_output=True, text=True)
+    assert not ret.stdout
+    assert "Can only" in ret.stderr
+
+    assert Path('subdir/laptop_apple_macbook.abc123').exists()
+    assert not Path('non/existing/directory/laptop_apple_macbook.abc123').exists()
+    assert repo.git.is_clean_worktree()
+
+    # Verify error for directory:
+    ret = subprocess.run(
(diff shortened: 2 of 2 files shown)


onyo/lib/commands.py
```diff
@@ -537,13 +537,14 @@ def onyo_mv(inventory: Inventory,
         subject = "mv"
         for s in sources:
             move_asset_or_dir(inventory, s, destination)
-    else:
-        # RENAME
+    elif len(sources) == 1 and inventory.repo.is_inventory_dir(sources[0]) and \
+            inventory.repo.is_inventory_dir(Path(*destination.parts[:-1])):
```
bpoldrack, line 541: Two things:

1. `Path(*destination.parts[:-1])` is a little strange. What you mean is `destination.parent`.

2. I don't think this last condition is needed. The first two make sure it's a directory rename already. `rename_directory` should fail on the destination criteria correctly, raising `InvalidInventoryOperation(f"Cannot rename {src} -> {dst}. Consider moving instead.")`. Which is more specific and accurate, than failing with the broader exception you raise from within here.

onyo/lib/commands.py
```diff
@@ -537,13 +537,14 @@ def onyo_mv(inventory: Inventory,
...
+        # RENAME directory
         subject = "ren"
-        if len(sources) != 1:
-            raise ValueError("Cannot rename multiple sources.")
-        else:
-            inventory.rename_directory(sources[0], destination)
+        inventory.rename_directory(sources[0], destination)
+    else:
+        raise ValueError("'onyo mv' can only move into an existing directory",
+                         "rename a single directory.")
```
bpoldrack, line 547: A caller knows where this is coming from and `'onyo mv'` would be command line specific. No need to name it, though. It's either being spit out by a call to `onyo mv` via CLI or raised from `onyo_mv()` when called via python interface. Either way, there's no need to mention its (context-specific) name.

```suggestion
        raise ValueError("Can only move into an existing directory",
                         " or rename a single directory.")
```

onyo/lib/commands.py
TobiasKadelka: 1. Yes definetely right!
2. That is right from perspective of `rename_directory()`, but does in the other cases not adjust the issue this PR tries to solve. But I agree that the check could be simpler and `rename_directory()` addresses some issues more specifically, so `.is_dir()` checks here are sufficient.

bpoldrack: What other case? The two issues are both about targeting a non-existing dir. Whether `destination.parent` exists or not does not distinguish these two somehow. The difference between both issues is the source. That's the second condition - the third seems irrelevant. What's the case where the fact that `destination.parent` exists makes a difference?

TobiasKadelka: The difference when `destination.parent` exists is that it might be possibly a valid rename. Without that check comes the message `Consider moving instead.` which is not more specific, but just wrong -> you can't consider moving, because it is also not allowed.
In 100% of cases when the `destination.parent` does not exist, we know beforehand that a renaming will fail.

bpoldrack: Ok, considering that #426 seems likely to change more of that, I'll resolve here and merge. Let's only reconsider when looking at the whole picture.

Issue #516, opened 2023-12-01
BUG: `onyo mv a/b/c/ a/d/c/` errors "already exists" instead of  `a/d/c/c/`


TobiasKadelka:
AFAIK we allow the same name for multiple directories in an inventory.
However, when moving a directory `c` into another directory also named `c`, it does not lead to a directory `c/c/`, but instead onyo raises an error:

`onyo mv`:
```
❱ onyo mv a/b/c/ a/d/c/ 
ERROR: Target /Users/tkadelka/onyo/a/d/c already exists.
```
Normal `mv`:
```
❱ mv a/b/c/ a/d/c/                                                                         1 !
❱ ls a/d/c/
c
```

This is caused by `onyo_mv()` in `commands.py` going into the ` #MOVE special case`.

Traceback:
```
❱ onyo --debug mv a/b/c/ a/d/c/                                                            1 !
DEBUG:onyo:Onyo repo found at '/Users/tkadelka/onyo'
DEBUG:onyo:Looking up tracked files
DEBUG:onyo:Running 'git ls-files -z'
ERROR: Target /Users/tkadelka/onyo/a/d/c already exists.
DEBUG:onyo:Traceback (most recent call last):
  File "/Users/tkadelka/INM7/onyo/onyo/lib/commands.py", line 493, in move_asset_or_dir
    inventory.move_asset(source, destination)
  File "/Users/tkadelka/INM7/onyo/onyo/lib/inventory.py", line 277, in move_asset
    raise NotAnAssetError(f"No such asset: {src}.")
onyo.lib.exceptions.NotAnAssetError: No such asset: /Users/tkadelka/onyo/a/b/c.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/tkadelka/INM7/onyo/onyo/main.py", line 301, in main
    args.run(args)
  File "/Users/tkadelka/INM7/onyo/onyo/commands/mv.py", line 44, in mv
    onyo_mv(inventory=inventory,
  File "/Users/tkadelka/INM7/onyo/onyo/lib/commands.py", line 536, in onyo_mv
    move_asset_or_dir(inventory, sources[0], destination.parent)
  File "/Users/tkadelka/INM7/onyo/onyo/lib/commands.py", line 495, in move_asset_or_dir
    inventory.move_directory(source, destination)
  File "/Users/tkadelka/INM7/onyo/onyo/lib/inventory.py", line 409, in move_directory
[...]

PR #517, opened 2023-12-12
TST: add tests for `onyo_mv()`

TobiasKadelka:
This adds tests for `commands.onyo_mv()`, and cleans up the CLI tests for `onyo mv`.

There might be some cases missing that should be tested, and during review it should be checked that I did not remove CLI tests that are wanted after all. I also did not test yet for specific commit message contents.

PS: The missing coverage is because of the temporary test for issue #516. I skip the test, because it is not yet working and we have to decide on how to handle this.

### DIFF
--- a/onyo/lib/tests/test_commands_mv.py
+++ b/onyo/lib/tests/test_commands_mv.py
@@ -7,11 +7,12 @@ from ..commands import onyo_mv
 
 
 @pytest.mark.ui({'yes': True})
-def test_onyo_mv_into_self(inventory: Inventory) -> None:
+def test_onyo_mv_errors(inventory: Inventory) -> None:
+    """`onyo_mv` must raise the correct error in different illegal or impossible calls."""
     asset_path = inventory.root / "somewhere" / "nested" / "TYPE_MAKER_MODEL.SERIAL"
     dir_path = inventory.root / 'empty'
 
-    # move into itself:
+    # move directory into itself
     pytest.raises(InvalidInventoryOperation,
                   onyo_mv,
                   inventory,
@@ -19,7 +20,7 @@ def test_onyo_mv_into_self(inventory: Inventory) -> None:
                   destination=dir_path,
                   message="some subject\n\nAnd a body")
 
-    # move asset into non-existing
+    # move asset into non-existing directory
     pytest.raises(ValueError,
                   onyo_mv,
                   inventory,
@@ -27,7 +28,7 @@ def test_onyo_mv_into_self(inventory: Inventory) -> None:
                   destination=dir_path / "doesnotexist",
                   message="some subject\n\nAnd a body")
 
-    # move dir into non-existing
+    # move dir into non-existing directory
     pytest.raises(ValueError,
                   onyo_mv,
                   inventory,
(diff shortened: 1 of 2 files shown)


onyo/lib/tests/test_commands_mv.py
```diff
@@ -69,12 +70,79 @@ def test_onyo_mv_into_self(inventory: Inventory) -> None:
...
+    assert inventory.repo.git.get_hexsha() == old_hexsha
+    # TODO: verifying cleanness of worktree does not work,
+    #       because fixture returns inventory with untracked stuff
+    # assert inventory.repo.git.is_clean_worktree()
+
+
+@pytest.mark.skip(reason="still a BUG: #516")
+@pytest.mark.ui({'yes': True})
+@pytest.mark.repo_dirs("a/b/c", "a/d/c")
+def test_onyo_mv_possible_error(inventory: Inventory) -> None:
```
bpoldrack, line 112: I have trouble understanding the name of this test and how it relates to what' supposed to be tested here.

Also: An alternative to skipping would be to achieve the outcome with the current implementation. We have the issue and addressing it would simply involve a change in the test. So, if you make the target explicit (`a/d/c/c`) the test should pass and reflects the implemented logic.

onyo/lib/tests/test_commands_mv.py
```diff
@@ -69,12 +70,79 @@ def test_onyo_mv_into_self(inventory: Inventory) -> None:
...
+    """Allow to move a directory into another one with the same name."""
+    source_path = inventory.root / "a" / "b" / "c"
+    destination_path = inventory.root / "a" / "d" / "c"
+    old_hexsha = inventory.repo.git.get_hexsha()
+
+    # move an asset and a dir to the same destination
```
bpoldrack, line 118: Is this comment just copy-pasted from elsewhere? What's the asset it talks about? I only see one source and it's a dir.

onyo/lib/tests/test_commands_mv.py
```diff
@@ -69,12 +70,79 @@ def test_onyo_mv_into_self(inventory: Inventory) -> None:
...
+
+    # source
+    assert not source_path.exists()
+    assert (source_path / OnyoRepo.ANCHOR_FILE_NAME) not in inventory.repo.git.files
+    # directory was moved
+    assert (destination_path / source_path.name).is_dir()
+    assert (destination_path / source_path.name / OnyoRepo.ANCHOR_FILE_NAME).is_file()
+    assert inventory.repo.is_inventory_dir(destination_path / source_path.name)
+    assert (destination_path / source_path.name / OnyoRepo.ANCHOR_FILE_NAME) in inventory.repo.git.files
+    # git is correct
```
bpoldrack, line 132: Hm. This comment says nothing really. What's a correct git? ;-)
As far as I'm concerned, no comment needed, but if you want one here, then it should be more clear.
"exactly one commit added" or something like this.

bpoldrack (commented):
- Some minor remarks on code comments and naming stuff.
- The skipped test could test current way of doing this instead of waiting for the issue to be resolved.

Other than that, I'm fine with this, @TobiasKadelka  :)

onyo/lib/tests/test_commands_mv.py
TobiasKadelka: Hm, I can see that, but prefer the current way. The "explicit target" test could be added additionally, but I find it more helpful to already have a test that I can use to verify when I later try to fix the underlying bug.

That said -- the name of the test is a temporary one, and I'll adjust it. I just forgot to change it from before -- oops.

TobiasKadelka:
The PR is updated, thanks for the feedback!

Issue #495, opened 2023-11-08
Tests + docstring for onyo_tree


bpoldrack:
Properly document `onyo/lib/commands.py:onyo_tree` and provide tests in `onyo/lib/tests/test_command_tree.py`.
Based on that reworked docstring, provide a respective command line version in `onyo/commands/tree.py`.

aqw:
This was completed long ago as part of moving the functionality from system `tree` to a Python implementation. 

The doc strings are good, and the tests are sufficient.

Issue #493, opened 2023-11-08
Tests + docstring for onyo_cat


bpoldrack:
Properly document `onyo/lib/commands.py:onyo_cat` and provide tests in `onyo/lib/tests/test_command_cat.py`.
Based on that reworked docstring, provide a respective command line version in `onyo/commands/cat.py`.

Issue #491, opened 2023-11-08
Tests + docstring for `onyo_rm()`


bpoldrack:
Properly document `onyo/lib/commands.py:onyo_rm` and provide tests in `onyo/lib/tests/test_command_rm.py`.
Based on that reworked docstring, provide a respective command line version in `onyo/commands/rm.py`.

Issue #490, opened 2023-11-08
Tests + docstring for `onyo_set()`


bpoldrack:
Properly document `onyo/lib/commands.py:onyo_set` and provide tests in `onyo/lib/tests/test_command_set.py`.
Based on that reworked docstring, provide a respective command line version in `onyo/commands/set.py`.

PR #521, opened 2024-01-12
Docs and Tests

### LINKED ISSUE #522: `onyo_tree()` docstring lies about its default value
The CLI call `onyo tree` (without specifying paths) does automatically use the CWD, but the python implementation does require an argument for `paths`, otherwise it errors with `TypeError: onyo_tree() missing 1 required positional argument: 'paths'`.

However, the python implementation states:
```
    paths: list of Path
        The paths to directories for which to print the directory tree.
        If no path is specified, prints the directory tree for CWD.
```

This is additionally problematic, because it is not obvious to me that the python function would be called from the "correct" directory, so IMO one cannot just add `paths: list[Path] = [Path.cwd()]` as a default to the function to make it work.

Since the function is called as `onyo_tree(inventory)` I feel like the most reasonable path to print by default for the python implementation would be the root of the inventory. However, I do not like the idea that the default behavior when not specifying a path is different for the CLI and the python function.

Alternatively, one might also just except the current behavior of the function, and delete the un-true comment about the CWD.

### LINKED ISSUE #523: `onyo_cat()` excepts empty list of paths, but `onyo cat` errors without paths
`onyo_cat(inventory, paths=[])` does not raise an error when an empty list of paths is specified, but `onyo_cat(inventory)` errors because the paths variable is required, and `onyo cat` errors with "onyo cat: error: the following arguments are required: ASSET".

We should clarify if we require that at least one path is cat-able, or if we allow a cat without paths, i.e. all three szenarios above should lead to the same result.

I think all three szenarios should error, but if somebody wants to argue that empty paths are valid, I am open to hear about it.

### LINKED ISSUE #524: `onyo_mkdir()` excepts empty list of paths, but `onyo mkdir` errors without paths
This issue is for `onyo_mkdir()`, and discusses the same underlying problem as in #523 for cat.

IMO, the python implementation and the CLI command should behave similar whenever reasonably possible, but currently `onyo mkdir` errors without paths:
```
❱ onyo mkdir                                                                               1 !
usage: onyo mkdir [-h] [-m MESSAGE] DIR [DIR ...]
onyo mkdir: error: the following arguments are required: DIR
```

And `onyo_mkdir(..., dirs=[], ...)` does not error:
```
    onyo_mkdir(inventory,
               dirs=[],
               message="some subject\n\nAnd a body")
```

I plan on resolving this with adding a check `if not dirs` which raises an error, again, in parallel to `onyo_cat()`.

### LINKED ISSUE #526: `onyo_set()` accepts empty list of keys, but `onyo set` errors without keys
Currently one of the underlying called functions raises an `AttributeError`, but `onyo_set()` should check before if keys are given and raise a specific error if not.

This is in parallel to #523 and #524.

TobiasKadelka:
Works towards:
- #487
- #491
- #495
  - closes #522 -> here I chose as a solution to display the inventory.root if the python function `onyo_tree(inventory)` is called without paths. This needs probably feedback/discussion!
  - I am not verifying the correctness of the output of `onyo_tree()`
- #493
  - closes #523 -> at least one valid asset path is required by `onyo_cat()`
- #488
  - closes #524 -> at least one valid directory path is required by `onyo_mkdir()`
- does not finish but works towards #490 
  -  closes #526
- Updates tests and doc-strings for `onyo_new()`, but will not close the issue
- Tests for `onyo init` are contained in `onyo/lib/tests/test_onyo.py` and were updated, too

### DIFF
--- a/onyo/commands/cat.py
+++ b/onyo/commands/cat.py
@@ -21,8 +21,10 @@ args_cat = {
 
 
 def cat(args: argparse.Namespace) -> None:
-    """
-    Print the contents of ``ASSET``\\(s) to the terminal without parsing.
+    """Print the contents of ``ASSET`` file(s) to the terminal.
+
+    At least one valid asset path is required. Assets can be given multiple times.
+    If any path specified is invalid, no contents are printed and an error is raised.
     """
     paths = [Path(p).resolve() for p in args.asset]
 
--- a/onyo/commands/mv.py
+++ b/onyo/commands/mv.py
@@ -34,7 +34,17 @@ def mv(args: argparse.Namespace) -> None:
     rename a ``SOURCE`` directory to ``DEST``.
 
     Files cannot be renamed using ``onyo mv``, since their names are generated from their contents.
-    To rename a file, use ``onyo set``.
+    To rename a file, use ``onyo set``, or use ``onyo edit`` and change the keys used for the
+    asset's name.
+    To rename a directory, call ``onyo mv`` with a single ``SOURCE`` to rename, and a different and
+    non-existing ``DEST`` name in the same directory.
+
+    Otherwise, when called on one or multiple assets or directories, the command will move
+    ``SOURCE``\\(s) into ``DEST``.
+
+    A list of all files and directories to modify will be presented, and the user prompted for
+    confirmation.
+
     """
     inventory = Inventory(repo=OnyoRepo(Path.cwd(), find_root=True))
 
--- a/onyo/commands/new.py
+++ b/onyo/commands/new.py
@@ -36,7 +36,10 @@ args_new = {
         nargs='+',
         help=(
             'Key-value pairs to set in the new asset(s). Multiple pairs can be '
-            'specified (e.g. key=value key2=value2)')),
+            'specified (e.g. key=value key2=value2). All fields that are part of '
+            'asset filenames (defined in .onyo/config under `onyo.assets.filename`) '
+            'are required. If the value `faux` is assigned to the key `serial`, '
+            'a random, repository-unique string will be filled in instead.')),
 
     'path': dict(
         args=('-p', '--path'),
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -98,18 +98,22 @@ def onyo_cat(inventory: Inventory,
              paths: list[Path]) -> None:
     """Print the contents of assets.
 
+    At least one valid asset path is required.
+    The same paths can be given multiple times.
+    If any path specified is invalid, no contents are printed and an error is raised.
+
     Parameters
     ----------
     inventory: Inventory
         The inventory containing the assets to print.
-
-    paths: Path or list of Path
+    paths: list of Path
         Path(s) to assets for which to print the contents.
 
     Raises
     ------
     ValueError
-        If paths point to a location which is not an asset.
+        If paths point to a location which is not an asset, or `paths`
+        is empty.
 
     OnyoInvalidRepoError
         If paths are not valid assets, e.g. because their content is not valid
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -846,6 +846,8 @@ def onyo_set(inventory: Inventory,
         renaming an asset, while `rename` is not true.
     """
     paths = paths or []
+    if not keys:
+        raise ValueError("At least one key-value pair must be specified.")
 
     if not rename and any(k in inventory.repo.get_asset_name_keys() for k in keys.keys()):
         raise ValueError("Can't change asset name keys without --rename.")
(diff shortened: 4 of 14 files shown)


onyo/lib/commands.py
```diff
@@ -116,6 +116,8 @@ def onyo_cat(inventory: Inventory,
         YAML format.
     """
     from .assets import validate_yaml
+    if not paths:
+        raise ValueError("At least one asset path must be specified.")
```
bpoldrack, line 124: Agree on the addition, but maybe add it in the docstring for when a `ValueError` is raised?

onyo/lib/commands.py
```diff
@@ -902,14 +902,16 @@ def onyo_tree(inventory: Inventory,
 
     paths: list of Path
         The paths to directories for which to print the directory tree.
-        If no path is specified, prints the directory tree for CWD.
+        If no path is specified, `onyo_tree(inventory)` prints the
+        directory tree for the root of the inventory.
```
bpoldrack, line 928: Agree.

onyo/commands/cat.py
```diff
@@ -21,8 +21,10 @@
 
 
 def cat(args: argparse.Namespace) -> None:
-    """
-    Print the contents of ``ASSET``\\(s) to the terminal without parsing.
+    """Print the contents of ``ASSET``\\(s) to the terminal without parsing.
```
bpoldrack, line 24:
```suggestion
    """Print the contents of ``ASSET`` file(s) to the terminal.
```

I don't think "without parsing" adds anything useful. And the notion of an asset file seems a bit clearer to me than that of an asset, since in other context that would be the dictionary. Here it's all about the file as is.

onyo/commands/mv.py
```diff
@@ -34,7 +34,17 @@ def mv(args: argparse.Namespace) -> None:
...
 
     Files cannot be renamed using ``onyo mv``, since their names are generated from their contents.
-    To rename a file, use ``onyo set``.
+    To rename a file, use ``onyo set``, or use ``onyo edit`` and change the keys used for the
+    asset's name.
+    To rename a directory, call ``onyo mv`` with a single ``SOURCE`` to rename, and a different and
+    non-existing ``DEST`` name in the same directory.
+
+    Otherwise, when called on one or multiple assets or directories, the command will move
+    ``SOURCE``\\(s) to ``DEST``.
```
bpoldrack, line 43:
```suggestion
    ``SOURCE``\\(s) into ``DEST``.
```

onyo/commands/new.py
```diff
@@ -36,7 +36,10 @@
         nargs='+',
         help=(
             'Key-value pairs to set in the new asset(s). Multiple pairs can be '
-            'specified (e.g. key=value key2=value2)')),
+            'specified (e.g. key=value key2=value2). All fields that are part of '
+            'asset filenames (defined in .onyo/config under `onyo.assets.filename`) '
+            'are required. If `--key serial=faux` a random string will be generated '
+            'and attached to the serial field.')),
```
bpoldrack, line 42:
```suggestion
            'are required. If the value `faux` is assigned to the key `serial`, '
            'a random, repository-unique string will be filled in instead.')),
```

onyo/lib/tests/test_commands_cat.py
```diff
@@ -0,0 +1,118 @@
...
+
+@pytest.mark.ui({'yes': True})
+def test_onyo_cat_single(inventory: Inventory) -> None:
+    """`onyo_cat()` a single valid asset."""
+    asset_path = inventory.root / "somewhere" / "nested" / "TYPE_MAKER_MODEL.SERIAL"
+
+    # cat single asset
+    onyo_cat(inventory,
+             paths=[asset_path])
+
```
bpoldrack, line 86: Here you only test that cat does not error, but not what it does.
One could use pytest's `capsys` fixture to check the file's content is put on stdout.

Same for the variants below.

bpoldrack (approved):
Thanks much, @TobiasKadelka!

Some minor comment you may wnat to address, but overall - I'm fine with merging.
It's an important step either way :)

bpoldrack:
One nitpick: If a commit addresses an issue I'd prefer to have the commit link the issue irrespective of whether the PR itself does so. Because for restrospective investigation, code tends to be the starting point.

bpoldrack:
One more, though. I didn't pay attention to the PR description initially.

If you say "works towards" and then "close ISSUE", that's somewhat contradictory. With respect to the doc+test issues per command, I'd say this is not closing. At least the test part needs another iteration (pending the stuff I'm on). Hence, I'll remove the keyword. The bugfixes are closing of course.

onyo/lib/commands.py
TobiasKadelka: Good point, I added it here and everywhere else were I added such a check!

PR #529, opened 2024-01-30
TST/DOC: add/update tests and doc-strings

### LINKED ISSUE #527: BUG: `onyo set` and `onyo_set()` do not fail for illegal key `path`, but do not change asset
When calling `onyo set` on valid paths (the root dir in example) but with a key called `path` onyo should error, (i.e. `onyo new` does this, too). But instead `onyo set` does not fail, it just does not do anything with the key `path`.

IMO all commands should behave the same way -- in this case error.

Example calls:

1. When called just with `-k path=a` nothing is updated:
```
❱ onyo set -k path=a -p .                                                                  1 !
No assets updated.
```

2. When called with other keys `path` is ignored, but other key-values are set:
```
❱ onyo set -k path=a key=value -p .
The following assets will be changed:
--- /Users/tkadelka/onyo/a_a_a.fauxKJG3b9
+++ /Users/tkadelka/onyo/a_a_a.fauxKJG3b9
@@ -2,3 +2,4 @@
 make: a
 model: a
 serial: fauxKJG3b9
+key: value
```

3. `onyo new`, however, does fail correctly when called with a key `path=a`:
```
(valid) tkadelka@bnbmac73 in ~/onyo on git:master
❱ onyo new -k type=a make=a model=a serial=faux path=a --path ./
ERROR: Pseudo key 'path' must not be specified.
```

### LINKED ISSUE #528: BUG: `onyo unset` and `onyo_unset()` do not check for `path` in keys to unset
The call of `onyo unset --keys path` errors, which is technically correct, but IMO it does not error nicely (e.g. it gives no explainatory message), and probably just "accidently": it states a `KeyError` (named properly enough) was raised because "path" is not in the asset file, but gives no further context.

For context: Normally, non-existing keys do not raise a `KeyError` in `onyo unset`, they just trigger a debug message.
Other commands (e.g. edit) state explicitly that "path" is not allowed.

There are IMO two ways to solve this and normalize the behavior:
1. there could be an explicite error when `onyo unset --keys path` (like for other commands when called with "path")
2. a debug message could be printed when called with "path" (like for other non-existing keys for `onyo unset`)

-------------

Here the behavior that currently seems incorrect to me:

```
(valid) tkadelka in ~/onyo on git:master
❱ onyo unset --keys path                                                                   1 !
ERROR: 'path'

(valid) tkadelka in ~/onyo on git:master
❱ onyo --debug unset --keys path a_a_a.fauxKJG3b9                                          1 !
DEBUG:onyo:Onyo repo found at '/Users/tkadelka/onyo'
DEBUG:onyo:Running 'git config --get onyo.assets.filename'
DEBUG:onyo:git config missed 'onyo.assets.filename'
DEBUG:onyo:Running 'git config --file /Users/tkadelka/onyo/.onyo/config --get onyo.assets.filename'
DEBUG:onyo:config 'onyo.assets.filename' acquired from /Users/tkadelka/onyo/.onyo/config: '{type}_{make}_{model}.{serial}'
DEBUG:onyo:Running 'git config --get onyo.assets.filename'
DEBUG:onyo:git config missed 'onyo.assets.filename'
DEBUG:onyo:Running 'git config --file /Users/tkadelka/onyo/.onyo/config --get onyo.assets.filename'
DEBUG:onyo:config 'onyo.assets.filename' acquired from /Users/tkadelka/onyo/.onyo/config: '{type}_{make}_{model}.{serial}'
DEBUG:onyo:Looking up tracked files
DEBUG:onyo:Running 'git ls-files -z'
DEBUG:onyo:a_a_a.fauxKJG3b9 not in {'type': 'a', 'make': 'qwerty', 'model': 'as df', 'serial': 'fauxmnU3e4', 'key': 'value', 'path': PosixPath('/Users/tkadelka/onyo/a/a_qwerty_as df.fauxmnU3e4')}
ERROR: 'path'
  File "/Users/tkadelka/INM7/onyo/onyo/main.py", line 301, in main
    args.run(args)
  File "/Users/tkadelka/INM7/onyo/onyo/commands/unset.py", line 69, in unset
    unset_cmd(inventory,
  File "/Users/tkadelka/INM7/onyo/onyo/lib/commands.py", line 1017, in onyo_unset
    new_content.pop(k)
  File "/Users/tkadelka/env/valid/lib/python3.12/site-packages/ruamel/yaml/comments.py", line 924, in pop
    result = self[key]
             ~~~~^^^^^
  File "/Users/tkadelka/env/valid/lib/python3.12/site-packages/ruamel/yaml/comments.py", line 851, in __getitem__
    return ordereddict.__getitem__(self, key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'path'
```

Trying to unset name fields errors nicely:
```
(valid) tkadelka in ~/onyo on git:master
❱ onyo unset --keys type                                                                 130 !
ERROR: Can't unset asset name keys.

```

### LINKED ISSUE #530: Define `onyo_set()`/`onyo set` behavior for empty keys/values
There are three szenarios in which onyo set can be called with missing keys/values, and they do not behave consistent:

1. both missing, just an `-k =` or `-k ""=` or `-k =""` is passed to satisfy the -k requirement.
```
❱ onyo set -k =
--- /Users/tkadelka/onyo/a_a_a.fauxKJG3b9
+++ /Users/tkadelka/onyo/a_a_a.fauxKJG3b9
@@ -3,3 +3,4 @@
 model: a
 serial: fauxKJG3b9
 key: ''
+'': ''
```

2. no key, but a value is specified:
```
❱ onyo set -k ""=value
zsh: value not found
```

3. just a key, no value, is specified:
```
❱ onyo set -k field=""
--- /Users/tkadelka/onyo/a_a_a.fauxKJG3b9
+++ /Users/tkadelka/onyo/a_a_a.fauxKJG3b9
@@ -3,3 +3,4 @@
 model: a
 serial: fauxKJG3b9
 key: ''
+field: ''
```

I think the szenarios 1.+2. should be just not allowed because they miss a key. I do not see any practical value not having a key.
The third szenario is the only one that might have a reasonable use-case, though even there I think we should discuss if this is wanted behavior.

TobiasKadelka:
Changes:
1. adds/updates tests/docs for the python implementations of commands
2. at least one key to unset is required
3. `match` is now actually optional for `onyo_unset()`
4. Check for illegal `path` in keys for  `onyo_set()` and `onyo_unset()`
  - close #527 
  - close #528 
5. check for illegal none values in `keys`for `onyo_set()` and `onyo_unset()`
  - close #530
6. Normalize default path for python functions of commands
  - some (e.g. `onyo_get()` and `onyo_tree()`) already used `inventory.root` when called like `onyo_<cmd>(inventory, ...)`, while `onyo_set()` and `onyo_unset()` were still using the CWD like the command line implementations do. I normalized these and added tests for this behavior
7. correct default values and optional parameters for `onyo_get()`
  - doc-strings of `get` stated this already, it was just not functionally allowed
8. `onyo_get()`requires sorting to be ascending/descending, illegal values error now instead of silently using the default

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -927,11 +927,12 @@ def onyo_set(inventory: Inventory,
     paths = paths or [inventory.root]
     if not keys:
         raise ValueError("At least one key-value pair must be specified.")
-
+    if any(not k or not k.strip() for k in keys.keys()):
+        raise ValueError("Keys are not allowed to be empty or None-values.")
     if not rename and any(k in inventory.repo.get_asset_name_keys() for k in keys.keys()):
         raise ValueError("Can't change asset name keys without --rename.")
     if any(k in RESERVED_KEYS + PSEUDO_KEYS for k in keys.keys()):
-        raise ValueError(f"Can't set reserved keys ({', '.join(RESERVED_KEYS + PSEUDO_KEYS)}).")
+        raise ValueError(f"Can't set reserved or pseudo keys ({', '.join(RESERVED_KEYS + PSEUDO_KEYS)}).")
 
     non_inventory_paths = [str(p)
                            for p in paths  # pyre-ignore[16]  `paths` not Optional anymore here
@@ -1064,7 +1065,7 @@ def onyo_unset(inventory: Inventory,
     if any(k in inventory.repo.get_asset_name_keys() for k in keys):
         raise ValueError("Can't unset asset name keys.")
     if any(k in RESERVED_KEYS + PSEUDO_KEYS for k in keys):
-        raise ValueError(f"Can't unset reserved keys ({', '.join(RESERVED_KEYS + PSEUDO_KEYS)}).")
+        raise ValueError(f"Can't unset reserved or pseudo keys ({', '.join(RESERVED_KEYS + PSEUDO_KEYS)}).")
 
     asset_paths_to_unset = inventory.get_assets_by_query(paths=paths,
                                                          depth=depth,
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -1037,6 +1037,8 @@ def onyo_unset(inventory: Inventory,
         An optional string to overwrite Onyo's default commit message.
     """
     paths = paths or []
+    if not keys:
+        raise ValueError("At least one key must be specified.")
 
     non_inventory_paths = [str(p) for p in paths
                            if not inventory.repo.is_asset_path(p) and
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -5,7 +5,7 @@ import logging
 import subprocess
 import sys
 from pathlib import Path
-from typing import Callable, Dict, Optional, TypeVar, ParamSpec
+from typing import Callable, Dict, Literal, Optional, ParamSpec, TypeVar
 from functools import wraps
 
 from rich import box
@@ -386,13 +386,13 @@ def onyo_edit(inventory: Inventory,
 
 @raise_on_inventory_state
 def onyo_get(inventory: Inventory,
-             paths: Optional[list[Path]],
-             depth: int,
-             machine_readable: bool,
-             match: Optional[list[Callable[[dict], bool]]],
-             keys: Optional[list[str]],
-             sort: str = 'ascending') -> list[dict]:
-    """Query the repository for assets.
+             paths: Optional[list[Path]] = None,
+             depth: int = 0,
+             machine_readable: bool = False,
+             match: Optional[list[Callable[[dict], bool]]] = None,
+             keys: Optional[list[str]] = None,
+             sort: Literal['ascending', 'descending'] = 'ascending') -> list[dict]:
+    """Query the repository for information about assets.
 
     Parameters
     ----------
@@ -400,20 +400,22 @@ def onyo_get(inventory: Inventory,
       The inventory to query.
     paths: list of Path, optional
       Limits the query to assets underneath these paths.
+      Paths can be assets and directories.
+      If no paths are specified, the inventory root is used as default.
     depth: int
       Number of levels to descent into. Must be greater or equal 0.
       If 0, descend recursively without limit.
-    machine_readable: bool
+    machine_readable: bool, optional
       Whether to print the matching assets as TAB-separated lines,
       where the columns correspond to the `keys`. If `False`,
       print a table meant for human consumption.
-    match: list of Callable
+    match: list of Callable, optional
       Callables suited for use with builtin `filter`. They are
       passed an asset dictionary and expected to return a `bool`,
       where `True` indicates a match. The result of the query
       consists of all assets that are matched by all callables in
       this list.
-    keys: list of str
+    keys: list of str, optional
       Defines what key-value pairs of an asset a result is composed of.
       If no `keys` are given the keys then the asset name keys are
       used. The 'path' pseudo-key is always appended.
@@ -421,6 +423,7 @@ def onyo_get(inventory: Inventory,
     sort: str
       How to sort the results by `keys`. Possible values are
       'ascending' and 'descending'. Default: 'ascending'.
+      If other values are specified an error is raised.
 
     Raises
     ------
@@ -445,6 +448,10 @@ def onyo_get(inventory: Inventory,
     if invalid_paths:
         err_str = '\n'.join([str(x) for x in invalid_paths])
         raise ValueError(f"The following paths are not part of the inventory:\n{err_str}")
+
+    allowed_sorting = ['ascending', 'descending']
+    if sort not in allowed_sorting:
+        raise ValueError(f"Allowed sorting modes: {', '.join(allowed_sorting)}")
 
     selected_keys = selected_keys or inventory.repo.get_asset_name_keys()
     results = inventory.get_assets_by_query(paths=paths,
@@ -1039,8 +1046,13 @@ def onyo_unset(inventory: Inventory,
     paths = paths or []
     if not keys:
         raise ValueError("At least one key must be specified.")
+<<<<<<< Updated upstream
 
     non_inventory_paths = [str(p) for p in paths
+=======
+    
+    non_inventory_paths = [str(p) for p in paths  # pyre-ignore[16]  `paths` not Optional anymore
+>>>>>>> Stashed changes
                            if not inventory.repo.is_asset_path(p) and
                            not inventory.repo.is_inventory_dir(p)]
 
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -923,8 +923,8 @@ def onyo_set(inventory: Inventory,
 
     if not rename and any(k in inventory.repo.get_asset_name_keys() for k in keys.keys()):
         raise ValueError("Can't change asset name keys without --rename.")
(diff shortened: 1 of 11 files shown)


onyo/lib/commands.py
```diff
@@ -340,41 +340,44 @@ def onyo_edit(inventory: Inventory,
...
 
 def onyo_get(inventory: Inventory,
-             paths: Optional[list[Path]],
-             depth: int,
-             machine_readable: bool,
-             match: Optional[list[Callable[[dict], bool]]],
-             keys: Optional[list[str]],
+             paths: Optional[list[Path]] = None,
+             depth: int = 0,
+             machine_readable: Optional[bool] = False,
```
bpoldrack, line 345: This seems to be a mistaken notion of `Optional`.
It's a bool and because it comes with a default a caller doesn't need to pass anything. But that's not what `Optional` declares. `Optional` means it could be `None`. (It's equivalent to `bool | None`) 


Note, however, that NumPy's notion of 'optional' in the docstring is different and correct as is, I think. This is just about type annotation.

onyo/lib/commands.py
```diff
@@ -400,6 +403,10 @@ def onyo_get(inventory: Inventory,
         err_str = '\n'.join([str(x) for x in invalid_paths])
         raise ValueError(f"The following paths are not part of the inventory:\n{err_str}")
 
+    allowed_sorting = ['ascending', 'descending']
+    if sort not in allowed_sorting:
+        raise ValueError(f"Allowed sorting modes: {', '.join(allowed_sorting)}")
+
```
bpoldrack, line 409: Seeing this, reminds me: We should probably include that kind of thing in type annotations:

So, `sort: Literal['ascending', 'descending']` instead of `sort: str`.

onyo/lib/commands.py
```diff
@@ -866,17 +873,18 @@ def onyo_set(inventory: Inventory,
         If a given path is invalid or changes are made that would result in
         renaming an asset, while `rename` is not true, or if `keys` is empty.
     """
-    paths = paths or []
+    paths = paths or [inventory.root]
     if not keys:
         raise ValueError("At least one key-value pair must be specified.")
-
+    if not all(keys.keys()):
+        raise ValueError("Keys are not allowed to be empty or None-values.")
```
bpoldrack, line 880: Two things:

1. This is missing strings consisting of white spaces (instead of being technically empty).
2. I think this is the wrong place. To guard against that consistently, the underlying inventory operations should fail instead - no matter where they are called from. So, `Inventory.modify_asset` in this case. But also `add_asset`. Same for the check against RESERVED_KEYS and PSEUDO_KEYS. If you think of that as a refactoring that's out of scope for this PR, that's fine - just stumbled upon it on this occasion. But then we should turn it into an issue.

onyo/lib/commands.py
```diff
@@ -866,17 +873,18 @@ def onyo_set(inventory: Inventory,
...
     if not rename and any(k in inventory.repo.get_asset_name_keys() for k in keys.keys()):
         raise ValueError("Can't change asset name keys without --rename.")
-    if any(k in RESERVED_KEYS for k in keys.keys()):
-        raise ValueError(f"Can't set reserved keys ({', '.join(RESERVED_KEYS)}).")
+    if any(k in RESERVED_KEYS + PSEUDO_KEYS for k in keys.keys()):
+        raise ValueError(f"Can't set reserved keys ({', '.join(RESERVED_KEYS + PSEUDO_KEYS)}).")
```
bpoldrack, line 884: 1. See above.
2. Messaging should adjust accordingly. (It's still talking about reserved keys).

onyo/lib/commands.py
```diff
@@ -963,29 +971,39 @@ def onyo_unset(inventory: Inventory,
...
+    ------
+    ValueError
+        If paths are invalid, or `keys` are empty or invalid.
+
+    """
+    paths = paths or [inventory.root]
+    if not keys:
+        raise ValueError("At least one key must be specified.")
+    if not all(keys):
+        raise ValueError("Keys are not allowed to be empty or None-values.")
```
bpoldrack, line 1005: I think this should not be necessary in case of unset. If there's no empty key in a matching asset, then we should just end up failing because we are trying to unset a key that's not there in the first place, not because the key is empty. If such a key exists for whatever reason, then we should be able to unset it (if only to fix the situation).

onyo/lib/tests/test_commands_get.py
```diff
@@ -0,0 +1,398 @@
...
+    pytest.raises(ValueError,
+                  onyo_get,
+                  inventory,
+                  paths=[inventory.root / "not-existing" / "TYPE_MAKER_MODEL.SERIAL"])
+
+    # get outside the repository
+    pytest.raises(ValueError,
+                  onyo_get,
+                  inventory,
+                  paths=[(inventory.root / "..").resolve()])
```
bpoldrack, line 25: Not a big deal, but what is the `resolve` supposed to do?

bpoldrack (commented):
Thx, @TobiasKadelka.
Left a couple of comments. Turning into issue and resolve conversation would be fine with me, though.

onyo/lib/commands.py
TobiasKadelka: Aren't pseudo keys reserved keys, too, just as a special case? Or is there another word we have for that?

onyo/lib/commands.py
TobiasKadelka: For now I added the whitespace test, and I will create an issue about this in regards to the inventory operations, and begin working on that in a separate PR.
Thanks for the feedback!

onyo/lib/commands.py
TobiasKadelka: I created the issue #535 to discuss handling of RESERVED_KEYS and PSEUDO_KEYS further, thanks for your feedback!

bpoldrack: > Aren't pseudo keys reserved keys, too, just as a special case? Or is there another word we have for that?

Not sure, I parse that. We have a word for that: "pseudo keys" ;-)

`RESERVED_KEYS` have different definition than `PSEUDO_KEYS`. And it's kinda different concepts. Reserved keys are for specifying properties that are not recorded as key-value. That's `template` for example.
Whereas pseudo keys are key-value pairs of an asset (and can be queried as such), but the values aren't defined arbitrarily in the YAML but reflect information retrieved from file system or git. That's `path` currently.

onyo/lib/commands.py
```diff
@@ -917,17 +924,18 @@ def onyo_set(inventory: Inventory,
         If a given path is invalid or changes are made that would result in
         renaming an asset, while `rename` is not true, or if `keys` is empty.
     """
-    paths = paths or []
+    paths = paths or [inventory.root]
     if not keys:
         raise ValueError("At least one key-value pair must be specified.")
-
+    if not all(keys.keys()) or any(key.isspace() for key in keys.keys()):
```
bpoldrack, line 930: How about this? ;-)
```suggestion
    if any(not k.strip() for k in keys.keys()):
```

TobiasKadelka: Actually, that fails on `None` for example, so I will do something like this, but have to adjust with a second condition still.

bpoldrack: True. Thx!

Issue #525, opened 2024-01-23
BUG: deploying onyo-demo is not working anymore


TobiasKadelka:
I am stil trying to identify the exact problem, but want to share/document what I found in case somebody else knows how to solve the problem.

Here an example deploy-demo workflow that failed:
https://github.com/psyinfra/onyo/actions/runs/7625219190/job/20769022194
Scrolling to the end shows this error at step "push to onyo-demo":
```
push to onyo-demo
remote: Support for password authentication was removed on August 13, 2021.
remote: Please see https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls for information on currently recommended modes of authentication.
fatal: Authentication failed for 'https://github.com/psyinfra/onyo-demo/'
Error: Process completed with exit code 128.
```

This fails because of these two lines in `.github/workflows/deploy_demo.yaml`; the second line is the point of failure, but as far as I can see, the problem is setting the location of the onyo repo incorrectly:
```
          printf '\nset the location of the demo repo\n'
          git -C /tmp/fresh_demo remote add origin https://aqw:${ONYO_DEMO_ACCESS_TOKEN}@github.com/psyinfra/onyo-demo

          printf '\npush to onyo-demo\n'
          git -C /tmp/fresh_demo push --force --set-upstream origin 
```

aqw:
The access token may have expired. I can look at this later.

aqw:
Ok, I found a better solution (because the access tokens have a max ttl of 1 year).

GitHub automatically sets up tokens to allow jobs to push back to their own repo. So the ideal solution is:

1) Remove the push part from the `onyo` tests. But keep the rest of the `onyo demo` generation. It's a useful test.
2) Add a GitHub Actions workflow to the demo repository, and have it [run daily](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule). It should pull from the `onyo` repo (unprivileged), run the demo generation, if successful, push the results to its own repo.

This way, it never expires, and we don't need to deal with secrets across repos.

TobiasKadelka:
Since workflows are defined in the directory `.github/workflows/` this might mean we have to add them at what you describe at some point during the demo-repo-generation so that it automatically gets pulled into the demo repo.
Otherwise I think this might "overwrite" the workflow you want to have after the first "pull" of the demo repo from the main onyo project.

So while I like your proposed solution, it needs a bit more effort to make sure it actually works like we want it to ;)

[...]

PR #531, opened 2024-02-19
demo: push demo repo to branch of onyo main repository

TobiasKadelka:
Works towards #525 

This PR changes the demo script so that the generated demo repo is not directly pushed to https://github.com/psyinfra/onyo-demo. Instead, the demo is build and pushed to a branch of this repository, and the demo repository will in the future pull this branch regularly.


aqw (changes requested):
The PR for this repo should simply remove support for publishing the demo.

Then a PR should be opened for the `onyo-demo` repo. That once should include the workflow for the "cron" job.

Daily it will:
- pull latest `onyo`
- use it to build a new `onyo-demo`
- push back to `onyo-demo`

PR #533, opened 2024-02-21
Command decorator for ensuring suitable inventory/repo state

bpoldrack:
This introduces a decorator for command implementations in order to have commands fail right away if there a running either on a dirty worktree or - in python context - on an `Inventory` object that still has pending operations.

A more command-tailored assessment of what's a valid state for a command to run on is imaginable, but we don't really have  a usecase for that so far and furthermore this introduces how to do it. We could still break theat down into several, different decorators anytime we encounter an actual need for this.

### DIFF
--- a/onyo/commands/tests/test_edit.py
+++ b/onyo/commands/tests/test_edit.py
@@ -61,7 +61,8 @@ def test_get_editor_precedence(repo: OnyoRepo) -> None:
     """
     # set locations
     repo.set_config('onyo.core.editor', 'local', location='local')
-    repo.set_config('onyo.core.editor', 'onyo', location='onyo')
+    # Use onyo-config to also commit and not end up with a modified worktree here:
+    subprocess.run(["onyo", "config", '--add', "onyo.core.editor", "onyo"])
     os.environ['EDITOR'] = 'editor'
 
     # git should win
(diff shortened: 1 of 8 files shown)


TobiasKadelka:
Hey, the PR looks good for me :)
There are just 2 minor points I noticed while reading through it:
1. Most of the cases in the new function `raise_on_inventory_state()`, especially the ones actually erroring, seem not to be triggered by any tests (if one trusts codecov). I think if a PR does not test the new functionality added, there should be at least an issue mentioning the need for this or a code todo or something :)
2. Now that we begin to have more and more different error classes, some of them deserve more specific names and/or doc-strings, since I find it not too self-explanatory which one of these should be raised in any given situation:
```
class OnyoRepoError(Exception):
    """Thrown if something is wrong with an onyo repository."""
class OnyoInvalidRepoError(OnyoRepoError):
    """Thrown if the repository is invalid."""
```

TobiasKadelka:
To be clear, neither of the two points is a blocker for this PR to be merged -- but they should be addressed in the future :)

bpoldrack:
Agree.
Just quick remark on the exceptions you pointed to: Note, that one is the base class of the other. That implies: You use the subclass if it fits. If your failure is less specific you go for the broader scope (base class). Or add a new one, of course.

PR #540, opened 2024-03-01
DOCS: correct changes in syntax, remove out-of-date information

TobiasKadelka:
This PR is not meant to add anything new or to make the docs more complete; Instead it brings everything up to date, corrects the docs in regards to changes in onyo (e.g. `onyo new --path` gets a directory, not an asset name) or removes information that is simply out of date (e.g. "_" is not a reserved character anymore and can be used in keys of asset names).

This should allow other people to focus later on adding new information and help during the hackathon to either use the documentation or see that information is just not given yet, instead of wondering permanently if the user does something wrong or if the docs might just be incorrect.

### DIFF
--- a/docs/source/cmd_cat.rst
+++ b/docs/source/cmd_cat.rst
@@ -17,6 +17,10 @@ Example Usage
 
     onyo cat accounting/Bingo\ Bob/laptop_lenovo_T490s.abc123
 
+    type: laptop
+    make: lenovo
+    model: T490s
+    serial: abc123
     RAM: 16GB
     display_size: '14.6'
     touch: yes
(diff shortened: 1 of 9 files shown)


bpoldrack (commented):
Changes seem fine.
Did you have a look at the build failure, @TobiasKadelka ?

TobiasKadelka:
Thanks for the quick feedback!
I saw it on Friday and it seemed to be because of a header/link. I will check that today, and then let you immediatly know :)

TobiasKadelka:
I updated this PR to fix the build-problem, so feel free to review when you have time.

bpoldrack:
Thanks!

Issue #542, opened 2024-03-01
investigate "rich print" behavior and unify both print functions of UI class


TobiasKadelka:
@aqw found an error that appeared in one of the tests for `onyo_get()`:

```
❱ pytest -vv
[...]
=================================================================================== FAILURES ===================================================================================
___________________________________________________________________________ test_onyo_get_no_matches ___________________________________________________________________________

inventory = <onyo.lib.inventory.Inventory object at 0x7fbf7aedb9d0>, capsys = <_pytest.capture.CaptureFixture object at 0x7fbf7ae094d0>

    @pytest.mark.repo_contents(
        ["one_that_exists.test", "type: one\nmake: that\nmodel: exists\nserial: test\nsome_key: value"])
    @pytest.mark.ui({'yes': True})
    def test_onyo_get_no_matches(inventory: Inventory,
                                 capsys) -> None:
        """`onyo_get()` behaves correctly when `match` matches no assets."""
        asset_path1 = inventory.root / "somewhere" / "nested" / "TYPE_MAKER_MODEL.SERIAL"
        asset_path2 = inventory.root / "one_that_exists.test"
        matches = [Filter("unfound=values").match]
    
        # `onyo_get()` is called, but no assets match
        onyo_get(inventory,
                 match=matches)  # pyre-ignore[6]
    
        # verify output contains no assets because nothing matched
        output = capsys.readouterr().out
>       assert "No assets matching the filter(s) were found" in output
E       AssertionError: assert 'No assets matching the filter(s) were found' in 'No assets matching the \x1b[1;35mfilter\x1b[0m\x1b[1m(\x1b[0ms\x1b[1m)\x1b[0m were found\n'
```

I can reproduce this error exclusively when I run all tests -- if I just run the one for the file with the error (`REPO_ROOT=$PWD pytest onyo/lib/tests/test_commands_get.py -vv`) the test is not raised. 

In the test, we assert for a specific string, but `ui.rich_print()` adds some special characters to the output, hence the error.

--------------

#541 changes the rich-print to a normal print, which removes the occurence of the error.
1. However, since the error seems to depend on first running some CLI tests, there seems to be something bleeding over into the python-tests -- possibly something connected to the UI class/object. This should be investigated.
2. We want to have just one `ui.print()` and remove `ui.rich_print()` on the long run, i.e. unify both functions into one that decides when to rich-print and when not.
3. #55 wants to colorize certain output; this issue should be considered when unifying both functions.

[...]

PR #543, opened 2024-03-03
BF: Disable rich's syntax highlighting

bpoldrack:
Syntax highlighting seems to be enabled by default. We don't print code, we don't want that. Brought to light by issue #542.

### DIFF
--- a/onyo/lib/ui.py
+++ b/onyo/lib/ui.py
@@ -72,8 +72,8 @@ class UI(object):
         else:
             self.logger.setLevel(logging.INFO)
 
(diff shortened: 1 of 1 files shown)


aqw:
This LGTM.

I do think highlighting has value beyond code blocks, and it useful such as how man-pages use bolding to make skimming easier.

However... I don't think blindly enabling highlighting everywhere is the right thing for Onyo's use case. And now we have a clear reason not to do so.

That's a long way of saying that I agree with this change, and can also foresee that at some point later we may want to enable highlighting in specific situations, and we'll adjust the tests for that if/when that happens.

Issue #496, opened 2023-11-08
Tests for shell_completion


bpoldrack:
While these tests do exist at `onyo/commands/tests/test_shell-completion.py`, their correctness in light of somewhat changed command behavior/parameterization needs to be double-checked.

Issue #204, opened 2022-09-28
BUG: spaces in the CWD breaks tab completion for files/folders


aqw:

TobiasKadelka:
I had a bit of a struggle to reproduce this, because this issue is missing a description, and tab completion does work partially, hence I want to expand on this:

```
# being in a dir "ware house" with space in CWD
(valid) tkadelka in ~/onyo/ware house on git:master
❱ ls                                                                                     130 !
laptop_asdf_asdf.fauxznwSUB
(valid) tkadelka in ~/onyo/ware house on git:master
❱ onyo rm <tab tab>
# This does nothing, here it does not work --- **This is the bug of this issue**.
(valid) tkadelka in ~/onyo/ware house on git:master
❱ onyo rm ./ <tab tab>
laptop_asdf_asdf.fauxznwSUB
# with the ./ in the beginning, it completes anyways
```

To complete the issue, here the "correct" example which works for CWDs without a space:
```
(valid) tkadelka in ~/onyo/warehouse on git:master
❱ onyo rm <tab tab>
headphones_JBL_pro.325gtt          laptop_apple_macbook.9r5qlk      
# here it shows correctly the available devices
```

Issue #157, opened 2022-09-14
shell-completion: support exclusion


aqw:
All flags can be used with each other. This is not the reality in practice. The most simple example is help text. When the `-h` argument is used, no other flags should be suggested, as they are all effectively no-op.

PR #545, opened 2024-03-04
BF/NF: improve tab completion

aqw:
This PR greatly simplifies the python logic sitting between argparse and ZSH. This simplification is achieved by simply ripping it all out. ;-)

The zsh completion is now a static file served by `onyo shell-completion`. I have chosen to keep this command because zsh does not (yet) have an XDG location to install tab-completion scripts for ZSH (unlike the bash-completion project). So for now, the best bet for the user is to `source` the output of this script.

The file is named according to zsh completion norms, so anyone packaging `onyo` can more easily install it into `vendor-completions`.

This also addresses the following bugs:
- #204
- #157 
- `-C` didn't tab complete from CWD
- `-C` to a dir with spaces caused arguments to subcommands to fail tab completion (related, but different bug from #204)
- significantly trimmed tab-completion help text so that it's actually readable and doesn't break rendering
- `onyo get -sort-descending` -> `onyo get --sort-descending`
- typos

This PR does not address adding/expanding tests for shell-completion (#496). That is still very important, but the focus here is to get much needed improvements in before the hackathon. That issue will remain open, to serve as a reminder.


onyo/commands/shell_completion.py
```diff
@@ -478,55 +28,13 @@ def shell_completion(args: argparse.Namespace) -> None:
         $ source <(onyo shell-completion)
         $ onyo --<PRESS TAB to display available options>
     """
-    parser = args.parser
+    content = ''
+    shell_completion_dir = Path(Path(__file__).resolve().parent.parent, 'shell_completion')
```
bpoldrack, line 32: This would be more consistent with the rest of the code base (and I think easier to read).
Also: Slightly faster than invoking the `Path` constructor again IIRC.

```suggestion
    shell_completion_dir = Path(__file__).resolve().parent.parent / 'shell_completion'
```

bpoldrack (commented):
> This PR greatly simplifies the python logic

It absolutely does! :-D

Thanks! (Also for the "minor fixes" alongside)

Just one minor comment - feel free to ignore, though. I'm fine with merging either way.

onyo/commands/shell_completion.py
aqw: Thanks for the suggestion! I will update and push.

TobiasKadelka:
Thanks for the PR, it is cool to see so many shell-completion bugs being addressed!

Issue #546, opened 2024-03-04
Inventory methods should consistently account for pending operations


bpoldrack:
This is most important for enabling the full potential of python API (and future commands/features by extension).

Inventory methods catch invalid cases based on assessing whether something is or isn't an asset, dir, etc.
They would need to systematically account for the impact of already pending operations, though.
This involves a change to executors that currently report paths to stage/commit. That reporting needs to be structured and more informative. For example, we'd need to now what paths may be new or removed by an operation, so that subsequent operations can tell, whether an input *will* is expected to be valid at the point of execution.

One flavor of that is addressing #426 (move and rename dir in one go): First operation moves `src` to `dst1`, second one is renaming `dst1` to `dst2`. Both think that `dst1` is a path that needs committing, but at the point of committing that path doesn't exist and was never known to git.

bpoldrack:
Note: Based on "hacky" solutions so far, it seems like `Inventory` should have equivalents to `OnyoRepo.is_inventory_dir` and the likes, that would additionally check the pending operations (whereas the repo methods only check what's already in the repo).
`Inventory` methods would then simply have their conditions based on these instead and otherwise the logic remains the same.

Issue #501, opened 2023-11-09
Document Inventory class


bpoldrack:
Add/edit docstrings for everything in `Inventory`. This should include a revisit of its tests.

PR #547, opened 2024-03-04
Allow mv to move + rename

### LINKED ISSUE #426: BUG: `onyo mv` does no allow renaming and moving a dir in one call

bpoldrack:
This allows `onyo mv` to move and rename a directory in one go, but still record two separate operations.

The first two commits are a bit of a hack. Issue #546 records the need for a more generalized solution. Was aware of that problem before, but I thought it would only be relevant for more sophisticated usage of python API. Turns out, it's not ;-)

Closes #426

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -620,12 +620,21 @@ def onyo_mv(inventory: Inventory,
         # MOVE special case
         subject = "mv"
         move_asset_or_dir(inventory, sources[0], destination.parent)
-    elif len(sources) == 1 and sources[0].is_dir() and destination.parent.is_dir():
+    elif len(sources) == 1 and sources[0].is_dir() and destination.parent.is_dir():  # TODO: last condition necessary?
         # RENAME directory
         subject = "ren"
-        inventory.rename_directory(sources[0], destination)
+        if sources[0].parent != destination.parent:
+            # This is a `mv` into non-existent dir not under same parent.
+            # Hence, first move and only then rename.
+            subject = "mv + " + subject
+            inventory.move_directory(sources[0], destination.parent)
+            inventory.rename_directory(destination.parent / sources[0].name, destination)
+            # TODO: Replace - see issue #546:
+            inventory._ignore_for_commit.append(destination.parent / sources[0].name)
+        else:
+            inventory.rename_directory(sources[0], destination)
     else:
-        raise ValueError("Can only move into an existing directory or rename a single directory.")
+        raise ValueError("Can only move into an existing directory, or rename a single directory.")
 
     if inventory.operations_pending():
         ui.print("The following will be {}:".format("moved" if subject == "mv" else "renamed"))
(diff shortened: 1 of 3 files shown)


onyo/lib/commands.py
```diff
@@ -620,12 +620,21 @@ def onyo_mv(inventory: Inventory,
         # MOVE special case
         subject = "mv"
         move_asset_or_dir(inventory, sources[0], destination.parent)
-    elif len(sources) == 1 and sources[0].is_dir() and destination.parent.is_dir():
+    elif len(sources) == 1 and sources[0].is_dir() and destination.parent.is_dir():  # TODO: last condition necessary?
```
aqw, line 623: Thinking about this: the last condition isn't necessary to know what mode we're in (move + rename mode). But having a valid parent is necessary (for all `mv` modes). Perhaps the check is in the wrong location.

aqw (approved):
This LGTM. Thanks!

onyo/lib/commands.py
bpoldrack: Yes, location is the point. The underlying `Inventory` method should fail on that (and cheaper than an additional `stat` by `pathlib`). However, need to double-check and doesn't matter for what this PR is about.

onyo/lib/inventory.py
```diff
@@ -190,6 +191,29 @@ def _get_pending_asset_names(self) -> list[str]:
...
+
+        Returns
+        -------
+        list of Path
+            Inventory dirs about to be created.
+        """
+        dirs = []
+        for op in self.operations:
+            if op.operator == OPERATIONS_MAPPING['new_directories']:
+                dirs.append(op.operands[1])
```
TobiasKadelka, line 213: It looks to me like the tests never go into this "if" case. It should not be hard or time consuming to add a test for this inventory function, too. :)

Alternatively, if this is more work than I think, it can be addressed with https://github.com/psyinfra/onyo/issues/501.

bpoldrack: Yes, unused ATM. #546 is the spot to rework and test this. :)

Issue #309, opened 2023-03-14
Add API workflow demo


TobiasKadelka:
While adding the `Repo.clear_caches()` functionality to the existing commands which were missing it, I  noticed how much we should have something like API-workflow tests, or like the demo, since it is quite hard to verify that not some other functions (e.g. `restore()`, `add()` and `commit()` as git functions) also would need a clearing of caches.

I think these other functions should not require this currently, because they get mainly called by the command-functions, or in the command-files, but we should add these tests soon anyways to be sure. 

This should not replace the need for the other API tests needed and described in issues, but instead supplement the tests and verify that the API works and can be used in a consistent way even outside the existing commands we wrote.

PR #549, opened 2024-03-05
BF: `rm` handles duplication and fails on inventory root

### LINKED ISSUE #519: BUG: calling `onyo rm <dir> <dir> <dir>` multiple times with same dir leaves git unclean
This issue is about calling `onyo rm a/ a/`, but the problem should be considered together with `onyo rm a/b/ a/` (e.g. different paths are given, but the call would on disk try to delete the same destination twice)!

The bug is that I can call `onyo rm` multiple times with the same input and onyo even begins running the command, but then errors and leaves the git tree unclean.

```
❱ onyo rm admin admin admin                                                              130 !
The following will be deleted:
-/Users/tkadelka/onyo/admin/Karl Krebs/laptop_apple_macbookpro.9sdjwa
-/Users/tkadelka/onyo/admin/Karl Krebs
-/Users/tkadelka/onyo/admin
-/Users/tkadelka/onyo/admin/Karl Krebs/laptop_apple_macbookpro.9sdjwa
-/Users/tkadelka/onyo/admin/Karl Krebs
-/Users/tkadelka/onyo/admin
-/Users/tkadelka/onyo/admin/Karl Krebs/laptop_apple_macbookpro.9sdjwa
-/Users/tkadelka/onyo/admin/Karl Krebs
-/Users/tkadelka/onyo/admin
Save changes? No discards all changes. (y/n) y
ERROR: [Errno 2] No such file or directory: '/Users/tkadelka/onyo/admin/Karl Krebs/.anchor'
```

Basically, when I call the command like this, it seems to work  (e.g. a diff is presented because all paths are "possible to delete" -> no error).  But then onyo tries to execute, and errors --- presumably during the second deletion when it tries to delete the already-not-existing-anymore file.

This causes the problem that the command stops in between executions, which means the commit is not executed and the git tree is afterwards unclean:

```
❱ git status                                                                             130 !
On branch master
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    admin/.anchor
	deleted:    admin/Karl Krebs/.anchor
	deleted:    admin/Karl Krebs/laptop_apple_macbookpro.9sdjwa

no changes added to commit (use "git add" and/or "git commit -a")
```

Would we actually want to allow this type of szenario (and how we would like to handle sub-paths e.g. `onyo rm a/b/  a`)? We can either allow it and make sure it does not fail on such calls, or disallow it, and then error correctly: 

Dis-allowing:
1. check input paths for duplicates/sub-paths in the beginning and error if duplicates are found. 

Allowing:
1. de-duplicate the paths before deleting anything.
2. check before deleting again if the path does actually still exist. 

Technically it would also be possibly to find a solution that just reverses/restores changes or fscks/commits them afterwards if such an error occours, though I find these solutions always problematic. IMO just deciding to allow/dis-allow this behavior and then either add appropriate checks or errors should be fairly straight forward and lead to the best behavior.

### LINKED ISSUE #520: BUG: `onyo rm .` in root prints illegal paths in diff
Calling `onyo rm .` in the **root directory of the onyo repository** prints the paths `.git` and `.onyo` even though they are not allowed to be deleted by the command. When I try to delete them then, Onyo does error correctly, so nothing is deleted incorrectly as I found, but the diff is still a lie and should not be printed in this form.

```
❱ onyo rm .
The following will be deleted:
-/Users/tkadelka/onyo/.onyo
-/Users/tkadelka/onyo/.git
-/Users/tkadelka/onyo
Save changes? No discards all changes. (y/n) y
ERROR: [Errno 1] Operation not permitted: '/Users/tkadelka/onyo/.onyo'
```

bpoldrack:
- Fixes `onyo rm` to account for already pending removals. This should be done more consistently (see #546).
- Fixes `onyo rm` to fail when trying to remove the inventory's root dir.

Note, that changes in `Inventory.remove_directory` are mostly about recursive removal of directories. For example, changing a condition to recursively call itself on `path.is_dir` rather than considering `is_inventory_dir` is done, because the latter, more restrictive condition would be checked again during recursion anyway.

Refactoring of the "dummy operation" for generic file removal is done, so it can be easily accounted for by `Inventory._get_pending_removals`, which would be harder to do if it remained that unnamed ad-hoc operation.

Lastly, note that the unskipped test's referral to #309 was wrong to begin with - #519 is what it was about.

(Closes #519)
(Closes #520)

### DIFF
--- a/onyo/lib/inventory.py
+++ b/onyo/lib/inventory.py
@@ -49,6 +49,7 @@ from onyo.lib.recorders import (
     record_move_directories
 )
 from onyo.lib.utils import deduplicate
+from onyo.lib.ui import ui
 
 
 @dataclass
@@ -98,6 +99,10 @@ OPERATIONS_MAPPING: dict = {'new_directories': InventoryOperator(executor=exec_n
                             'move_assets': InventoryOperator(executor=exec_move_assets,
                                                              differ=differ_move_assets,
                                                              recorder=record_move_assets),
+                            'remove_generic_file': InventoryOperator(
+                                executor=partial(generic_executor, lambda x: x[0].unlink()),
+                                differ=differ_remove_assets,
+                                recorder=lambda x: dict()),  # no operations record for this, not an inventory item.
                             }
 
 
@@ -214,6 +219,29 @@ class Inventory(object):
             elif op.operator == OPERATIONS_MAPPING['move_directories']:
                 dirs.append(op.operands[1] / op.operands[0].name)
         return dirs
+
+    def _get_pending_removals(self) -> list[Path]:
+        """Get paths that are removed by pending operations.
+
+        Notes
+        -----
+        Just like `_get_pending_asset_names` and `_get_pending_dirs`,
+        this needs to be replaced by a more structured way of assessing
+        what's in the queue. See issue #546.
+
+        Returns
+        -------
+        list of Path
+            To be removed paths.
+        """
+        paths = []
+        for op in self.operations:
+            if op.operator in [OPERATIONS_MAPPING['remove_directories'],
+                               OPERATIONS_MAPPING['remove_assets'],
+                               OPERATIONS_MAPPING['remove_generic_file']]:
+                paths.append(op.operands[0])
+        return paths
+
     #
     # Operations
     #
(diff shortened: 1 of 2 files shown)


TobiasKadelka:
Hey, thanks a lot for the PR!
I will merge it, and I want to state explicitely how happy I am that #519 and #520 are resolved! ;)

PR #551, opened 2024-03-05
`new`: --clone from existing asset

### LINKED ISSUE #461: FEATURE: onyo new creating a duplicate/clone (with new serial)
In the past while managing inventory we often used the clone feature to create a duplicate of an asset. Here in Onyo we should make sure that the serial needs to be redefined during this process. By default it could generate a faux serial if no new one was given.

bpoldrack:
Allow to clone new asset(s) from an existing one as an ad-hoc template.

- Closes #461

@aqw: shell completion doesn't seem to work yet. I get:
```
_arguments:comparguments:327: invalid argument: (-c --clone){-c,
```

when trying to complete. Can you have a look please?

### DIFF
--- a/onyo/shell_completion/zsh/_onyo
+++ b/onyo/shell_completion/zsh/_onyo
@@ -120,7 +120,8 @@ _onyo() {
             new)
                 args+=(
                     '(- : *)'{-h,--help}'[show this help message and exit]'
-                    '(-t --template)'{-t,--template}'[template to seed the new assets]:TEMPLATE:_path_files -W "$(_template_dir)" -g "*(.)"'
+                    '(-t --template -c --clone)'{-t,--template}'[template to seed the new assets]:TEMPLATE:_path_files -W "$(_template_dir)" -g "*(.)"'
+                    '(-c --clone -t --template)'{-c,--clone}'[asset path to clone from]:CLONE:_files -W "$(_onyo_dir)"'
                     '(-e --edit)'{-e,--edit}'[open new assets in an editor before creation]'
                     '(-k --keys)'{-k,--keys}'[key-value pairs to set in the new assets]:*-*:KEYS: '
                     '(-p --path)'{-p,--path}'[directory to create new assets in]:PATH:_files -W "$(_onyo_dir)"'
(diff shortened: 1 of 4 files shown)


aqw:
Not a problem. You need to add the following to `onyo/shell_completion/zsh/_onyo`:

```
            new)
                args+=(
                    ...
                    '(-c --clone -t --template)'{-c,--clone}'[path of an asset to clone]:CLONE:_files -W "$(_onyo_dir)"'
                    ...
                )
                ;;
```

I assume that `--template` and `--clone` exclude each other, so I added that. If so, then please also add that to `--template` in that file. It's the opening pair of `()`.

onyo/shell_completion/zsh/_onyo
```diff
@@ -121,6 +121,7 @@ _onyo() {
                 args+=(
                     '(- : *)'{-h,--help}'[show this help message and exit]'
                     '(-t --template)'{-t,--template}'[template to seed the new assets]:TEMPLATE:_path_files -W "$(_template_dir)" -g "*(.)"'
+                    '(-c --clone)'{-c, --clone}'[asset path to clone from]:CLONE:_files -W "$(_onyo_dir)"'
```
aqw, line 124: You have a spurious space here: `{-c, --clone}` -> `{-c,--clone}`

We're mixing compsys syntax with brace expansion here. So it can be a bit odd.

bpoldrack: Thx much!

onyo/shell_completion/zsh/_onyo
```diff
@@ -121,6 +121,7 @@ _onyo() {
                 args+=(
                     '(- : *)'{-h,--help}'[show this help message and exit]'
                     '(-t --template)'{-t,--template}'[template to seed the new assets]:TEMPLATE:_path_files -W "$(_template_dir)" -g "*(.)"'
```
aqw, line 123: This line also should be adjusted to exclude `-c` and `--clone`.

bpoldrack: That should've been obvious. Thanks!

bpoldrack: Pushed

aqw (approved):
LGTM

PR #554, opened 2024-03-07
TST: check clean-ness of git worktrees for python command functions

TobiasKadelka:
This could not be tested before, because the `inventory` fixture constructed an inventory which contained un-saved files.
Since that behavior was changed, commands that might change the worktree can check it again, especially when they do some changes, or to verify that nothing was left unclean after errors where triggered.

There is no issue as such for this, instead it removes the TODOs in the code describing the necessity of this.


aqw (commented):
I'm not familiar with the background behind this, so I will leave it to @bpoldrack for final approval.

But to my eye, this reads straightforward and looks good.

bpoldrack (approved):
Thanks, @TobiasKadelka !

PR #555, opened 2024-03-07
Docs: add examples and descriptions to CLI usage

TobiasKadelka:


docs/source/cmd_new.rst
```diff
@@ -8,6 +8,31 @@ onyo new
...
+new assets at once with a different value for each asset.
+
+**directory**
+
+    The ``directory`` key is an alternative to ``onyo new --path`` to specify
+    the location in which to create new assets.
+
+**is_asset_directory**
+
+    Can be set to create new assets as an asset directory instead of an asset
```
aqw, line 27: Are "asset directories" explained as a concept elsewhere in the docs yet? If so, it would be good to link "asset directory" to it.

docs/source/cmd_new.rst
```diff
@@ -30,17 +55,16 @@ This command writes a YAML file to ``shelf/laptop_lenovo_T490s.abc123``:
    model: T490s
    serial: abc123
 
-Create multiple new assets with content, and overwrite the default message
-with a more helpful one describing the action:
+**Create multiple new assets with content in different locations, and overwrite
+the default message with a more helpful one describing the action**
```
aqw, line 59: I recommend removing "with a more helpful one describing the action".

docs/source/cmd_new.rst
TobiasKadelka: Thanks for your quick feedback! Nope, it isn't, but yes it should. 

It should also be stated that `is_asset_directory` can just be True or False. And this still needs to be verified/tested by me.

If this does not yet work, I might remove the description of the parameter as a whole from the "reserved keys" section and just describe what is tested and functions.

onyo/commands/new.py
```diff
@@ -68,8 +68,14 @@
...
     Create new ``DIRECTORY/ASSET``\\(s), and add contents with ``--template``,
-    ``--keys`` and ``--edit``. If the directories do not exist, they will be
-    created.
+    ``--keys``, ``--clone`` and ``--edit``. If the directories do not exist,
+    they will be created.
+
+    When the same key is specified multiple times with ``onyo new --keys [...]``
+    onyo creates a new asset for each of them. A key that is specified just
+    one time will be applied to all new assets.
+    All keys that are used more than one time must be used the same time to
```
aqw, line 77: "All keys that are used more than one time must be used the same time"

This doesn't make sense to me. I think what you're trying to describe is that if creating two assets, and setting keys for them, all keys need to be specified two times. I think an example will also help here.

aqw (changes requested):
Looks good overall. One blocker on the multi-key language. Otherwise +1.

onyo/commands/new.py
TobiasKadelka: That is not exactly what I try to describe, which does underlines the need for adjustment:

"When creating X new assets, all keys must be specified exactly 1 or X to times; if specified once, a key will applied to all assets with the same value, otherwise they will be applied to each new asset in the order of specification."

docs/source/cmd_new.rst
TobiasKadelka: It does not work for me, so I while delete the description of it until I now how and that it works.

onyo/commands/new.py
aqw: I like this much better. Way more clear.

One oops:

> 1 or X to time

Remove "to".

TobiasKadelka: Great catch 🤦

aqw (approved):
LGTM

PR #556, opened 2024-03-07
BF: small shell-completion fixes

aqw:
Standardizing on `_files` should be a no-op. It just unifies the syntax for the sake of consistency.

The misbehavior of `onyo new --path` was caught by @bpoldrack :-)


TobiasKadelka:
While this looks in general good to me, I am struggling consistently to review the shell-completion functionality because I do not fully understand it.

aqw:
> While this looks in general good to me, I am struggling consistently to review the shell-completion functionality because I do not fully understand it.

Yeah; that's a fair concern. We'll just have to roll with that until more people become familiar with it.

In the commit message, I included the relevant section from the upstream docs, to help explain.

PR #557, opened 2024-03-07
Asset dirs

bpoldrack:


onyo/commands/tests/test_mkdir.py
```diff
@@ -153,25 +153,6 @@ def test_dir_exists(repo: OnyoRepo, directory: str) -> None:
     assert repo.git.is_clean_worktree()
 
 
-@pytest.mark.repo_files(*directories)  # used as files to test errors
-@pytest.mark.parametrize('file', directories)
-def test_dir_exists_as_file(repo: OnyoRepo, file: str) -> None:
```
TobiasKadelka, line 158: I think this test (and similar CLI tests later) should not be deleted.

Instead, the assertions just need minimal adjustments, e.g. not erroring anymore, and e.g.
`assert Path(file).is_file()` -> `assert Path(file).is_directory()`.

onyo/lib/tests/test_commands_mkdir.py
```diff
@@ -18,13 +17,6 @@ def test_onyo_mkdir_errors(inventory: Inventory) -> None:
                   dirs=[dir_path],
                   message="some subject\n\nAnd a body")
 
-    # mkdir on existing asset path
```
TobiasKadelka, line 21: I would additionally add here a test that verifies onyo errors correctly when calling `onyo mkdir` on a directory which is already an asset directory :)

Related to #565

onyo/lib/tests/test_commands_mv.py
```diff
@@ -36,14 +36,6 @@ def test_onyo_mv_errors(inventory: Inventory) -> None:
                   destination=dir_path / "doesnotexist" / "somewhere",
                   message="some subject\n\nAnd a body")
 
-    # move directory to existing file
-    pytest.raises(ValueError,
```
TobiasKadelka, line 40: Instead, this should be a test that checks for an error if `onyo mv` is tried to be used to rename an asset dir, which is illegal.

Related to #562

onyo/lib/tests/test_commands_mv.py
```diff
@@ -239,3 +231,32 @@ def test_onyo_mv_and_rename(inventory: Inventory) -> None:
...
+
+    onyo_mv(inventory,
+            source=source,
+            destination=asset_path,
+            message="Move asset into asset to create asset dir")
+    assert inventory.repo.git.get_hexsha('HEAD~1') == old_hexsha
+    assert inventory.repo.git.is_clean_worktree()
+    assert not source.exists()
+    # Turned destination into asset dir:
+    assert inventory.repo.is_asset_dir(asset_path)
```
TobiasKadelka, line 260: Nitpick: we might here want to test both, the caches and the filesystem, so as to be sure that both are updated through the change.

onyo/lib/tests/test_commands_new.py
```diff
@@ -255,3 +260,18 @@ def test_onyo_new_clones(inventory: Inventory) -> None:
     # equals existing asset except for path and serial:
     assert all(v == new_asset[k] for k, v in existing_asset.items() if k not in ['serial', 'path'])
     assert new_asset['serial'] == 'whatever'
+
+
+@pytest.mark.ui({'yes': True})
+def test_onyo_new_asset_dir(inventory: Inventory) -> None:
```
TobiasKadelka, line 266: Based on #561, we should add a test with `"is_asset_directory": False`, and possibly `"is_asset_directory": false`, too, that create a normal asset.

I assume that #561 is just happening for the CLI because everything gets interpreted as a string instead of a bool, but to make sure the value to the key is considered correctly is IMO still worth doing. 

What happens (or should happen) for things like `"is_asset_directory": ["this fullfills truthyness requirements"]`? > depends obviously on the check done, but since this is about a reserved key, it is worth considering :)

onyo/commands/tests/test_mkdir.py
bpoldrack: This should be tested in the python tests, IIRC.
I'd like to not repeat functionality tests in the CLI tests, but rather just test the wrapper does the correct wrapping :)

onyo/lib/tests/test_commands_set.py
```diff
@@ -466,82 +341,44 @@ def test_onyo_set_depth(inventory: Inventory) -> None:
...
 
 
-@pytest.mark.repo_contents(
-    ["one_that_exists.test", "type: one\nmake: that\nmodel: exists\nserial: test\nsome_key: value"])
 @pytest.mark.ui({'yes': True})
-def test_onyo_set_depth_zero(inventory: Inventory) -> None:
-    """Calling `onyo_set(depth=0)` is legal and selects
-    all assets from all subpaths."""
-    key_value = {"this_key": "that_value"}
+def test_onyo_set_asset_dir(inventory: Inventory) -> None:
```
TobiasKadelka, line 345: Maybe this happens already and I just overlooked it (the diff is a bit weird to read at times).

There should be another explicit test that verifies that `onyo_set()` turns an asset into an asset dir.

onyo/lib/tests/test_commands_unset.py
```diff
@@ -70,24 +60,6 @@ def test_onyo_unset_errors(inventory: Inventory) -> None:
                   message="some subject\n\nAnd a body")
 
 
-@pytest.mark.ui({'yes': True})
-def test_onyo_unset_on_empty_directory(inventory: Inventory) -> None:
```
TobiasKadelka, line 74: Should probably  add a test above that verifies that this type of call now errors on directories.

TobiasKadelka (commented):
I really like this one, thanks a lot for the PR! :)

Two things about my comments:

1. The changes are a lot, and the diff is for some files weird for me. I might have requested some additions, which are already there, just not were I looked for them. I hope you can forgive me when this happens :)
2. I know that not everything can/should be tested in this PR into every corner case, but I still added comments requesting some additional tests for scenarios that I consider relevant (e.g. likely to be done), or where I found bugs for, and I think these will be easy to add/fix. Whenever this is not as easy as I think, please add an issue for them, so that they can be added later!

onyo/lib/tests/test_commands_mv.py
bpoldrack: That's not necessary here. We test the command did `commit`. Cache update only needs to be done when committing, so this is included in `commit` and tested for with the respective unit tests.

bpoldrack: Furthermore the `repo.is_xxx` methods are in fact using the cache.

onyo/lib/tests/test_commands_new.py
bpoldrack: There's a confusion here. What is `"is_asset_directory": false` supposed to be? That's not valid python ;-)

As you suspected, it's only about strings coming in from the CLI and being converted correctly before handed off to us by argparse.

bpoldrack: (Fixed already)

onyo/lib/tests/test_commands_set.py
bpoldrack: Lines 367ff and 377ff do that (both directions).

onyo/lib/tests/test_commands_mv.py
bpoldrack: Done.

aqw (approved):
There's no way I can review this in-depth right now. But superficially it looks good to me. And it works.

What's important is that this works and gets in before the hackathon.

Afterwards we will immediately expand test coverage, and gain more confidence that it's actually doing the right thing. ;-)

Issue #395, opened 2023-07-24
Explain tsv-file format for help-text of `onyo new --tsv`


TobiasKadelka:
This came up in https://github.com/psyinfra/onyo/pull/390#discussion_r1270313311
> This should include a description of what the TSV is supposed to look like (don't know from the top of my head).

This should include the required columns:
| type | make | model | serial | directory |

And mention that optional columns can be given as:

| key |
| value1 |
| value2 |

This is the read the docs page of onyo new, explaining the current table format under the example section:
https://onyo.readthedocs.io/en/latest/cmd_new.html#example-usage

bpoldrack:
Note, that as of #403, there's `template` and `is_asset_directory` as well.

bpoldrack:
Ping #486

PR #568, opened 2024-03-08
BF: improve help text for `onyo new`

aqw:
This rewrites and reformats much of the help text for `onyo new`, and also the shared arguments.

Fixes: #395

### DIFF
--- a/onyo/commands/new.py
+++ b/onyo/commands/new.py
@@ -19,21 +19,36 @@ args_new = {
         metavar='TEMPLATE',
         required=False,
         type=template,
-        help='Name of the template to seed the new asset(s)'),
+        help="""
+            Name of a template to populate the contents of new assets.
+
+            This cannot be used with the ``--clone`` flag nor the ``template``
+            Reserved Key.
+        """
+    ),
 
     'clone': dict(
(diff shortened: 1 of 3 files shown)


onyo/commands/new.py
```diff
@@ -19,70 +19,119 @@
...
-        help=(
-            'Key-value pairs to set in the new asset(s). Multiple pairs can be '
-            'specified (e.g. key=value key2=value2). All fields that are part of '
-            'asset filenames (defined in .onyo/config under `onyo.assets.filename`) '
-            'are required. If the value `faux` is assigned to the key `serial`, '
-            'a random, repository-unique string will be filled in instead.')),
+        help="""
+            Key-value pairs to populate content of new assets.
+
+            Each key can key defined either 1 or N times (where N is the number
```
bpoldrack, line 62:
```suggestion
            Each key can be defined either 1 or N times (where N is the number
```

bpoldrack (approved):

onyo/commands/new.py
aqw: Thanks for catching this! Fixed and pushed.

PR #570, opened 2024-03-08
BF: more help improvements

aqw:
This improves the help text primarily for the top-level `onyo --help` and `onyo cat --help`.

It also includes two fixes for formatting help text broadly:
- strip leading and trailing newlines (allowing `"""` syntax)
- make de-rst-ifying consistent across description and arguments

### DIFF
--- a/onyo/main.py
+++ b/onyo/main.py
@@ -21,7 +21,18 @@ class SubcommandHelpFormatter(argparse.RawTextHelpFormatter):
         return parts
 
     def _split_lines(self, text, width):
+        """
+        This is a very, very naive approach to stripping rst syntax from
+        docstrings. Sadly, docutils does not have a plain-text writer. That
+        would be the ideal solution.
+        """
         text = textwrap.dedent(text).strip()
+
+        # `` -> `
+        text = text.replace('``', '`')
+        # remove escapes of characters; everything is literal here
+        text = text.replace('\\', '')
+
         text = super()._split_lines(text, width)
 
         return text
--- a/onyo/main.py
+++ b/onyo/main.py
@@ -21,7 +21,7 @@ class SubcommandHelpFormatter(argparse.RawTextHelpFormatter):
         return parts
 
     def _split_lines(self, text, width):
-        text = textwrap.dedent(text)
+        text = textwrap.dedent(text).strip()
         text = super()._split_lines(text, width)
 
         return text
@@ -32,7 +32,7 @@ class SubcommandHelpFormatter(argparse.RawTextHelpFormatter):
         docstrings. Sadly, docutils does not have a plain-text writer. That
         would be the ideal solution.
         """
-        text = textwrap.dedent(text)
+        text = textwrap.dedent(text).strip()
         text = super()._fill_text(text, width, indent)
 
         # `` -> `
--- a/onyo/main.py
+++ b/onyo/main.py
@@ -103,7 +103,7 @@ def setup_parser() -> argparse.ArgumentParser:
         'cat',
         description=commands.cat.__doc__,
         formatter_class=SubcommandHelpFormatter,
-        help=commands.cat.__doc__
+        help='Print the contents of assets to the terminal.'
     )
     cmd_cat.set_defaults(run=commands.cat)
     build_parser(cmd_cat, args_cat)
@@ -114,7 +114,7 @@ def setup_parser() -> argparse.ArgumentParser:
         'config',
         description=commands.config.__doc__,
         formatter_class=SubcommandHelpFormatter,
-        help=commands.config.__doc__
+        help='Set, query, and unset Onyo repository configuration options.'
     )
     cmd_config.set_defaults(run=commands.config)
     build_parser(cmd_config, args_config)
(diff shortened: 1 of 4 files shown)


onyo/main.py
```diff
@@ -179,7 +190,7 @@ def setup_parser() -> argparse.ArgumentParser:
         'mv',
         description=commands.mv.__doc__,
         formatter_class=SubcommandHelpFormatter,
-        help=commands.mv.__doc__
+        help='Move assets or directories into to a destination directory; or rename a directory.'
```
TobiasKadelka, line 193: "... into to ..." I think one of those might be enough :)

onyo/main.py
```diff
@@ -212,7 +223,7 @@ def setup_parser() -> argparse.ArgumentParser:
         'set',
         description=commands.set.__doc__,
         formatter_class=SubcommandHelpFormatter,
-        help=commands.set.__doc__
+        help='Set the value of keys for matching assets.'
```
TobiasKadelka, line 226: "for matching assets" is IMO missleading. `onyo set --match` does not exist anymore, it must be done via piping, xargs etc, like discussed yesterday, or simply via giving the paths to assets you want to change.

The sentence works perfectly without the word "matching" :)

onyo/main.py
```diff
@@ -245,7 +256,7 @@ def setup_parser() -> argparse.ArgumentParser:
         'unset',
         description=commands.unset.__doc__,
         formatter_class=SubcommandHelpFormatter,
-        help=commands.unset.__doc__
+        help='Remove keys for matching assets.'
```
TobiasKadelka, line 259: Same as for `onyo set` -- there is no more matching for unset than there is for `onyo rm`.

onyo/onyo_arguments.py
```diff
@@ -10,36 +10,49 @@
         required=False,
         default=Path.cwd(),
         type=directory,
-        help='Run Onyo commands from inside of DIR'),
+        help="""
+            Run as if Onyo is started in DIR instead of the current working
```
TobiasKadelka, line 14: Might be nitpicking, but I want to suggest an alternative, because it is not "run as if" but literally run in.

"Run Onyo from DIR instead of the current working directory."

If you consider your version to be more intuitive/understandable feel free to keep it :)

onyo/onyo_arguments.py
```diff
@@ -10,36 +10,49 @@
...
     'quiet': dict(
         args=('-q', '--quiet'),
         required=False,
         default=False,
         action='store_true',
-        help=(
-            'Silence messages printed to stdout. Does not suppress interactive '
-            'editors. Requires the --yes flag')),
+        help="""
+            Silence messages printed to stdout; does not suppress interactive
```
TobiasKadelka, line 44: Again, nitpick, but I notice that you use one time a `;` and another time a `.` and I am not totally sure what motivates different sentence dividers.

onyo/onyo_arguments.py
```diff
@@ -10,36 +10,49 @@
...
     'yes': dict(
         args=('-y', '--yes'),
         required=False,
         default=False,
         action='store_true',
-        help=(
-            'Respond "yes" to any prompts. The --yes flag is required to use '
-            '--quiet')),
+        help="""
+            Respond "yes" to any prompts.
```
TobiasKadelka, line 55: Now I am definitely out of my league but I find "any prompt" or "all prompts" sounds more correct?

onyo/main.py
aqw: Thanks for catching this! Fixed and pushed.

onyo/main.py
aqw: I was leaving those until after Ben's PR is merged. But I can change those now.

aqw: Fixed and pushed.

onyo/main.py
aqw: Fixed and pushed.

onyo/onyo_arguments.py
aqw: Both are ok. IMO "any" means that there could or could not be prompts, and we will answer yes to them. While "all" implies that there /will/ be prompts, and we will answer yes to them.

If you find this to be confusing though (as opposed to a nitpick), then I'm comfortable changing it to improve likelihood of comprehension.

Let me know what you think. :-)

onyo/onyo_arguments.py
aqw: It's a stylistic choice. This is something that there are no clear rules on because we're dealing with sentence fragments. In this case, I chose to use a semi-colon because it was connected to the first idea. The period was then used to begin a new idea.

If you find this to be confusing (vs a nitpick) let me know and I'll commit a more clear approach.

Let me know.

onyo/onyo_arguments.py
TobiasKadelka: Hm yeah, the "all" argument I see. "Any prompt" in singular still sounds the smoothest for me :)

onyo/onyo_arguments.py
TobiasKadelka: It is definitely not confusing and can stay in this form without any loss.
I just felt like it interrupted the flow of reading for like 20 millisecs, hence "nitpick" felt really the only word describing it ;)

onyo/onyo_arguments.py
aqw: The noun following "any" can be either singular or plural, but is usually plural (in my experience).

"We will say hello to any people we see."

I will leave this as is. Thanks for responding. :-)

onyo/onyo_arguments.py
aqw: Cool; thanks for the feedback!

Leaving this as is and resolving.

PR #576, opened 2024-03-10
BF: Don't commit what onyo hasn't staged itself

### LINKED ISSUE #571: `onyo init` commits staged changes
`onyo init`ing a directory with contents largely leaves them unaffected. If it's not a git repo, onyo init's, and only commits the `.onyo` dir.

If it is a git repo, we use it. If there are untracked or tracked changes, onyo ignores them.

However, if the git repository has something changed, onyo will include that in its initialization commit.

While in many ways I view this as the user's fault, I struggle to see a use-case, and it becomes "onyo's fault" since it's included in our commit.

I think `onyo init` should check for staged changes and error, telling the user to unstage or commit their changes.

And if the use /really/ was to combine their changes into onyo's commit, that is what  `git rebase` is for.

bpoldrack:
This limits git-commit calls to the paths that were requested to be
committed. Long standing issue, that I somehow keep forgetting about. Thanks, @aqw for making an issue about it ;-)

Closes #571

### DIFF
--- a/onyo/lib/git.py
+++ b/onyo/lib/git.py
@@ -190,8 +190,9 @@ class GitRepo(object):
         """
         if isinstance(paths, Path):
             paths = [paths]
-        self._git(['add'] + [str(p) for p in paths])
-        self._git(['commit', '-m', message])
+        pathspecs = [str(p) for p in paths]
+        self._git(['add'] + pathspecs)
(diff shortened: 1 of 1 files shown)


aqw (approved):
Oh that's really cool. It makes sense, but I never realized that you could commit just a portion of what you have staged. :exploding_head: 

LGTM!

PR #577, opened 2024-03-10
Fix help on unrecognized arguments

### LINKED ISSUE #356: Usage message for subcommand not useful
```
❱ onyo get newgroup/newuser 
usage: onyo [-h] [-C DIR] [-d] [--version] <command> ...
onyo: error: unrecognized arguments: newgroup/newuser
```

This is reporting general options for onyo rather than how to use `get` (missing the `-p`).

### LINKED ISSUE #408: DOC: help text for `onyo new` needs an example
The correct usage for `onyo new` is not clear.

It took me a bit to realize that `-p` is  necessary to pass a destination directory as an argument.

```
❱ onyo new --edit haha/This\ \&\ That/                                                     
usage: onyo [-h] [-C DIR] [-d] [-v] [-q] [-y] <command> ...
onyo: error: unrecognized arguments: haha/This & That/
```

bpoldrack:
This patch tries to workaround an issue with `argparse`, where an error on unrecognized arguments, is not associated with the selected subparser.
See https://bugs.python.org/issue34479

Solution here is, to rely on `parse_known_args` instead of `parse_args`, returning unrecognized arguments in a dedicated list and call the subparsers `print_usage` in addition to reproducing the toplevel parsers' error.

The reason for doing both, is because we can't distinguish `onyo -x new` from `onyo new -x` via argparse (we'd need to reparse `argv` ourselves). Hence, it's not clear in all cases, whether the usage for `new` or `onyo` would be relevant to the user.

So, in the above example this path ends up with:
```
❱ onyo -x new                                                                                                                                                                   1 !
usage: onyo new [-h] [-t TEMPLATE] [-c CLONE] [-e] [-k KEYS [KEYS ...]] [-p PATH] [-tsv TSV] [-m MESSAGE]
usage: onyo [-h] [-C DIR] [-d] [-v] [-q] [-y] <command> ...
onyo: error: unrecognized arguments: -x
```

(exact same for `onyo new -x`)

Still: A lot better than before, I think.

(Closes #356)
(Closes #408)

### DIFF
--- a/onyo/main.py
+++ b/onyo/main.py
@@ -68,6 +68,9 @@ def build_parser(parser, args: dict) -> None:
             parser.add_argument(**args[cmd])
 
 
+subcmds = None
+
+
 def setup_parser() -> argparse.ArgumentParser:
     from onyo.onyo_arguments import args_onyo
     from onyo.commands.cat import args_cat
@@ -85,6 +88,8 @@ def setup_parser() -> argparse.ArgumentParser:
     from onyo.commands.tree import args_tree
     from onyo.commands.unset import args_unset
 
+    global subcmds
+
     parser = argparse.ArgumentParser(
         description='A text-based inventory system backed by git.',
         formatter_class=SubcommandHelpFormatter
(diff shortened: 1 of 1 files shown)


aqw (approved):
I won't call it beautiful, but it solves this sin.

Thank you!

PR #600, opened 2024-03-13
add missing character to fix doc build error

TobiasKadelka:
There was a missing character that let to a failure of the doc-build which I added.


aqw (approved):
Thanks for catching this!

PR #601, opened 2024-03-13
DOC/BF: stylize help output with `rich`

### LINKED ISSUE #429: DOCS: `onyo --help` uses characters that do not resolve in the CLI
Using `onyo --help` returns documentation including "\`\`ASSETS\`\`\\(s)". It is fine to capitalize them, but the characters do not add to the readability of the help text.

aqw:
This PR:

- adds rich styling to help text
- don't leak RST-isms into CLI help text (fix #429)
- improves and normalize styling across help text 
- improves wording of some help text

This introduces a new dependency: `rich-argparse`

Sadly, there is no `rich.RST` analogue to `rich.Markdown`, so I intercept the help text, and convert relevant RST-isms into rich markup. It's not elegant, but functional.

Annoyingly, rich-argparse passes around rich `Text` objects rather than strings, which lacks basic string functionality. Thus, I need to intercept text at all three originating points, rather than the single rendering point.

If we want to remove the dependency, we could lift some pieces from the `rich-argparse` project (MIT licensed), and integrate our changes into them (flattening the structure by a layer).

### DIFF
--- a/onyo/commands/get.py
+++ b/onyo/commands/get.py
@@ -20,8 +20,8 @@ args_get = {
         required=False,
         default=0,
         help="""
-            Descend up to DEPTH levels into the directories specified. DEPTH=0
-            descends recursively without limit.
+            Descend up to **DEPTH** levels into the directories specified. A
+            depth of **0** descends recursively without limit.
         """
     ),
 
@@ -30,7 +30,7 @@ args_get = {
         metavar='KEY',
         nargs='+',
         help="""
-            KEY values to print. Pseudo-keys (information not stored in the
+            **KEY** values to print. Pseudo-keys (information not stored in the
             asset file) are also available for queries.
         """
     ),
@@ -51,9 +51,10 @@ args_get = {
         type=str,
         default=None,
         help="""
-            Criteria to match assets in the form ``KEY=VALUE``, where VALUE is a
-            python regular expression. Pseudo-keys such as ``path`` can also be
-            used. Special values supported are:
+            Criteria to match assets in the form ``KEY=VALUE``, where **VALUE**
+            is a python regular expression. Pseudo-keys such as ``path`` can
+            also be used. Special values supported are:
+
               * ``<dict>``
               * ``<list>``
               * ``<unset>``
@@ -66,7 +67,7 @@ args_get = {
         type=path,
         nargs='+',
         help="""
-            PATHs to assets or directories to query.
+            **PATH**\s to assets or directories to query.
         """
     ),
 
@@ -75,7 +76,7 @@ args_get = {
         action='store_true',
         default=False,
         help="""
-            Sort output in ascending order (excludes --sort-descending).
+            Sort output in ascending order (excludes ``--sort-descending``).
         """
     ),
 
@@ -84,7 +85,7 @@ args_get = {
         action='store_true',
         default=False,
         help="""
-            Sort output in descending order (excludes --sort-ascending).
+            Sort output in descending order (excludes ``--sort-ascending``).
         """
     ),
 }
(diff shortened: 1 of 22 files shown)


onyo/commands/get.py
```diff
@@ -51,9 +51,10 @@
         type=str,
         default=None,
         help="""
-            Criteria to match assets in the form ``KEY=VALUE``, where VALUE is a
-            python regular expression. Pseudo-keys such as ``path`` can also be
-            used. Special values supported are:
+            Criteria to match assets in the form ``KEY=VALUE``, where **VALUE**
```
TobiasKadelka, line 54: > ``KEY=VALUE``, where **VALUE**

I am never sure about the english comma rules. Is that one here correct?

aqw: Yes, this looks correct to me.

onyo/commands/get.py
```diff
@@ -84,21 +85,21 @@
...
+    Return values of the requested **KEY**\s for matching assets.
 
-    If no KEYs are given, all keys in the asset name are used (see
-    ``onyo.assets.filename``). If no PATHs are given, the current working
+    If no **KEY**\s are given, all keys in the asset name are printed (see
+    ``onyo.assets.filename``). If no **PATH**\s are given, the current working
     directory is used.
 
-    In addition to keys in asset contents, PSEUDO-KEYS can be queried and
+    In addition to keys in asset contents, **PSEUDO-KEYS** can be queried and
```
TobiasKadelka, line 102: Here it says `**PSEUDO-KEYS**` capitalized and bold-highlighted.
In the help-text for `'-M', '--match'` it is not highlighted this way.

onyo/commands/set.py
```diff
@@ -60,13 +60,13 @@
...
 
-    KEY names can be any valid YAML key-name. If a KEY does not exist, it is
-    added and set appropriately.
+    **KEY** names can be any valid YAML key-name. If a key is not present in an
+    asset, it is added and set appropriately.
 
-    Setting KEYs used in the asset name require the ``--rename`` flag, and will
-    result in the asset being renamed on the filesystem.
+    Setting **KEY**\s that are used in the asset name requires the ``--rename``
+    flag, and will result in the asset being renamed on the filesystem.
```
TobiasKadelka, line 69: Nitpick: maybe instead of
"... and will result in the asset being renamed on the filesystem." 
->
"... and will result also in the asset file or directory being renamed."
Or something similar. It is a bit more clunky naming both types of assets, but I find the usage of "filesystem" a bit overly precise, because we never use that word otherwise and it sounds to me like it happens "just" on the file-system (i.e. as if we opose it to something different that would normaly happen) and not in git or in the asset contents?
I am struggling to formulate my concern, but the word just does not feel that good to me, since we never highlight the "filesystem" part in other scenarios.

Feel free to ignore this comment if it sounds to you like I am just too tired, please!

onyo/commands/unset.py
```diff
@@ -20,7 +20,7 @@
         nargs='+',
         type=str,
         help="""
-            KEYs to unset in assets. Multiple keys can be given
+            Keys to unset in assets. Multiple keys can be given
             (e.g. key1 key2 key3).
```
TobiasKadelka, line 24: For onyo set the example for multiple keys was ticks-quoted in this stile: "(e.g. ``key1 key2 key3``)."

onyo/commands/unset.py
```diff
@@ -32,7 +32,7 @@
         nargs='+',
         type=path,
         help="""
-            Assets unset KEYs in.
+            Assets unset **KEY**s in.
```
TobiasKadelka, line 35: "Assets *to* unset ..." with a "to"?

TobiasKadelka (approved):
Thanks a lot for re-working the texts and adding colors, whitespace and much more!

I left some comments, but none of them are blocking a merge for me, since I really want to see the changes merged and then review the new help texts and documentation in action! :)

onyo/commands/set.py
aqw: I see your point. The language could certainly be tighter. I will simplify this to:

> Setting **KEY**\s that are used in the asset name requires the ``--rename`` flag.

aqw: Fixed and pushed.

onyo/commands/unset.py
aqw: Fixed and pushed.

onyo/commands/unset.py
aqw: Fixed and pushed.

Issue #591, opened 2024-03-12
`xargs` usage in demo script does not work on macOS


TobiasKadelka:
```
❱ onyo get --match type=laptop -H --keys path | xargs -d "\n" onyo -y set --keys USB_A=2 --path

xargs: invalid option -- d
usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements] [-S replsize]]
             [-J replstr] [-L number] [-n number [-x]] [-P maxprocs]
             [-s size] [utility [argument ...]]
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
BrokenPipeError: [Errno 32] Broken pipe
```

TobiasKadelka:
Notes:
1. Currently working on both systems would be:
   `onyo set --keys USB_A=2 --path $(onyo get --match type=laptop -H --keys path)`
2. In the future this solution will also work (a.k.a. like above, but without xargs) 
   `onyo get --match type=laptop -H --keys path | onyo set --keys USB_A=2`

Once functioning and tested both variants should be described under https://onyo.readthedocs.io/en/latest/

aqw:
The fix was merged in #602

Closing. :-)

TobiasKadelka:
My plan was to first verify if it now actually does work on macOS or something else would happen :P
And yes, works just like intended.

PR #602, opened 2024-03-13
Machine-readable `get` quotes values

bpoldrack:
This is changing the output of `onyo_get` when used in machine-readable mode to quote the values it prints. Hence, whitespaces in values aren't an issue (especially when piping `onyo get` into something else.

Demo generation script is adjusted accordingly, not requiring the `-d "\n"` option for `xargs` anymore, when reading `onyo get`'s output to construct the `onyo set` call.


onyo/commands/tests/test_get.py
```diff
@@ -152,6 +152,7 @@ def test_get_all(
 
     if machine_readable:
         for line in output:
+            line = [field.strip("\"") for field in line]
```
TobiasKadelka, line 155: IIRC you can use in cases like this the single quotes for strings, so as to not escape the double-quotes: `field.strip('"')`.
Obviously it is a matter of taste which one any given person finds more readable :)

TobiasKadelka (approved):
You missed to highlight the most important part of this PR: It should allow the demo-script to run on macOS again ;)
The `-d "\n"` option of xargs broke it.

Hence ping-worthy for https://github.com/psyinfra/onyo/issues/591

Issue #592, opened 2024-03-12
`onyo get --match directory=shelf/` does not interprete directory as alternative to `--path`


TobiasKadelka:
- `onyo new --keys directory=shelf` is equivalent to `onyo new --path shelf`
- `onyo get --match directory=shelf` is **not** equivalent to `onyo get --path shelf`

-> All commands should interprete the usage of a key "directory" the same.

-------
Here a scenario where I find the behavior un-intuitive and wrong:

```
❱ ls warehouse 
headphones_JBL_pro.325gtt      laptop_apple_macbookpro.0io4ff laptop_lenovo_thinkpad.iu7h6d
laptop_apple_macbook.9r5qlk    laptop_apple_macbookpro.1eic93

❱ onyo get --match type=laptop directory=warehouse -H 
# prints nothing (because directory is not a key inside the asset?)

❱ onyo get --match type=laptop --path warehouse -H                                                                          130 !
laptop	apple	macbook	9r5qlk	warehouse/laptop_apple_macbook.9r5qlk
laptop	apple	macbookpro	0io4ff	warehouse/laptop_apple_macbookpro.0io4ff
laptop	apple	macbookpro	1eic93	warehouse/laptop_apple_macbookpro.1eic93
laptop	lenovo	thinkpad	iu7h6d	warehouse/laptop_lenovo_thinkpad.iu7h6d
```

aqw:
Part of this is resolved now that #603 is merged.

IMO, `directory` should not be an alternative to `path`. It should be the parent directory of the asset (which is how it is used in `onyo new`).

The proposal is for `directory` to be promoted form a Reserved Key (only used by `new`) to a Pseudo Key that's queryable like any other key.

Issue #583, opened 2024-03-11
`onyo get` rich table is trimmed when output is wider than the terminal


aqw:
![image](https://github.com/psyinfra/onyo/assets/765557/077a6f4c-eeb0-48e3-aecf-d24511b5f5e4)

And super oddly, it trims from the left.

It's not a problem with `-H`.

Issue #579, opened 2024-03-10
`onyo get` always returns `path` even when not requested


aqw:
I agree with the decision to include `path` in the default output for `onyo get`. But I disagree with the decision to /always/ include `path` in the output, even when not requested. If I as the user don't want it, I cannot be rid of it.

In my opinion, it's kinder to the user to let them request it when they wish.

```
❱ onyo get --keys type --path accounting
 ─────────────────────────────────────────────────────────────────── 
  type         path                                                  
 ─────────────────────────────────────────────────────────────────── 
  headphones   accounting/Bingo Bob/headphones_apple_airpods.uzl8e1  
  laptop       accounting/Bingo Bob/laptop_apple_macbook.oiw629      
  monitor      accounting/Bingo Bob/monitor_dell_PH123.86JZho        
 ─────────────────────────────────────────────────────────────────── 
 ```

Issue #578, opened 2024-03-10
`onyo get`  returns `<unset>` rather than `false` for `is_asset_directory`


aqw:
In my opinion, `onyo get` is misleading by returning `<unset>` rather than `false` for asset files.

The pseudo key isn't unset, rather we /know/ that indeed this asset is not an Asset Directory.

```
❱ onyo get --keys is_asset_directory --path accounting
 ─────────────────────────────────────────────────────────────────────────── 
  is_asset_directory   path                                                  
 ─────────────────────────────────────────────────────────────────────────── 
  <unset>              accounting/Bingo Bob/laptop_apple_macbook.oiw629      
  <unset>              accounting/Bingo Bob/monitor_dell_PH123.86JZho        
  True                 accounting/Bingo Bob/headphones_apple_airpods.uzl8e1  
 ─────────────────────────────────────────────────────────────────────────── 
```

PR #607, opened 2024-03-15
BF: `onyo get` fixes

aqw:
This PR fixes the following bugs:

- #579 
- #583
- #578
- #592

As a brief digression... the test `test_onyo_get_reserved_keys()` ended up running into bug #583. Printing `path` twice exceeded the 80 char limit that Rich detects when running in a virtual TTY.

I looked into how to increase the faux-terminal width, but I could not find a reasonable way. [This rich bug](https://github.com/Textualize/rich/issues/317) was infuriatingly closed with no explanation, and the rich docs make no mention of `capsys`, and only `pytest` once...

Other suggestions involve [monkey patching `Popen`](https://stackoverflow.com/questions/60272369/pytest-monkeypatch-terminal-size), but it wasn't clear to me the right way to mock this through fixtures.

And at that point... I decided that if we don't want output to be interfered with, then we should use `machine-readable=True`.

### DIFF
--- a/onyo/commands/get.py
+++ b/onyo/commands/get.py
@@ -93,9 +93,9 @@ def get(args: argparse.Namespace) -> None:
     """
     Return values of the requested **KEY**\s for matching assets.
 
-    If no **KEY**\s are given, all keys in the asset name are printed (see
-    ``onyo.assets.filename``). If no **PATH**\s are given, the current working
-    directory is used.
+    If no **KEY**\s are given, the path and all keys in the asset name are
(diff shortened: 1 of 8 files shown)


bpoldrack (commented):
Overall LGTM.

One thing, though: Is there a reason to not add `directory` to `PSEUDO_KEYS`?
Which leads to the next question: I'm not sure why there aren't test failures. A bunch of assertions should differ when a key is or isn't a `PSEUDO_KEY`. I would have expected some things to fail when a key is added like that w/o being put into that constant. So - something about that doesn't feel quite right yet. Could be due to insufficient tests, though.

Issue #322, opened 2023-04-03
Caching using `functools.cached_property`


nhjjr:
Currently we're caching certain properties in the `Repo` class by running a setter when a getter is first called, and then skipping the setter for future calls. The cache for that property is cleared by setting it to `None` (e.g., via a `clear_caches` function).

The [functools](https://docs.python.org/3/library/functools.html) `@cached_property` decorator is made for this. This would look like, for example:

```Python3
from functools import cached_property
from typing import Generator

class Repo:
    @cached_property
    def assets(self) -> Generator[Path]:
        return (Path(a) for a in self.get_assets()) # imagine the retrieval code here

    def clear(self, attrs: Union[str, list[str]]) -> None:
        if isinstance(attrs, str):
            attrs = [attrs]

        for attr in attrs:
            delattr(self, attr)
```

If we call the assets attribute we first execute the code in the return line, but on subsequent calls this is retrieved from the cache. If we want to clear the cache, we can delete the attribute using `repo.clear('assets')`. This removes the contents from memory, but crucially the attribute is still accessible and will simply execute the underlying code again and store it in cache when it is called again.

Note that the `clear` function can have a list of "clearable_attrs" in case we want to restrict its use. We could also default `attrs` to `None`, and clear all attributes in the class instance that aren't routines.

The functools docs lists a few restrictions on cached properties, so we should make sure that those restrictions do not apply prior to its implementation and test accordingly.

aqw:
I just discovered this issue. Cool find! I did not know about `@cached_property`.

> Useful for expensive computed properties of instances that are otherwise effectively immutable.

Sounds promising indeed. If there's a common way of doing this, I like the idea of following that path.

However, as I read into this, I see one primary blocker:

> The mechanics of [cached_property()](https://docs.python.org/3/library/functools.html#functools.cached_property) are somewhat different from [property()](https://docs.python.org/3/library/functions.html#property). A regular property blocks attribute writes unless a setter is defined. In contrast, a cached_property allows writes.

Which sounds like a silly default for something that's "effectively immutable"

> Similar to [property()](https://docs.python.org/3/library/functions.html#property), with the addition of caching.

Which makes me wonder if we can use `@<property>.setter` like we can with `@property`:

```
    @files.setter
    def files(self, value) -> None:
        raise AttributeError("can't set attribute")
```

But that fails, `AttributeError: 'cached_property' object has no attribute 'setter'`. So it's not as similar to `@property` as I hoped.

[...]

PR #610, opened 2024-03-20
BF/DOC: unify syntax of `OnyoRepo` and `GitRepo` `.clear_cache()`

aqw:
I noticed these while looking into #322

This unifies the approach, and updates the doc strings.


TobiasKadelka:
Thanks, looks good to me; I will merge! :)

PR #611, opened 2024-03-21
NF/DOC: generate Python API docs

aqw:
This PR is a bit bigger than I intended, but some cleanup and reorganization was necessary (and/or convenient) in order to build the docs warning free.

Major changes include:
- no longer re-export at `onyo` and `onyo.lib`; things live where they are
- `onyo/commands` -> `onyo/cli`
- remove redundant type information from docstrings (now built automatically from Type Hints)
- generate docs for Onyo's Python API


docs/source/onyo.rst
```diff
@@ -0,0 +1,27 @@
+Python API
+============
```
TobiasKadelka, line 2: Nitpick: length of the `===` does not align with Header :)

onyo/lib/commands.py
```diff
@@ -87,10 +87,10 @@ def fsck(repo: OnyoRepo,
 
     Parameters
     ----------
-    repo: OnyoRepo
+    repo
         The Repository on which to perform the fsck on.
 
-    tests: list of str, optional
```
TobiasKadelka, line 93: I think @bpoldrack pointed out a while ago that the meaning of "optional" in docstrings does not 100% align with the meaning of type hints.

IIRC In docstrings it means "is not required to be given" while optional as typehint means "must be given but can be `None`". But if not given, python errors with "argument missing" if not e.g. a default value is specified.

I could imagine that this is not concidered correctly here when auto-generating types?

onyo/lib/ui.py
```diff
@@ -26,14 +26,14 @@ class UI(object):
 
     Attributes
     ----------
-    log: Logger
+    logger
```
TobiasKadelka, line 29: This should be "log", not "logger", based on the variable in line 9 :)

TobiasKadelka (commented):
This looks really good to me!

The only thing worth considering besides my inline comments is that typehints were applied inconsistently by different developers along files. So whenever something was not specified, they would miss in the documentation.

It might be worth to invest a few minutes into adding them where they are missing but clear, or at least running `pyre -i` (which should figure out some typehints automatically) and select the ones that fit the code.

docs/source/onyo.rst
aqw: Yeah, that's due to using `ex` to replace the header in the first line.

But it's a good excuse for me to improve my `ex`-fu. I'll fix this.

onyo/lib/ui.py
aqw: Interesting about line 9. That seems to be unused in the rest of the file.

But this indeed should be `logger`.  `log()` is a function, and `self.logger` is an attribute of the `UI` class. And everything uses that.

docs/source/onyo.rst
aqw: Fixed and pushed. :-)

onyo/lib/commands.py
aqw: Correct. `Optional[]` means "or type `None`". The use of default values means that the parameter is optional.

In the docs, default values are rendered:
- in the function header
- in the type description of the parameter (I just enabled this)
- in the parameter description (written manually when it deviates from the declared default, such as with mutable types)

Even though that doesn't use the word "optional", I think it's pretty clear what's going on.

aqw:
> This looks really good to me!
> 
> The only thing worth considering besides my inline comments is that typehints were applied inconsistently by different developers along files. So whenever something was not specified, they would miss in the documentation.
> 
> It might be worth to invest a few minutes into adding them where they are missing but clear, or at least running `pyre -i` (which should figure out some typehints automatically) and select the ones that fit the code.

Thanks for the review!

I checked through all of the methods that I modified, and confirmed that all already had sufficient type hints. We did not lose any information there.

However, I did find that the types of class attributes aren't automatically detected (which makes sense). So I restored those.

I have fixed and pushed everything. IMO, it is good to go.

There any many errors that remain in the docs, but none that are introduced by this PR. Fixing other preexisting issues will come in future PRs.

PR #612, opened 2024-03-26
DOC/TST: assorted flake8 and docstring improvements

aqw:
This PR is eclectic, but it all relates.

- brings comments into the flake8 config
- escapes "\`\`ASSET\`\`\ s" consistently (needed for Python 3.12; flake8; unknown future Python version)
- converts all docstrings and help text into raw strings
- retires one ignored rule from flake8 (thanks to the above fixes)
- bumps RTD to Python 3.12 (thanks to the above fixes)
- makes sphinx render docstrings for `__init__` and `__call__`
- improves the docstring for `StoreKeyValuePairs`
- improves the error message generated by `StoreKeyValuePairs`
- adds a test for conflicting number of keys passed to `onyo new`

### DIFF
--- a/onyo/argparse_helpers.py
+++ b/onyo/argparse_helpers.py
@@ -61,27 +61,3 @@ class StoreKeyValuePairs(argparse.Action):
                 d[k] = cvt(v)
             results.append(d)
         setattr(namespace, self.dest, results)
-
-
-def parse_key_values(string):
-    r"""
-    Convert a string of key-value pairs to a dictionary.
-
-    The shell interprets the key-value string before it is passed to argparse.
-    As a result, no quotes are passed through, but the chunking follows what the
-    quoting declared.
-
-    Because of the lack of quoting, this function cannot handle a comma in
-    either the key or value.
-    """
-    results = {k: v for k, v in (pair.split('=') for pair in string.split(','))}
-    for k, v in results.items():
-        try:
-            results.update({k: int(v)})
-        except ValueError:
-            try:
-                results.update({k: float(v)})
-            except ValueError:
-                pass
-
-    return results
--- a/onyo/argparse_helpers.py
+++ b/onyo/argparse_helpers.py
@@ -1,7 +1,11 @@
+from __future__ import annotations
+
 import argparse
 import os
+from typing import TYPE_CHECKING
 
-from typing import Sequence
+if TYPE_CHECKING:
+    from typing import Sequence
 
 
 class StoreKeyValuePairs(argparse.Action):
@@ -52,15 +56,16 @@ class StoreKeyValuePairs(argparse.Action):
         for kv in key_values:
             if "=" not in kv:
                 parser.error(f"Invalid argument '{kv}'. Expected key-value pairs '<key>=<value>'.")
+
         pairs = [p.split('=', maxsplit=1) for p in key_values]
-        register_dict = {k: [] for k, v in pairs}
-        [register_dict[k].append(v) for k, v in pairs]
-        number_of_dicts = max(len(v) for v in register_dict.values())
-        invalid_keys = [(k, len(v)) for k, v in register_dict.items() if 1 < len(v) < number_of_dicts]
-        if invalid_keys:
-            parser.error(f"All keys given multiple times must be provided the same number of times."
-                         f"Max. times a key was given: {number_of_dicts}.{os.linesep}"
-                         f"But also got: {', '.join(['{} {} times'.format(k, c) for k, c in invalid_keys])}")
+        key_lists = {k: [] for k, v in pairs}
+        [key_lists[k].append(v) for k, v in pairs]
+
+        key_counts = {k: len(v) for k, v in key_lists.items()}
+        key_max_count = max(key_counts.values())
+        if any([True for k, c in key_counts.items() if 1 < c < key_max_count]):
+            parser.error(f"All keys given multiple times must be given the same number of times:{os.linesep}"
+                         f"{f'{os.linesep}'.join(['{}: {}'.format(k, c) for k, c in key_counts.items() if 1 < c])}")
 
         def cvt(v: str) -> int | float | str | bool:
             if v.lower() == "true":
@@ -77,9 +82,9 @@ class StoreKeyValuePairs(argparse.Action):
             return r
 
         results = []
-        for i in range(number_of_dicts):
+        for i in range(key_max_count):
             d = dict()
-            for k, values in register_dict.items():
+            for k, values in key_lists.items():
                 v = values[0] if len(values) == 1 else values[i]
                 d[k] = cvt(v)
             results.append(d)
--- a/onyo/argparse_helpers.py
+++ b/onyo/argparse_helpers.py
@@ -18,7 +18,7 @@ class StoreKeyValuePairs(argparse.Action):
                  namespace: argparse.Namespace,
                  key_values: list[str],
                  option_string: str | None = None) -> None:
-        """Turn a list of 'key=value' pairs into a list of dictionaries
+        r"""Turn a list of 'key=value' pairs into a list of dictionaries
 
         Every key appearing multiple times in `key=value` is applied to a new dictionary every time.
         All keys appearing multiple times, must appear the same number of times (and thereby define the number of dicts
@@ -64,7 +64,7 @@ class StoreKeyValuePairs(argparse.Action):
 
 
 def parse_key_values(string):
-    """
+    r"""
     Convert a string of key-value pairs to a dictionary.
 
     The shell interprets the key-value string before it is passed to argparse.
--- a/onyo/argparse_helpers.py
+++ b/onyo/argparse_helpers.py
@@ -5,11 +5,22 @@ from typing import Sequence
 
 
 class StoreKeyValuePairs(argparse.Action):
+    r"""Store a list of dictionaries of key-value pairs.
+    """
+
     def __init__(self,
                  option_strings: Sequence[str],
                  dest: str,
                  nargs: int | str | None = None,
                  **kwargs) -> None:
+        r"""
+        Parameters
+        ----------
+        option_strings
+        dest
+        nargs
+        **kwargs
+        """
         self._nargs = nargs
         super().__init__(option_strings, dest, nargs=nargs, **kwargs)
 
@@ -18,12 +29,24 @@ class StoreKeyValuePairs(argparse.Action):
                  namespace: argparse.Namespace,
                  key_values: list[str],
                  option_string: str | None = None) -> None:
-        r"""Turn a list of 'key=value' pairs into a list of dictionaries
+        r"""Turn a list of 'KEY=VALUE' pairs into a list of dictionaries.
 
-        Every key appearing multiple times in `key=value` is applied to a new dictionary every time.
-        All keys appearing multiple times, must appear the same number of times (and thereby define the number of dicts
-        to be created). In case of different counts: raise.
-        Every key appearing once in `key_values` will be applied to all dictionaries.
+        Each KEY can be defined either 1 or N times (where N is the number of
+        dictionaries to be created).
+
+        A KEY that is declared once will apply to all dictionaries.
+
+        All KEYs appearing N times must appear the same number of times. If not,
+        a message will print to standard error and the program will exit with
+        status code 2.
+
+        Parameters
+        ----------
+        parser
+        namespace
+        key_values
+            List of strings containing key-value pairs.
+        option_string
         """
 
         for kv in key_values:
--- a/onyo/cli/cat.py
+++ b/onyo/cli/cat.py
@@ -20,7 +20,7 @@ args_cat = {
 
 
 def cat(args: argparse.Namespace) -> None:
-    """
+    r"""
     Print the contents of **ASSET**\ s to the terminal.
 
     If any of the paths are invalid, then no contents are printed and an error
--- a/onyo/cli/cat.py
+++ b/onyo/cli/cat.py
@@ -21,7 +21,7 @@ args_cat = {
 
 def cat(args: argparse.Namespace) -> None:
     """
-    Print the contents of **ASSET**\s to the terminal.
+    Print the contents of **ASSET**\ s to the terminal.
 
     If any of the paths are invalid, then no contents are printed and an error
     is returned.
--- a/onyo/cli/tests/test_new.py
+++ b/onyo/cli/tests/test_new.py
@@ -284,6 +284,28 @@ def test_keys_flag(repo: OnyoRepo, directory: str) -> None:
     assert repo.git.is_clean_worktree()
 
 
+def test_error_keys_flag_mismatch_count(repo: OnyoRepo) -> None:
+    r"""
+    Test that `onyo new --keys KEY=VALUE` requires that repeated keys are given
+    the same number of times.
+    """
+    key_values = asset_spec + ['serial=1', 'mode=0', 'mode=1', 'mode=2']
+
+    # create asset with --keys
+    ret = subprocess.run(
+        ['onyo', '--yes', 'new', '--keys'] + key_values,
+        capture_output=True, text=True)
+
+    # verify correct error
+    assert not ret.stdout
+    assert "All keys given multiple times must be given the same number of times" in ret.stderr
+    assert ret.returncode == 2
+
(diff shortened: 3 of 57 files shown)


onyo/argparse_helpers.py
```diff
@@ -18,26 +32,39 @@ def __call__(self,
...
-            parser.error(f"All keys given multiple times must be provided the same number of times."
-                         f"Max. times a key was given: {number_of_dicts}.{os.linesep}"
-                         f"But also got: {', '.join(['{} {} times'.format(k, c) for k, c in invalid_keys])}")
+        key_lists = {k: [] for k, v in pairs}
+        [key_lists[k].append(v) for k, v in pairs]
+
+        key_counts = {k: len(v) for k, v in key_lists.items()}
+        key_max_count = max(key_counts.values())
+        if any([True for k, c in key_counts.items() if 1 < c < key_max_count]):
+            parser.error(f"All keys given multiple times must be given the same number of times:{os.linesep}"
```
TobiasKadelka, line 66: Seeing this, I feel like `parser.error()` should possibly replaced with `ui.error()` -- except obviously if it does something special and there is a reason for the usage of this function.

At least, when I began to apply `ui.error()` and `ui.print()` throughout the code, I just looked for normal prints, so it might have slipped through earlier.

onyo/cli/cat.py
```diff
@@ -20,8 +20,8 @@
 
 
 def cat(args: argparse.Namespace) -> None:
-    """
-    Print the contents of **ASSET**\s to the terminal.
+    r"""
+    Print the contents of **ASSET**\ s to the terminal.
```
TobiasKadelka, line 24: I know you updated this based on what was there already -- however: I notice quite often the inconsistency between having the docstring begin in the line with the `"""` or in the line after that.

I think Ben pointed out a while ago, that these differences have actually a different meaning, where the first line is used by some documentation tools as a short description; whereas when begin in the second line it is considered part of the "long and complete" description.

That difference should be verified and unified along the code base at some point, though it might be out of scope for this PR and be done whenever as long as the API docs you are building right now can handle it properly :)

onyo/cli/tests/test_new.py
```diff
@@ -284,9 +284,31 @@ def test_keys_flag(repo: OnyoRepo, directory: str) -> None:
     assert repo.git.is_clean_worktree()
 
 
+def test_error_keys_flag_mismatch_count(repo: OnyoRepo) -> None:
+    r"""
+    Test that `onyo new --keys KEY=VALUE` requires that repeated keys are given
+    the same number of times.
+    """
+    key_values = asset_spec + ['serial=1', 'mode=0', 'mode=1', 'mode=2']
```
TobiasKadelka, line 292: I know it is a nitpick, but I find this a bit un-readable, since it is not obvious that "asset_spec" contains also a serial, and that with the added second serial and the three modes the error is triggered.

My first thought was "well, here we have something 1 time (serial) and another thing 3 times (mode) -- this should work just fine". So I would prefer e.g. either having instead of "mode" 2 times "make", or alternatively, instead of "serial" two times something else not in asset_spec.

Either way it is just a small change that would make it easier visible why the error is triggered, when not part is from the `asset_spec` variable and part stated explicitely :)

onyo/cli/tests/test_set.py
```diff
@@ -121,7 +121,7 @@ def test_set_multiple_assets(repo: OnyoRepo,
 @pytest.mark.parametrize('no_assets', non_existing_assets)
 def test_set_error_non_existing_assets(repo: OnyoRepo,
                                        no_assets: list[str]) -> None:
-    """Test that `onyo set KEY=VALUE <asset>` errors correctly for:
+    r"""Test that `onyo set KEY=VALUE <asset>` errors correctly for:
```
TobiasKadelka, line 124: Not caused by you, since it was like this before: but I want to note that it should be `onyo set --keys KEY=VALUE --path <asset>` (path might be called asset now?) because the flags need to be stated explicitely.

This is probably a legacy error from way back when the flags were not required for `onyo set`.

onyo/cli/tests/test_unset.py
```diff
@@ -173,7 +173,7 @@ def test_unset_key_does_not_exist(repo: OnyoRepo,
 @pytest.mark.parametrize('key', ['num'])
 def test_unset_multiple_assets(repo: OnyoRepo,
                                key: str) -> None:
-    """Test that `onyo unset --keys KEY --asset ASSET [ASSET2 ...]` removes keys from of assets."""
+    r"""Test that `onyo unset --keys KEY --asset ASSET [ASSET2 ...]` removes keys from of assets."""
```
TobiasKadelka, line 176: A similar complain as above: are the flags up to date, i.e. is it still `--asset`?
Our doc-strings would deserve a bit more love, especially if they would appear in documentation.

onyo/cli/tests/test_unset.py
```diff
@@ -279,7 +279,7 @@ def test_unset_message_flag(repo: OnyoRepo,
     ["num", "type"]])
 def test_unset_error_unset_name_fields(repo: OnyoRepo,
                                        name_field: list[str]) -> None:
-    """Test that `onyo unset KEY <asset>` throws the correct error without printing the usual
+    r"""Test that `onyo unset KEY <asset>` throws the correct error without printing the usual
```
TobiasKadelka, line 282: "throws" -> "raises" would be probably the better word. (I have a suspection that this was written by myself :P)

onyo/cli/tree.py
```diff
@@ -20,8 +20,8 @@
 
 
 def tree(args: argparse.Namespace) -> None:
-    """
-    List the assets and directories of **DIRECTORY**\s in a tree-like format.
+    r"""
+    List the assets and directories of **DIRECTORY**\ s in a tree-like format.
```
TobiasKadelka, line 24: Again, I know that you are just reformatting, without necessarily reading these all.
But I think in another place, instead of "in a tree-like format" we say "in the tree format"?

onyo/conftest.py
```diff
@@ -240,7 +240,7 @@ def helpers() -> Type[Helpers]:
 
 @pytest.fixture(scope='function', autouse=True)
 def set_ui(request) -> None:
-    """Set up onyo.lib.ui according to a dict provided by the 'ui' marker"""
+    r"""Set up onyo.lib.ui according to a dict provided by the 'ui' marker"""
```
TobiasKadelka, line 243: Trailing `.` missing.

onyo/lib/commands.py
```diff
@@ -575,7 +575,7 @@ def move_asset_or_dir(inventory: Inventory,
 def _maybe_rename(inventory: Inventory,
                   src: Path,
                   dst: Path) -> None:
-    """Helper for `onyo_mv`"""
+    r"""Helper for `onyo_mv`"""
```
TobiasKadelka, line 578: Trailing `.` missing.

onyo/lib/commands.py
```diff
@@ -1039,7 +1039,7 @@ def onyo_set(inventory: Inventory,
 @raise_on_inventory_state
 def onyo_tree(inventory: Inventory,
               dirs: list[tuple[str, Path]]) -> None:
-    """Print the directory tree of paths.
+    r"""Print the directory tree of paths.
```
TobiasKadelka, line 1042: Here is a good example for a different way of phrasing the tree thing along onyo texts.
Is it the directory tree, file tree? tree-like output (like stated by docstring of `def _tree()` below), or "in a tree like format"?

onyo/lib/exceptions.py
```diff
@@ -2,45 +2,45 @@
...
-    """Thrown if a requested operation is a Noop."""
+    r"""Thrown if a requested operation is a Noop."""
     # This is intended to signal that an inventory operation would not result in any change, so that callers can decide
     # on their failure paradigm:
     # "Result oriented already-fine-no-failure" vs "Task oriented can't-do-failure".
 
 
 class NotAnAssetError(Exception):
-    """Thrown if an object was expected to be an asset but isn't"""
+    r"""Thrown if an object was expected to be an asset but isn't"""
```
TobiasKadelka, line 42: Here and for `NotADirError` trailing `.` missing.

onyo/lib/executors.py
```diff
@@ -15,7 +15,7 @@
 
 
 def exec_new_assets(repo: OnyoRepo, operands: tuple) -> tuple[list[Path], list[Path]]:
-    """Executor for the 'new_asset' operation
+    r"""Executor for the 'new_asset' operation
```
TobiasKadelka, line 18: Everywhere in this file the trailing `.` is missing. (At this point I comment this nitpick only because I did it above. Sorry for comment-spam)

onyo/lib/filters.py
```diff
@@ -49,7 +49,7 @@ def _re_match(text: str, r: str) -> bool:
             return False
 
     def match(self, asset: dict) -> bool:
-        """match self on a dictionary"""
+        r"""match self on a dictionary"""
```
TobiasKadelka, line 52: Trailing `.` missing.

onyo/lib/git.py
```diff
@@ -109,7 +109,7 @@ def files(self) -> list[Path]:
         return self._files
 
     def clear_cache(self) -> None:
-        """Clear cache of this instance of GitRepo.
+        r"""Clear cache of this instance of GitRepo.
```
TobiasKadelka, line 112: "GitRepo" should probably be in backticks?

onyo/lib/inventory.py
```diff
@@ -124,15 +124,15 @@ def __init__(self, repo: OnyoRepo) -> None:
 
     @property
     def root(self):
-        """Path to root inventory directory"""
+        r"""Path to root inventory directory"""
```
TobiasKadelka, line 127: Trailing `.` missing everywhere in this file.

onyo/lib/inventory.py
```diff
@@ -124,15 +124,15 @@ def __init__(self, repo: OnyoRepo) -> None:
...
         return self.repo.git.root
 
     def reset(self) -> None:
-        """throw away pending operations"""
+        r"""throw away pending operations"""
```
TobiasKadelka, line 131: Capitalization of "throw".

onyo/lib/inventory.py
```diff
@@ -201,7 +201,7 @@ def _get_pending_asset_names(self) -> list[str]:
         return names
 
     def _get_pending_dirs(self) -> list[Path]:
-        """Gets inventory dirs that would come into existence due to pending operations.
+        r"""Gets inventory dirs that would come into existence due to pending operations.
```
TobiasKadelka, line 204: "Gets" -> "Get"

onyo/lib/onyo.py
```diff
@@ -120,15 +120,15 @@ def set_config(self,
...
+        r"""Get effective value of config `name`.
 
         This is considering regular git-config locations and checks
         `OnyoRepo.ONYO_CONFIG` as fallback.
         """
         return self.git.get_config(name) or self.git.get_config(name, self.git.root / self.ONYO_CONFIG)
 
     def get_asset_name_keys(self) -> list[str]:
-        """Get a list of keys required for generating asset names
+        r"""Get a list of keys required for generating asset names
```
TobiasKadelka, line 131: Trailing `.` missing.

onyo/lib/onyo.py
```diff
@@ -491,7 +491,7 @@ def get_template(self,
         return yaml_to_dict(template_file)
 
     def validate_anchors(self) -> bool:
-        """Check if all dirs (except those in `.onyo/`) contain an .anchor file.
+        r"""Check if all dirs (except those in `.onyo/`) contain an .anchor file.
```
TobiasKadelka, line 494: `.anchor` in backticks?

TobiasKadelka (commented):
Hey, thanks a lot for your PR!

I want to begin with excusing and justifying my behavior ;)
Most things I commented on are not caused by you, they are little nitpicks about inconsistencies in docstrings that were there before.
Thanks to your change with the `r"""` this is the first time that I see so many doc-strings at once throughout onyo, and when I began commenting them I did not expect that these things happen so often.

Long story short: feel free to ignore the stile nitpicks that are out of scope for this PR, IMO it can be merged at any time. I collected some stile inconsistencies in #613, it is no w linked with this PR so the comments are reachable at any time somebody cares to look into it later :)

onyo/argparse_helpers.py
```diff
@@ -1,15 +1,29 @@
...
 import os
+from typing import TYPE_CHECKING
 
-from typing import Sequence
+if TYPE_CHECKING:
+    from typing import Sequence
 
 
 class StoreKeyValuePairs(argparse.Action):
+    r"""Store a list of dictionaries."""
```
TobiasKadelka, line 12: Looking at the example of https://numpydoc.readthedocs.io/en/latest/format.html#short-summary it looks like even one-liner short-summary docstrings have the `"""` in the next line.

onyo/cli/tests/test_new.py
aqw: I agree that it's non-obvious. I was trying to follow the conventions in this file, which I find in general to be non-obvious.

I have been told that there's a desire for the tests to be a similar as possible and not re-define the same things over and over. So I followed that convention.

>  or alternatively, instead of "serial" two times something else not in asset_spec.

Serial needs to be multiple times, because the multiple assets need to be unique.

onyo/cli/tests/test_set.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/cli/tests/test_unset.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/cli/tests/test_unset.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/cli/tree.py
aqw: I like the phrasing "in a tree-like format." For we are formatting it like a tree, and like `tree`.

onyo/lib/exceptions.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/lib/executors.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/lib/filters.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/lib/git.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/lib/inventory.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/lib/inventory.py
aqw: Thanks for the catch. I have changed this to "Discard".

I have fixed it locally, but will not include it in this PR.

onyo/lib/inventory.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/lib/onyo.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/lib/onyo.py
aqw: Thanks for the catch. I have fixed it locally, but will not include it in this PR.

onyo/argparse_helpers.py
aqw: Thanks for the catch. I have committed the fix and pushed.

onyo/argparse_helpers.py
bpoldrack: Nope. This is telling argparse to error out before anything onyo is called. It would report what's wrong with the call itself.

bpoldrack (approved):

Issue #417, opened 2023-10-23
BUG: `onyo set` errors with no message when setting the same key twice in one call


aqw:
```
(onyo) aqw@hrududu in /tmp/yope on git:master
❱ onyo set -k make=yup1 make=yup2 --rename -p dp/Alex\ Waite/laptop_framework_gen1.faux 
ERROR: 
```

bpoldrack:
Note for fixing this: Here is the traceback:

```
ERROR: 
DEBUG:onyo:Traceback (most recent call last):
  File "/home/ben/work/hacking/onyo/onyo/main.py", line 302, in main
    args.run(args)
  File "/home/ben/work/hacking/onyo/onyo/commands/set.py", line 80, in set
    assert len(args.keys) == 1
           ^^^^^^^^^^^^^^^^^^^
AssertionError

```

PR #614, opened 2024-03-27
NF: add StoreSingleKeyValuePair for `onyo set`

### LINKED ISSUE #480: Single dict variant of `StoreKeyValuePair`
#417 (improper failure when giving a key multiple times to `set`) was fixed in PR #477. While this works (note, that only the CLI is affected by the problem),  a better solution is to let argparse fail right away, by providing a  variant of the `StoreKeyValuePair` class, that does not allow for key  repetition and can provide a more detailed failure more easily (what key led to the failure?). The current implementation of that class was designed for  `new` where providing multiple occurences of a key indicate several assets to create. Other commands should not use that exact same Action for their `keys` option.

aqw:
fix #480

Proper fix for #417

### DIFF
--- a/onyo/argparse_helpers.py
+++ b/onyo/argparse_helpers.py
@@ -8,8 +8,12 @@ if TYPE_CHECKING:
     from typing import Sequence
 
 
-class StoreKeyValuePairs(argparse.Action):
+class StoreMultipleKeyValuePairs(argparse.Action):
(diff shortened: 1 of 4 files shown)


TobiasKadelka:
Thanks for the PR, LGTM!

bpoldrack (approved):
Nice, thanks!

PR #617, opened 2024-03-30
DOC/BF: drop `rich_argparse` for home-grown solution

aqw:
This has multiple advantages:

- drops a dependency (that is also unpackaged in Debian)
- `rst_to_rich()` is only used in the parser, thus rich markup no longer leaks into sphinx output
- fixes the "usage" section of sphinx output (which was garbled with CLI colorization syntax)

I also moved the commit from #616 to cleanup `main.py` that I missed in #615 that @TobiasKadelka pointed out.


TobiasKadelka:
This looks very cool!

Issue #492, opened 2023-11-08
Tests + docstring for onyo_config


bpoldrack:
Properly document `onyo/lib/commands.py:onyo_config` and provide tests in `onyo/lib/tests/test_command_config.py`.
Based on that reworked docstring, provide a respective command line version in `onyo/commands/config.py`.

Issue #489, opened 2023-11-08
Tests + docstring for onyo_edit


bpoldrack:
Properly document `onyo/lib/commands.py:onyo_edit` and provide tests in `onyo/lib/tests/test_command_edit.py`.
Based on that reworked docstring, provide a respective command line version in `onyo/commands/edit.py`.

aqw:
The docstrings and help have been updated. This has remained open because there is little in the tests that we're actually satisfied with.

Closing, as there is nothing special about `edit`'s tests with respect to other tests. Further improvements will come as part of refactoring and overall testing improvements.

PR #621, opened 2024-04-09
DOC/TST/NF/RF: another collection of docstrings and general improvements

aqw:
This PR again focuses on docstrings, and then makes adjustments/cleanups that i encountered while focusing on them.

Docstring improvements:
- all of `command_utils.py`
- about 20% of `commands.py`

Non-docstring improvements include:
- `sanitize_args_config()` -> `allowed_config_args()`
- `onyo_edit()`'s behavior has been changed to match other onyo commands
- normalization of type checking imports
- normalization of some help text string declaration

Works towards:
- #489 
- #492 
- #493

### DIFF
--- a/onyo/cli/cat.py
+++ b/onyo/cli/cat.py
@@ -14,7 +14,9 @@ args_cat = {
     'asset': dict(
         metavar='ASSET',
         nargs='+',
-        help='Paths of assets to print'
+        help=r"""
+            Paths of assets to print.
+        """
     ),
 }
 
--- a/onyo/cli/tree.py
+++ b/onyo/cli/tree.py
@@ -14,7 +14,9 @@ args_tree = {
     'directory': dict(
         metavar='DIR',
         nargs='*',
-        help='Directories to list'
+        help=r"""
+            Directories to list.
+        """
     )
 }
 
--- a/onyo/lib/command_utils.py
+++ b/onyo/lib/command_utils.py
@@ -12,13 +12,23 @@ from .onyo import OnyoRepo
 log: logging.Logger = logging.getLogger('onyo.command_utils')
 
 
-def sanitize_args_config(git_config_args: list[str]) -> list[str]:
-    r"""
-    Check the git config arguments against a list of conflicting options. If
-    conflicts are present, the conflict list will be printed and will exit with
-    error.
+def allowed_config_args(git_config_args: list[str]) -> bool:
+    r"""Check a list of arguments for disallowed ``git config`` flags.
 
-    Returns the unmodified  git config args on success.
+    ``git-config`` stores configuration information in a variety of locations.
+    This makes sure that such location flags aren't in the list (and --help).
+
+    A helper for the ``onyo config`` command.
+
+    Parameters
+    ----------
+    git_config_args
+        The list of arguments to pass to ``git config``.
+
+    Raises
+    ------
+    ValueError
+        If a disallowed flag is detected.
     """
     # git-config supports multiple layers of git configuration. Onyo uses
     # ``--file`` to write to .onyo/config. Other options are excluded.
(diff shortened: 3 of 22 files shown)


onyo/cli/cat.py
```diff
@@ -14,7 +14,9 @@
     'asset': dict(
         metavar='ASSET',
         nargs='+',
-        help='Paths of assets to print'
+        help=r"""
+            Paths of assets to print
```
TobiasKadelka, line 18: I know this change is just string-moving here, but comparing with the lines in config.py/edit.py there should be a trailing dot here too.

onyo/cli/tree.py
```diff
@@ -14,7 +14,9 @@
     'directory': dict(
         metavar='DIR',
         nargs='*',
-        help='Directories to list'
+        help=r"""
+            Directories to list
```
TobiasKadelka, line 18: Considering my earlier comment about trailing dots -- now there seem to be two help-texts that have one and two that do not. I think you should decide what is more typical and make it consistent eventually :)

onyo/lib/command_utils.py
```diff
@@ -76,16 +77,16 @@ def fill_unset(assets: Generator[dict, None, None] | filter,
 def natural_sort(assets: list[dict],
                  keys: list[str] | None = None,
                  reverse: bool = False) -> list[dict]:
-    r"""Sort an asset list by a given list of `keys`.
+    r"""Sort an asset list by a list of `keys`.
```
TobiasKadelka, line 80: Not sure about the rule anymore, but is "keys" with one backtick correct, or should it have two, like above in the doc-string of `def fill_unset()`?

onyo/lib/commands.py
```diff
@@ -70,31 +74,29 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
...
-    r"""Run a suite of checks to verify the integrity and validity of an Onyo
-    repository and its contents.
+    r"""Run a suite of integrity checks on an Onyo repository and its contents.
 
-    By default, the following tests will be performed:
+    By default, the following tests are performed:
 
-    * ``clean-tree``: verify that git has no changed (staged or unstaged) or
-      untracked files
     * ``anchors``: verify that all directories (outside of .onyo) have an
```
TobiasKadelka, line 81: `.onyo/` and all its subdirectories have anchor files too. No idea if they are checked for, this might be a bug, but I find this sentence to not reflect the reality of my onyo repository :)

Also, the path `.onyo` should probably be in backticks or double backticks even.

onyo/lib/commands.py
```diff
@@ -137,71 +139,76 @@ def onyo_cat(inventory: Inventory,
...
+    If any path is not an asset, nothing is printed.
+    If any asset content is invalid, the content of all assets is still printed.
 
     Parameters
     ----------
     inventory
         The inventory containing the assets to print.
     paths
-        Path(s) to assets for which to print the contents.
+        Path(s) of assets to print the contents of.
```
TobiasKadelka, line 152: Again, something I am not sure about, but:
`Path(s)` is the way we wrote it in the beginning. Now we often just talk about "assets", "directories", and "paths" in doc-strings (when they are not help-text?).
So while reading this PR it felt a bit like a "surprising" inconsistency here, even though a small one, obviously :P

TobiasKadelka (commented):
Hey, thanks for the PR!
I commented on some minor points that might be small inconsistencies, but overall I am just happy to see things becoming a unified whole picture! :)

onyo/lib/commands.py
aqw: Thanks for pointing this out. I've dropped the `(s)` to make it match everything else.

aqw:
Thanks for the review! I have fixed everything and pushed.

PR #622, opened 2024-08-02
`rm`: Replace modes with simple recursive flag

### LINKED ISSUE #596: replace `onyo rm -a/-d` with better  `onyo rm --recursive`
```
❱ onyo rm --help                                                                                                            130 !
usage: onyo rm [-h] [-a] [-d] [-m MESSAGE] PATH [PATH ...]

Delete ASSETs and/or DIRECTORYs.

Directories and asset directories are deleted along with their contents.

The `--asset` and `--dir` flags can be used to constrain actions to
either assets or directories (respectively).

If any of the given paths are invalid, Onyo will error and delete none of
them.

positional arguments:
  PATH                  Assets and/or directories to delete.

options:
  -h, --help            show this help message and exit
  -a, --asset           Operate only on assets. Asset Files are removed. Asset Directories
                        are converted into normal directories.
                        
                        This cannot be used with the `--dir` flag.
  -d, --dir             Operate only on directories. Directories are removed. Asset
                        Directories are converted into Asset Files.
                        
                        This cannot be used with the `--asset` flag.
  -m MESSAGE, --message MESSAGE
                        Use the given MESSAGE as the commit message (rather than the default).
                        If multiple `-m` options are given, their values are concatenated as
                        separate paragraphs.

```

I find the formulations of `--dir` and `--asset` could be a bit more clear. "Operate only on assets. Asset Files are removed. Asset Directories are converted into normal directories." and "Operate only on directories. Directories are removed. Asset Directories are converted into Asset Files." are of course technically true statements, but I am convinced that the "Operate only ..." part will confuse every single new onyo user (while advanced users probably wont read it) and I think that phrase should either simply be omitted or be "Assets and asset dirs".

It is also missleading in so far that "Operate only on dirs" feels weird to me as a first sentence because it sounds like this is the main effect of this flag -- but instead, if literally changes how `onyo rm` opperates on asset dirs. Instead of deleting them, it just undoes the asset dir aspect.
Why I find this explanation very problematic becomes obvious with this example (using `onyo rm --dir` to delete a dir and an asset dir in one call):
```
❱ onyo rm --dir ethics/Achilles\ Book/laptop_microsoft_surface.oq782j shelf 
The following will be deleted:
-/Users/tkadelka/onyo/ethics/Achilles Book/laptop_microsoft_surface.oq782j
-/Users/tkadelka/onyo/shelf/None_None_None.None
-/Users/tkadelka/onyo/shelf
```

I know that some people will argue "asset dirs are assets" and "asset dirs are directories", and I will argue that explicite help text for commands like this will help new users to understand the "onyo mindset".

bpoldrack:
Simplify the `rm` command, by removing the modes that would have the
command operate on asset or directory 'aspects' only. Instead have a
standard `--recursive` flag, making `onyo rm` aligned with standard
`rm`.

For only removing an aspect of an asset dir, `onyo set` is still
available.

Closes #596

### DIFF
--- a/onyo/cli/rm.py
+++ b/onyo/cli/rm.py
@@ -20,29 +20,13 @@ args_rm = {
         """
     ),
 
-    'assets': dict(
-        args=('-a', '--asset'),
+    'recursive': dict(
+        args=('-r', '--recursive'),
         required=False,
         default=False,
         action='store_true',
         help=r"""
-            Operate only on assets. Asset Files are removed. Asset Directories
-            are converted into normal directories.
-
-            This cannot be used with the ``--dir`` flag.
-        """
-    ),
-
-    'dirs': dict(
-        args=('-d', '--dir'),
-        required=False,
-        default=False,
-        action='store_true',
-        help=r"""
-            Operate only on directories. Directories are removed. Asset
-            Directories are converted into Asset Files.
-
-            This cannot be used with the ``--asset`` flag.
+            If set, remove directories recursively including their content.
         """
     ),
 
@@ -70,25 +54,16 @@ def rm(args: argparse.Namespace) -> None:
     r"""
     Delete **ASSET**\ s and/or **DIRECTORY**\ s.
 
-    Directories and asset directories are deleted along with their contents.
-
-    The ``--asset`` and ``--dir`` flags can be used to constrain actions to
-    either assets or directories (respectively).
+    Directories and asset directories are deleted along with their contents,
+    if the ``--recursive`` flag is set. Otherwise, fails on non-empty directories.
 
     If any of the given paths are invalid, Onyo will error and delete none of
     them.
     """
     inventory = Inventory(repo=OnyoRepo(Path.cwd(), find_root=True))
     paths = [Path(p).resolve() for p in args.path]
-    if args.assets and args.dirs:
-        raise ValueError("'--dir' and '--asset' are mutually exclusive.")
-    mode = "all"
-    if args.assets:
-        mode = "asset"
-    elif args.dirs:
-        mode = "dir"
 
     onyo_rm(inventory,
             paths=paths,
-            mode=mode,  # pyre-ignore[6]  check doesn't understand that this is in fact one of "all", "asset", "dir"
+            recursive=args.recursive,
             message='\n\n'.join(m for m in args.message) if args.message else None)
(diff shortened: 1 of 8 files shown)


onyo/lib/commands.py
```diff
@@ -886,44 +887,27 @@ def onyo_rm(inventory: Inventory,
...
-        One of 'asset', 'dir', or 'all'.
-        In mode 'all' any given path is removed (recursively).
-        In mode 'asset' only paths to assets are accepted. If an asset
-        is in fact an asset dir, this removes the asset aspect of it only,
-        leaving behind a regular inventory dir.
-        In mode 'dir' only paths to dirs are accepted. If a dir happens to
-        be an asset dir, this removes the dir aspect from it, turning it
-        into a regular asset file.
+    recursive
+        Recursively remove a directory with all its content. If not set,
```
aqw, line 891: IMO, this makes sense here to keep the "If not set" description.

onyo/shell_completion/zsh/_onyo
```diff
@@ -133,8 +133,7 @@ _onyo() {
                 args+=(
                     '(- : *)'{-h,--help}'[show this help message and exit]'
                     '(-m --message)'{-m,--message}'[use the given MESSAGE as the commit message]:MESSAGE: '
-                    '(-d --dir -a --asset)'{-d,--dir}'[only remove directories]'
-                    '(-a --asset -d --dir)'{-a,--asset}'[only remove assets]'
+                    '(-r --recursive)'{-r,--recursive}'[recursively remove directories with all their content]'
```
aqw, line 136: Nit: I like making these as terse as possible:
```
recursively remove directories with all their content
```
to
```
remove directories and their contents
```

aqw (approved):
This looks great. Very straightforward. I left a few nits, but nothing that's a blocker. :-D

onyo/shell_completion/zsh/_onyo
bpoldrack: Done

onyo/cli/rm.py
```diff
@@ -20,29 +20,14 @@
...
-        required=False,
-        default=False,
-        action='store_true',
-        help=r"""
-            Operate only on directories. Directories are removed. Asset
-            Directories are converted into Asset Files.
-
-            This cannot be used with the ``--asset`` flag.
+            If set, remove directories recursively including their content.
+            Otherwise fail on non-empty directories.
```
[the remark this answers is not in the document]
bpoldrack: Done

PR #623, opened 2024-08-02
`mkdir`: Allow adding existing non-inventory dirs

### LINKED ISSUE #597: Should `onyo mkdir <dir>` add an anchor to non-onyo dirs in onyo?
```
(valid) tkadelka@bnbmac73 in ~/onyo on git:master
❱ mkdir asdf
(valid) tkadelka@bnbmac73 in ~/onyo on git:master
❱ onyo mkdir asdf
ERROR: /Users/tkadelka/onyo/asdf already exists.
(valid) tkadelka@bnbmac73 in ~/onyo on git:master
❱ onyo rm asdf                                                                                                                1 !
ERROR: Not an inventory directory: /Users/tkadelka/onyo/asdf
```

This is the error currently happening. I created a directory that is not in onyo and has not an `.anchor` file.
Maybe instead of erroring like this, calling `onyo mkdir` on it should add it to onyo if possible?

bpoldrack:
This changes `onyo mkdir` to not fail on an existing non-inventory dir, but make it one instead.

Closes #597

### DIFF
--- a/onyo/cli/tests/test_mkdir.py
+++ b/onyo/cli/tests/test_mkdir.py
@@ -136,13 +136,13 @@ def test_mkdir_quiet_flag(repo: OnyoRepo) -> None:
 def test_dir_exists(repo: OnyoRepo, directory: str) -> None:
     r"""
     Test the correct behavior when `onyo mkdir <path>` is called on an
-    existing directory name.
+    existing directory.
     """
     ret = subprocess.run(['onyo', 'mkdir', directory], capture_output=True, text=True)
 
     # verify output
     assert not ret.stdout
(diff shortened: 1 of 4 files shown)


aqw (approved):
LGTM. Again nice and straightforward.

There's a potential for a lot of weird corner-cases here, that I don't really want to down down the rabbit hole on. But there is one that I think should be checked for:

```
mkdir -p one/two/three
onyo mkdir one/two/three
```

Will this add the `.anchor` to just `three` or all dirs?

bpoldrack:
Good point. Added a test and (small) fix for that.

aqw (approved):
LGTM!

PR #624, opened 2024-08-02
`new`: Allow actual paths for `--template`

### LINKED ISSUE #593: `onyo new --template <path>` does not work with absolute path
`onyo new --template` expects *just* the template name, or alternatively the absolute path. 

```
❱ onyo new --keys directory=shelf --template /Users/tkadelka/onyo/.onyo/templates/laptop.example
The following will be created:
+/Users/tkadelka/onyo/shelf
--- 
[....... rest of git diff ........]


❱ onyo new --keys directory=shelf --template .onyo/templates/laptop.example                                                   1 !
ERROR: Template .onyo/templates/laptop.example does not exist.
```

The relative path should work for `--template` too.

bpoldrack:
Changes `new`'s `--template` parameter to accept actual paths rather than just names. If a path given is relative, it is first looked for underneath the template dir, maintaining the previous behavior when one gives a name of template file therein. Only, if that doesn't exist is a relative path resolved normally (CWD).

Closes #593

### DIFF
--- a/onyo/cli/new.py
+++ b/onyo/cli/new.py
@@ -31,7 +31,10 @@ args_new = {
         metavar='TEMPLATE',
         required=False,
         help=r"""
-            Name of a template to populate the contents of new assets.
+            Path to a template to populate the contents of new assets.
+            Relative paths will first attempt to resolve relative the to the
+            template directory (i.e. just the template name), otherwise it
+            will resolve relative to the current working directory.
 
             This cannot be used with the ``--clone`` flag nor the ``template``
             Reserved Key.
@@ -159,9 +162,17 @@ def new(args: argparse.Namespace) -> None:
         used with the ``--clone`` or ``--template`` flags.
     """
     inventory = Inventory(repo=OnyoRepo(Path.cwd(), find_root=True))
+    if args.template:
+        template = Path(args.template)
+        if not template.is_absolute():
+            probe_template = inventory.root / inventory.repo.TEMPLATE_DIR / template
+            if probe_template.exists():
+                template = probe_template
+    else:
+        template = None
     onyo_new(inventory=inventory,
              directory=Path(args.directory).resolve() if args.directory else None,
-             template=args.template,
+             template=template,
              clone=Path(args.clone).resolve() if args.clone else None,
              tsv=Path(args.tsv).resolve() if args.tsv else None,
              keys=args.keys,
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -675,7 +675,7 @@ def onyo_mv(inventory: Inventory,
 @raise_on_inventory_state
 def onyo_new(inventory: Inventory,
              directory: Path | None = None,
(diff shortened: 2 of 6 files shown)


onyo/cli/new.py
```diff
@@ -31,7 +31,11 @@
         metavar='TEMPLATE',
         required=False,
         help=r"""
-            Name of a template to populate the contents of new assets.
+            Path to a template to populate the contents of new assets.
+            If a relative path is given, it will first tried to be interpreted
```
aqw, line 35: Nit: tried -> try

But I'd like to trim this down a bit. How about something like:

> If a relative path is given, it will first tried to be interpreted as relative to the template directory, allowing to simply give a name. Only if that doesn't exist will the path be resolved as relative to the current working directory.

to

> Relative paths will first attempt to resolve relative the to the template directory (i.e. just the template name), otherwise it will resolve relative to the current working directory.

onyo/lib/commands.py
```diff
@@ -675,7 +675,7 @@ def onyo_mv(inventory: Inventory,
 @raise_on_inventory_state
 def onyo_new(inventory: Inventory,
              directory: Path | None = None,
-             template: str | None = None,
+             template: Path | str | None = None,
```
aqw, line 678: Both `Path` and `str`?

From my read, it'll always come in as `Path` or `None` now.

onyo/lib/commands.py
```diff
@@ -717,8 +717,8 @@ def onyo_new(inventory: Inventory,
         table column.
 
     template
-        The name of a template file in ``.onyo/templates/`` that is copied as a
-        base for the new assets to be created.
+        Path to a template file. If relative, this is allowed to be relative to ``.onyo/templates/``.
```
aqw, line 720: or CWD.

onyo/lib/tests/test_onyo.py
```diff
@@ -141,8 +141,9 @@ def test_is_onyo_path(onyorepo) -> None:
 def test_Repo_get_template(onyorepo: OnyoRepo) -> None:
     """
     The function `OnyoRepo.get_template` returns a dictionary representing
-    a template in `.onyo/templates/*`. Default can be configured via 'onyo.new.template'.
-    With no config and no name given, returns an empty dict.
+    a template file. If a relative path is specified, it will first be looked
```
aqw, line 144: "be looked for underneath" -> "look in"

aqw (approved):
Overall LGTM. Some nits.

onyo/lib/commands.py
bpoldrack: Nope. CWD is for CLI only. We resolve at the outer layer. Things in lib usually expect absolute paths.

onyo/lib/commands.py
bpoldrack: From the CLI layer, yes. But if one is using the python interface directly, `str` for a template name remains valid and is tested for. See no reason to change that (yet).

PR #652, opened 2024-09-25
BF: make sorting case sensitive

### LINKED ISSUE #651: BUG: sorting in `onyo get` is case insensitive
I'm choosing to file this as a bug rather than a feature request.

Sorting with `onyo get` currently is case insensitive:
```
❱ onyo get --keys make -s make
 ────── 
  make  
 ──────
...
  dell
  dell
  Dell
  dell
  dell
...
```
This interferes with getting an accurate list of all unique values for a particular field:
```
❱ onyo get --keys make -s make | uniq                
 ─────────────── 
  make           
 ─────────────── 
  apple          
  ctouch         
  dasgip         
  dell           
  Dell           
  dell           
  delta          
...
 ─────────────── 
```

aqw:
Previously, sorting was case insensitive, so that the words would actually be grouped together. However, within a word, it was not sorted.

`GROUPLETTERS` does not do what we want, but using `LOCALE` does (at least with 'en_US.UTF-8').

For now, I am choosing to use whatever the user has configured on their OS. That makes sense to me, respecting their choice. If this ends up biting us in practice, then we'll simply hardcode it or make it a configurable option in `.onyo/config`.

fix #651

### DIFF
--- a/onyo/lib/command_utils.py
+++ b/onyo/lib/command_utils.py
@@ -96,10 +96,14 @@ def natural_sort(assets: list[dict],
     reverse
         Whether to sort in reverse order.
     """
+    import locale
     import natsort
 
+    # set the locale for all categories to the user’s default setting
+    locale.setlocale(locale.LC_ALL, '')
+
     for key in reversed(keys.keys()):
-        alg = natsort.ns.IGNORECASE | natsort.ns.INT
+        alg = natsort.ns.LOCALE | natsort.ns.INT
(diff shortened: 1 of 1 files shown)


aqw:
RTD failure is unrelated.

aqw:
I filed #653 to unblock that task from this PR.

aqw:
+1 from Ben on Matrix. Merging!

PR #658, opened 2024-10-07
RF: simplify template tab completion

aqw:
I don't quite know what I was trying to achieve here before, but with how we're currently using Onyo, the code seemed needlessly complicated.

Rarely, people report seeing the following when they first use `onyo new --template`:

```
onyo new -t (eval):1: command not found: _template_dir
```

I do not know what causes that, but it so far is unique to `_template_dir`.

These changes simplify it to work like other tab completion of files/dirs.


TobiasKadelka:
LGTM; the changes are small, and tab completion is a part of the code that can't break other basic functionalities, so I will go ahead and merge :)

Issue #659, opened 2024-10-07
Bug: onyo does not preserve `---` headers for templates or assets


aqw:
The `---` that signifies the beginning a YAML file is not retained.

Issue #644, opened 2024-09-09
Provide hint on YAML for editors


bpoldrack:
Either a leading `---` in asset files again (incl. temporary files) or a `.yaml` ending.

TobiasKadelka:
If we decide to go only with `.yaml`, there is a need to modify `onyo cat` in parallel to show the `---` for readability when multiple files are `onyo cat`ed at the same time.

TobiasKadelka:
I continued to reflect on this, and tested this: I tried to manually add the `---` with `onyo edit`, which is legal but of course does not work: 

```
(onyo_new) tkadelka in ~/inventory on git:master
❱ onyo edit spaces/14.6y/3040/display_dell_u2722de.BZPK5P3         
Effective changes:
Accept changes? (y)es / continue (e)diting / (s)kip asset / (a)bort command[Default: yes]y
Save changes? No discards all changes. (y/n) [Default: yes]y
```

It is weird that manually editing does not work as expected by a user, and it destroys the round trip mode we want to have and discussed in #159 and #261.

aqw:
In my tests, the `---` was insufficient to hint `vim` that the file is YAML (though I could swear I've seen that before...). Ending the file in `.yaml` certainly did the trip (and I'm sure will too for other editors).

I do agree that having the document start `---` is helpful, especially for the `cat` usecase. I have opened #659 to track that.

PR #660, opened 2024-10-07
BF: hint to editors to highlight for YAML; retain `---` in assets

aqw:
These changes address two bugs:

1) #644 : temp files now simply have `.yaml` appended, which is an easy enough hack to have editors pickup on the content type and provide some helpful syntax highlighting.
2) #659 : preserve (by force adding) the `---` document start token to YAML files. They will only be added to new assets and existing assets when modified.


TobiasKadelka:
I see that the tests still need work, but I want to already thank you for addressing this!
It is really bugging me, and will greatly improve the cat-ability of multiple assets! :)

aqw:
Ahh, the commit hashes changed due to the new file content. I have updated the reference log, and it should pass now. :-)

TobiasKadelka:
Thanks, LGTM!

PR #666, opened 2024-10-09
TST: add Python 3.13; bump various Action versions

aqw:
This adds the newly release Python 3.13 to the test matrix.

It also bumps the versions of various Actions. None of  them should bring in breaking changes.


TobiasKadelka:
LGTM :)

PR #668, opened 2024-10-10
Support nested dictionaries

### LINKED ISSUE #114: Add dictionary keys and initializing of lists to `onyo set`
There are some functionalities described in the readme, which are not implemented or not tested.

From the old readme (should be added again, together with the implementation):

1.)
> Dictionary keys can be addressed
>  using a `.` (e.g. `parent.child`).

While I am not against this way of addressing dictionary keys, this would block a user from having a dot in key names and update these with `onyo set` in a situation like this:
```
slots.USB: 4
slots.lightning: 2
```

2.)
>  Key values can be initialized as a dictionary or list --- but only as empty.
>  - lists: `key=[list]` or `key=[]`
>  - dictionaries: `key=[dict]` or `key={}`

### LINKED ISSUE #642: Adjust workaround for dict support
13301d2dcd44b99ef8cf860fa322002a901aba55 implemented a workaround for pending dict support.

This may need some adjustment when implementing full (nested) dict support.

In addition the little quirk about the format string replacement needs to be properly documented, including how to go with a "proper" python format string (referencing `asset` as the toplevel dict), which does/should avoid the replacement.

bpoldrack:
Closes #642

I think this closes #114, despite not yet addressing lists. WDYT, @aqw ?


aqw (changes requested):
I am /thrilled/ to see this feature.

Great PR. I left a lot of detailed feedback, most of which is exactly that: feedback and not blockers.

The primary thing I see that I would like to talk about is that keys of dicts are not returned in the key list. This creates some problems, but I would also like to understand what problems it solves.

Otherwise, huge +1 from me.

onyo/lib/filters.py
```diff
@@ -61,6 +61,8 @@ def match(self, asset: dict) -> bool:
                 asset[self.key] is None or asset[self.key] == ''):
             return True
         elif self.key not in asset.keys() or self.value == UNSET_VALUE:
+            # Note: This is weird.
```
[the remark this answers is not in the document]
bpoldrack: Right, this shouldn't be in here, will fix that in a separate PR.
However, to clarify this:

- your second read is correct: The code amounts to an XOR and that's OK, it just reads weird and therefore isn't nice to maintain
- what it handles are two cases:
  - The key we are looking for is not in the asset and we are looking for a value other than `<unset>`. 
  - The key actually is present, but we are looking for `<unset>`.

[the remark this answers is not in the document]
aqw: Ok. The comment referencing the previous `elif` statements made me think that the check was intentionally a complex intersection of the previous statements and current one.

Since the statements each `return`, there's no performance gained by using `elif`. In the PR you said you'll open, I propose changing these to `if` statements, so that one does not need to mentally load all four in order to understand the last one.

onyo/lib/utils.py
```diff
@@ -20,10 +20,125 @@
...
+
+    > d = {'key': 'value', 'nested': {'key': 'another value'}}
+    > wrapper = DotNotationWrapper(d)
+    > wrapper['nested.key']
+    'another value'
+    > list(wrapper.keys())
+    ['key', 'nested.key']
+
+    That implies that iteration considers the flattened view only. In the example above the key `nested`
+    (and its value - the dictionary) will not be yielded when using `wrapper.keys()`, `wrapper.values()`,
```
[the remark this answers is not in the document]
bpoldrack: Not exactly. Such keys are still accessible, they are simply not yielded by iteration.

However, we'll talk about it ...

onyo/lib/utils.py
```diff
@@ -20,10 +20,125 @@
...
+        if isinstance(key, str):
+            parts = key.split('.')
+            effective_dict = self.data
+            if len(parts) > 1:
+                for lvl in range(len(parts) - 1):
+                    try:
+                        effective_dict = effective_dict[parts[lvl]]
+                    except KeyError:
+                        # nested dict doesn't exist yet
+                        effective_dict[parts[lvl]] = dict()
```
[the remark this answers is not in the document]
bpoldrack: > I like thinking about it as a convenience wrapper more than a data type.

Yes, that's what it is.

onyo/lib/commands.py
```diff
@@ -516,6 +517,13 @@ def onyo_get(inventory: Inventory,
                                             exclude=exclude,
                                             depth=depth,
                                             match=match)
+
+    # Problem: We may fill in stuff, that changes the structure. This approach did not consider nested dicts.
```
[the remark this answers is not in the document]
bpoldrack: Your understanding is correct, but this entire comment is no longer relevant in removed in the next push.

onyo/cli/tests/test_set.py
```diff
@@ -62,8 +68,14 @@ def test_set(repo: OnyoRepo,
     assert ret.returncode == 0
 
     # verify changes, and the repository clean
+    file_content = Path.read_text(Path(asset))
     for value in set_values:
-        assert value.replace("=", ": ") in Path.read_text(Path(asset))
+        k, v = value.split('=')
+        if '.' in k:
```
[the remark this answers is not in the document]
bpoldrack: So, this is the one thing that is not addressed in the coming push. Not sure I even parse that right.
What should be looped over? You are literally commenting within a loop.

bpoldrack:
FTR: I don't intend to address the concerns code climate raises in this PR. I think this wrapper implementation will likely see a couple of changes over the coming weeks before it's time to actually simplify it a bit.

onyo/lib/command_utils.py
```diff
@@ -62,29 +58,8 @@ def allowed_config_args(git_config_args: list[str]) -> bool:
     return True
 
 
-def fill_unset(assets: Generator[dict, None, None] | filter,
```
aqw, line 65: Nice to see this evaporate.

onyo/lib/commands.py
```diff
@@ -513,22 +513,39 @@ def onyo_get(inventory: Inventory,
...
+                                                 depth=depth,
+                                                 # pyre's complaint boils down to "Dict" not being "Dict | UserDict".
+                                                 # This is useless nonsense.
+                                                 match=match))  # pyre-ignore[6]
+
+    # convert paths for sorting and printing
     for r in results:
         r['path'] = r['path'].relative_to(inventory.root)
 
+    # Note: Sorting is done before any further filtering/replacing. Therefore, one can sort by keys that aren't actually
```
aqw, line 527: Pure love for this feature.

This would be great information to add to the help text.

onyo/lib/commands.py
```diff
@@ -513,22 +513,39 @@ def onyo_get(inventory: Inventory,
...
+    #       in the output of the command. This behavior is utilized in tests.
     results = natural_sort(
         assets=results,
         # pyre can't tell SORT_ASCENDING is not an arbitrary string but matches the Literal declaration:
         keys=sort or {'path': SORT_ASCENDING})  # pyre-ignore[6]
 
-    # filter output for `keys` only
-    results = [{k: v for k, v in r.items() if k in selected_keys} for r in results]
+    # Filter results for `selected_keys` first in order to not iterate over irrelevant parts of the assets in subsequent
+    # filters, replacements, etc.
```
aqw, line 535: Replacements makes sense, but filters, we don't filter any later. And if we did, this would be a bug here as onyo allows filtering (matching) on fields that aren't output.

onyo/lib/commands.py
```diff
@@ -513,22 +513,39 @@ def onyo_get(inventory: Inventory,
...
+    results = [{k: r[k] if k in r else UNSET_VALUE
+                for k in selected_keys}
+               for r in results]
+    # Note: We now have a list of regular dicts. DotNotation would be resolved at this point by having keys with
```
aqw, line 539: > DotNotation would be resolved at this point ...

I don't quite parse this.

In general, I don't like the word "would" in comments. I really prefer stating what /is/ going on. We don't allow literal dots either. In effect, this comment increases my confusion rather than reducing it.

onyo/lib/commands.py
```diff
@@ -513,22 +513,39 @@ def onyo_get(inventory: Inventory,
...
+    #       literal dots.
+
+    # Replace structures by an indication of type.
```
aqw, line 542: Nit: "by" -> "with"

onyo/lib/commands.py
```diff
@@ -513,22 +513,39 @@ def onyo_get(inventory: Inventory,
...
+    # Instead of outputting `{}`, `[]` or even `{some: {other: value}}`, just print `<dict>`/`<list>`.
```
aqw, line 543: More of a comment rather than anything else, but I'm liking the idea that `<dict>` is the type, and would match `{}` or `{some: {other: value}}`. But `{}` would only match `{}` (i.e. an empty dict) and not ``{some: {other: value}}`. Same with lists, etc.

onyo/lib/commands.py
```diff
@@ -513,22 +513,39 @@ def onyo_get(inventory: Inventory,
...
+    # If information on content was wanted, the respective `keys` should be specified instead.
```
aqw, line 544: Nit: "was" -> "is" (only commenting because the confusion from the above "would" has continued to prime me. ;-)

aqw (approved):
Looks great! No blockers from my perspective. Some comments that would be nice to see fixed, but overall great. :-D

onyo/lib/commands.py
bpoldrack: In the line above we create an output dictionary that actually has literal dots in its keys. So code following that never needs to consider the wrapper.
A result is not an asset. It's a collection of key-value pairs from an asset. And it's assembled in order to print it. At this point it's perfectly fine to not carry the original structure around.

I'll think about rephrasing.

onyo/lib/commands.py
bpoldrack: I'm a bit confused. May be just because where you put that comment.

This `{}` only matching empty dicts is what's implemented. But this piece of code is about printing the value in `get`'s output not about matching.

onyo/lib/utils.py
```diff
@@ -19,10 +19,127 @@
...
+    flattened dictionary:
+
+    > d = {'key': 'value', 'nested': {'key': 'another value'}}
+    > wrapper = DotNotationWrapper(d)
+    > wrapper['nested.key']
+    'another value'
+    > list(wrapper.keys())
+    ['key', 'nested.key']
+
+    Iteration only considers the flattened view, and keys that contain dictionaries will not be yielded when using
```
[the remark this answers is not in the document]
bpoldrack: Agree. list-keys is when this may become relevant.

onyo/lib/commands.py
aqw: Complete agreement here. My comment was indeed just a comment, and rather only loosely on-topic for where I chose to do so.

onyo/lib/tests/test_commands_unset.py
```diff
@@ -299,3 +299,36 @@ def test_onyo_unset_asset_dir(inventory: Inventory) -> None:
...
+    """`unset` does work on empty dicts and lists, and None values"""
+    # Note: This is making sure onyo does not confuse an "empty value",
+    #       w/ the key not being present.
+
+    asset = dict(type="TYPE",
+                 make="MAKE",
+                 model=dict(name="MODEL"),
+                 serial="SERIAL2",
+                 directory=inventory.root)
+    test_pairs = dict(emptydict=dict(),
```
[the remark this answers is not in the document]
bpoldrack: Good catch. Tested and fixed.

bpoldrack:
@aqw : Pushed again. Noteworthy: Added a test for get's output for `None`, `''` values as suggested. Turns out that did indeed not output `<unset>`. It now does. I adjusted a test for the opposite (in case of empty string) accordingly. While I find it debatable that `<unset>` matches keys that are not present as well as `None` and empty strings, this is what's implemented all along and out should match that idea. If, given that at least for an empty string the opposite was actually tested, you think we should re-consider, make it an issue please. I'd consider this out-of-scope here.

onyo/lib/commands.py
```diff
@@ -483,6 +484,8 @@ def onyo_get(inventory: Inventory,
       `onyo.lib.consts.SORT_ASCENDING` and `onyo.lib.consts.SORT_DESCENDING`.
       If other values are specified an error is raised.
       Default: `{'path': SORT_ASCENDING}`.
+      Note, that this sorts the matching assets rather than the output.
```
aqw, line 487: Nit: The user only needs to know the feature.

Simply:

> One can sort by keys that are not in the output.

onyo/cli/get.py
```diff
@@ -96,12 +100,14 @@
...
         help=r"""
-            Sort output by **SORT-KEY** in ascending order.
+            Sort matches by **SORT-KEY** in ascending order.
             Can be given multiple times. Sorting by multiple **SORT-KEY** will be done in order
             (earlier given keys take precedence over subsequent keys).
             This can be intermixed with ``-s/--sort-descending``.
             Note, that if a **SORT-KEY** appears multiple times, the latest appearance will
-            overrule what ws specified before.
+            overrule what was specified before.
+            This sorts the matching assets rather than the output.
```
aqw, line 109: Nit: There's a tendency to over-explain here. But the user doesn't need to know /why/. They just need to know the feature.

Simply:

> One can sort by keys that are not in the output.

onyo/cli/get.py
```diff
@@ -111,12 +117,14 @@
...
         nargs='+',
         help=r"""
-            Sort output by **SORT-KEY** in descending order.
+            Sort matches by **SORT-KEY** in descending order.
             Can be given multiple times. Sorting by multiple **SORT-KEY** will be done in order
             (earlier given keys take precedence over subsequent keys).
             This can be intermixed with ``-s/--sort-ascending``.
             Note, that if a **SORT-KEY** appears multiple times, the latest appearance will
             overrule what ws specified before.
+            This sorts the matching assets rather than the output.
```
aqw, line 126: Nit: The user only needs to know the feature.

Simply:

> One can sort by keys that are not in the output.

onyo/cli/get.py
```diff
@@ -54,7 +56,9 @@
         help=r"""
             Criteria to match assets in the form ``KEY=VALUE``, where **VALUE**
             is a python regular expression. Pseudo-keys such as ``path`` can
-            also be used. Special values supported are:
+            also be used. Dictionary subkeys can be addressed using a period
+            (e.g. ``model.name``, ``model.year``, etc.)
+            Special values supported are:
```
aqw, line 61: Nit: Due to the new text for sorting, I propose adding here:

> One can match keys that are not in the output.

aqw (approved):
I'm sold, and feel this is ready to merge.

I did leave comments on the help text. I am comfortable with those fixes being added to this PR, or making them myself outside of it.

Thanks for making this feature happen!

aqw:
> I think this closes #114, despite not yet addressing lists. WDYT, @aqw ?

I agree that this closes #114 (and does actually address initialization of lists).

bpoldrack:
Thanks for thoroughly reviewing this beast! :)
Merging.

Issue #665, opened 2024-10-09
Asset dictionary type annotation is overly specific


bpoldrack:
Lots of functions are currently annotated like `Dict[str, Union[int, str, float, Path, bool]]`.
This is a pain and mostly wrong (since most functions would actually work perfectly fine with other types as values or even keys). It's main purpose is to make sure that reading and writing YAML works the way we expect it too. The place to make sure of that is unit tests for `dict_to_asset_yaml` and `asset_yaml_to_dict`, though.

Hence: Write those unit tests and reduce the type annotation for everything else to `Dict[str, Any]` at most.

aqw:
This has been address by my recent commits and supported by Ben's dot notation PR (#668).

Closing. See also #671

Issue #585, opened 2024-03-11
Sourcing shell completion outside of onyo repo results in `command not found: _onyo_dir`


nhjjr:
`(eval):1: command not found: _onyo_dir`

If you run `source <(onyo shell-completion)` before initializing an onyo repository (or perhaps when doing so outside of an onyo repository) it will return the above error/warning when attempting tab completion, even if at that point you're in a valid onyo repository.

nhjjr:
Today I made sure to `cd` into my onyo inventory and activate the virtual environment containing onyo's install before I sourced onyo's shell completion, yet still got the same eval error/warning upon attempting to use tab completion. Note that the root of my onyo inventory contains no assets, only a folder that has assets inside of it (in case that matters for reproducibility).

nhjjr:
I can no longer reproduce this, it seems to have been resolved. I'll close the issue.

aqw:
Great. Thanks!

PR #669, opened 2024-10-11
BF/TYPE: simplify some type checks; fix shell completion bug

aqw:
This PR has two unrelated fixes

- simplifies some unnecessary (and actively annoyingly) specific type annotations (#665)
- silences a one-time user error when first using shell completion (#585)

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -741,7 +741,7 @@ def onyo_new(inventory: Inventory,
              template: Path | str | None = None,
              clone: Path | None = None,
              tsv: Path | None = None,
-             keys: list[Dict[str, str | int | float]] | None = None,
+             keys: list[Dict] | None = None,
              edit: bool = False,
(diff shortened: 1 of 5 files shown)


bpoldrack:
I'm kinda  fine with that as is, but I'd like to point out, that I changed the type annotation for most things in PR #668 to `dict | UserDict` - not even restricting it to `Dict[str, Any]`. That's because most of the functions would work perfectly fine with any key type. It does not currently make sense to actually use other keys, but to me that's a matter of testing and it's only really relevant when loading/writing YAML. And even then: We'd simply get an exception, whenever ruamel doesn't know how to represent that key - no need to cover that by restrictive type annotation IMO.

So, in short: I'd be even more "loose" here (`dict | UserDict`) unless we know specifically that a particular function actually cares for the keys being strings.

aqw:
I'm comfortable with that. Can do. :-)

PR #683, opened 2024-10-17
NF: match `onyo get`'s exit codes to `grep`'s

### LINKED ISSUE #680: New Feature: `onyo get` return codes should match `grep`'s
Currently, `onyo get` returns `1` only if there is an error.

I propose that we match `grep`'s behavior, which allows for some useful functionality.

From `grep`'s manpage:
> EXIT STATUS
> Normally the exit status is 0 if a line is selected, 1 if no lines were selected, and 2 if an error occurred.  However, if the `-q` or `--quiet` or `--silent` is used and a line is selected, the exit status is 0 even if an error occurred.
> ...
> -q, --quiet, --silent
> Quiet; do not write anything to standard output.  Exit immediately with zero status if any match is found, even if an error was detected.

I propose that `onyo get`:
 
- return `0` if matches were found
- return `1` if no matches found (i.e. no results)
- return `2` for error

We already have `--quiet` as part of `onyo --quiet`

I cannot think of an error in `onyo get` that still leads to a successful query. So we can ignore that aspect of their functionality.

The real value of `--quiet` is to abort searching early once a match is found. But that optimization (and its architectural tradeoffs) is not within the scope of this issue. If desired, a separate issue should be opened to track that.

In the end, this functionality would allow for scripts such as the following:
```
for d in $(onyo get --type directory); do
    onyo --quiet get --type directory,asset --include "$d" || echo "empty dir: $d"
done
```

aqw:
Previously, `onyo get` returned `1` if there was an error and `0` if not.

This change exits `2` on error, exits `1` if no results are found, and `0` if there are results found.

I considered adding a `returncode` attribute to exceptions, matching what `CalledProcessError` does. But how the CLI exits has nothing to do with exceptions. So I did the check there and categorically modified the behavior of `onyo get`.

close #680

### DIFF
--- a/onyo/cli/get.py
+++ b/onyo/cli/get.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import sys
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -121,6 +122,13 @@ args_get = {
 }
 
 epilog_get = r"""
+.. rubric:: Exit Status
+
+The exit status is ``0`` if at least one result is found, ``1`` if there are no
+results, and ``2`` if an error occurred.
+
+These exit values match those of ``grep``.
+
 .. rubric:: Examples
 
 List all assets belonging to a user:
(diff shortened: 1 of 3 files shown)


onyo/cli/get.py
```diff
@@ -121,6 +122,13 @@
 }
 
 epilog_get = r"""
+.. rubric:: Exit Status
+
+The exit status is ``0`` if at least one result is found, ``1`` if there are no
```
TobiasKadelka, line 127: It surprises me that you would not use a list to explain 0, 1, 2 :)

TobiasKadelka (approved):
Nice and easy change, I like the mirroring of `grep` behavior, +1 :)

onyo/cli/get.py
aqw: Yeah, I copied `grep`'s manpage and worked from there. I think it's clear enough as it is, and uses less vertical space.

TobiasKadelka:
LGTM; I'll merge.

aqw:
Thanks!!

PR #687, opened 2024-10-22
Simplify Filter.match

bpoldrack:
There's no functional change in this patch. Just refactoring the code to be a bit simpler, slightly more efficient and (most importantly) a little easier to read and maintain. Previous version of the implementation was a bit convoluted as pointed out in https://github.com/psyinfra/onyo/pull/668#discussion_r1798456545

### DIFF
--- a/onyo/lib/filters.py
+++ b/onyo/lib/filters.py
@@ -50,30 +50,26 @@ class Filter:
 
     def match(self, asset: dict) -> bool:
         r"""match self on a dictionary"""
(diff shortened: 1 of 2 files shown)


aqw (approved):
Oh sweet wonder. Thank you so much for refactoring this. The new version is a delight to read. :-D

aqw (approved):
LGTM!

aqw:
I spoke too soon. Tests seem to fail.

aqw (approved):
LGTM!

PR #690, opened 2024-10-22
`tree:`: directories only

### LINKED ISSUE #677: New Feature: `onyo tree --directories`
Currently, `onyo tree` outputs directories and assets. It would be nice to be able to generate a tree of just directories (equivalent of `tree -d`).

I see two ways of implementing this:
1) `onyo tree --no-assets` (`-A`): this will not print any assets (and thus logically exclude asset directories)
2) `onyo tree --directories` (`-d`): this will print only directories (and thus logically include asset directories)

`onyo tree` is not meant to be parsed or filtered by anything. So there's no need to build a clever filtering mechanism here.

I currently favor option 2. And it has the added bonus of matching `tree`'s syntax.

---
Note: I /really/ liked the idea of having a `--type` flag like (#676) and the default as `--type directory asset` (current behavior). But the entire point of `tree` is the directory structure. `--type asset` would not work; `directory` is not optional.

bpoldrack:
This implements a directories only mode for `onyo tree`. Breaking change: Rename the existing `directories` argument for the python interface to `paths` in order to be less confusing. This makes sense in python as that argument is an interable of `Path` objects and it's irrelevant to the CLI, since it's a positional argument - so, no confusing option names.

In addition this fixes the command's help referring to a metavar for the positional argument that doesn't exist.
Lastly, a slight performance improvement for `onyo tree` that avoids `stat`ing the same path several times to tell whether it's a directory.

Closes #677

TODO: Tests


onyo/cli/tree.py
```diff
@@ -12,12 +12,19 @@
...
         help=r"""
             Directories to list.
         """
-    )
+    ),
+    'directories': dict(
+        args=("-d", "--dirs-only"),
+        action="store_true",
+        help=r"""
+            Whether to print directories only.
```
aqw, line 25: I prefer the more declarative:  "Print only directories."

onyo/lib/commands.py
```diff
@@ -1070,40 +1070,43 @@ def onyo_set(inventory: Inventory,
 
 @raise_on_inventory_state
 def onyo_tree(inventory: Inventory,
-              dirs: list[tuple[str, Path]]) -> None:
+              paths: list[tuple[str, Path]],
+              directories: bool = False) -> None:
     r"""Print the directory tree of paths.
```
aqw, line 1075: I'm fine with this change, but I just want to comment that I originally chose the var `dirs` because files are not a valid target for `onyo_tree()`. It must be a dir.

onyo/cli/tree.py
```diff
@@ -47,4 +54,5 @@ def tree(args: argparse.Namespace) -> None:
     dirs = dirs if dirs else [('.', Path.cwd())]
 
     onyo_tree(inventory,
-              dirs)
+              paths=dirs,
+              directories=args.directories)
```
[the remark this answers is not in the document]
bpoldrack: Yeah, probably better. Will do.

onyo/lib/tests/test_commands_tree.py
```diff
@@ -59,15 +59,36 @@ def test_onyo_tree_single(inventory: Inventory,
     assert inventory.repo.git.is_clean_worktree()
 
 
+@pytest.mark.ui({'yes': True})
+def test_onyo_tree_dirs_only(inventory: Inventory,
+                             capsys) -> None:
+    r"""Display directory structure w/o any assets or other files"""
```
aqw, line 65: Nit: this comment conflates assets with files (ignoring asset dirs, which are allowed).

onyo/lib/tests/test_commands_tree.py
```diff
@@ -59,15 +59,36 @@ def test_onyo_tree_single(inventory: Inventory,
...
+
+    onyo_tree(inventory,
+              paths=[(str(inventory.root), inventory.root)],
+              dirs_only=True)
+    tree_output = capsys.readouterr().out
+    for path in inventory.root.rglob('*'):
```
aqw, line 71: This test will clearly include and check for asset directories (good), but my question is: does the test repo have an asset dir in it?

I'd just like to make sure that case is tested (and there's at least a comment explicitly stating so).

aqw (changes requested):
Looks great. I'd like to see a test/comment explicitly checking that asset directories are in the output.

Other than that: perfect. :-D

aqw (approved):

aqw:
LGTM! Merging. :-D

PR #695, opened 2024-10-24
RF: `os.linesep` -> `\n`

aqw:
`os.linesep` is of no use to us, as git translates line endings to match the user's OS.

The codebase used both `linesep` and `\n`. It makes sense to choose the one that is 5-10 characters shorter, faster, and doesn't require an import.

This removes 9 out of 16 imports of `os`.


bpoldrack:
Thx!

PR #701, opened 2024-10-24
Rework `--message` parameter

### LINKED ISSUE #685: `--message` should append to rather than overwrite the commit message
In practice, we have found that it is far, far more common to want to *add* a note to the commit message than it is to overwrite the informative, auto-generated commit message. For example:

```
onyo mv shelf/laptop_apple_mba2023.abc1234 Bingo\ Bob/ --message "close #1127"
```

I propose the following:
- `-m, --message MESSAGE`: Append the given MESSAGE to the auto-generated commit message. If multiple `--message` options are given, their values are concatenated as separate paragraphs.
- `-M, --message-overwrite MESSAGE`: Use the given MESSAGE as the commit message (discarding the default). If multiple `--message` options are given, their values are concatenated as separate paragraphs.

In a conversation with Ben, I had a better idea than `--message-overwrite`, but I have since forgotten in. Grr.

What is unclear to me is whether `-m` and `-M` can be used together. I feel like they should not. But, they both fundamentally have the same append behavior.

Alternate idea that avoids the behavioral ambiguity above:

- `--discard-auto-generated-commit-message`: Discard the auto-generated commit message. Requires `-m, --message`.
- `-m, --message MESSAGE`: Append the given MESSAGE to the commit message. If multiple `--message` options are given, their values are concatenated as separate paragraphs. To discard the auto-generated commit message, see `--discard-auto-generated-commit-message`.
- `-M, --message-overwrite`: short for `--message` and `--discard-auto-generated-commit-message`

I don't like the verbosity of `--discard-auto-generated-commit-message`, but at least it's clear. ;-) Better ideas are most welcome.

Open questions:
- will overwriting (discarding, whatever) retain or remove the inventory operations log? In other words: are we overwriting the subject line or indeed the entire message? Do we want each to be controllable? Like `--no-autogen-msg=subject,operations`.

bpoldrack:
Closes #685

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -391,7 +391,8 @@ def _edit_asset(inventory: Inventory,
 @raise_on_inventory_state
 def onyo_edit(inventory: Inventory,
               paths: list[Path],
-              message: str | None) -> None:
+              message: str | None,
+              auto_message: bool | None = None) -> None:
     r"""Edit the content of assets.
 
     Parameters
@@ -401,16 +402,20 @@ def onyo_edit(inventory: Inventory,
     paths
         Paths of assets to edit.
     message
-        A custom commit message.
+        Commit message to append to the auto-generated message.
+    auto_message
+        Generate a commit-message subject line.
+        If ``None``, take the value from the 'onyo.commit.auto-message' configuration.
 
     Raises
     ------
     ValueError
         If a provided asset is not an asset, or if ``paths`` is empty.
     """
-
     from functools import partial
 
+    if auto_message is None:
+        auto_message = inventory.repo.auto_message
     if not paths:
         raise ValueError("At least one asset must be specified.")
 
@@ -426,16 +431,15 @@ def onyo_edit(inventory: Inventory,
 
     if inventory.operations_pending():
         if ui.request_user_response("Save changes? No discards all changes. (y/n) "):
-            if not message:
+            if auto_message:
                 operation_paths = sorted(deduplicate([
                     op.operands[0].get("path").relative_to(inventory.root)
                     for op in inventory.operations
                     if op.operator == OPERATIONS_MAPPING['modify_assets']]))
-                message = inventory.repo.generate_commit_message(
+                message = inventory.repo.generate_auto_message(
                     format_string="edit [{len}]: {operation_paths}",
                     len=len(operation_paths),
-                    operation_paths=operation_paths)
-
+                    operation_paths=operation_paths) + (message or "")
             inventory.commit(message=message)
             return
 
@@ -578,7 +582,8 @@ def onyo_get(inventory: Inventory,
 @raise_on_inventory_state
 def onyo_mkdir(inventory: Inventory,
                dirs: list[Path],
-               message: str | None) -> None:
+               message: str | None = None,
+               auto_message: bool | None = None) -> None:
     r"""Create new directories in the inventory.
 
     Intermediate directories will be created as needed (i.e. parent and
@@ -603,13 +608,18 @@ def onyo_mkdir(inventory: Inventory,
     dirs
         Paths to directories which to create.
     message
-        An optional string to overwrite Onyo's default commit message.
+        Commit message to append to the auto-generated message.
+    auto_message
+        Generate a commit-message subject line.
+        If ``None``, take the value from the 'onyo.commit.auto-message' configuration.
 
     Raises
     ------
     ValueError
         If `dirs` is empty.
     """
+    if auto_message is None:
+        auto_message = inventory.repo.auto_message
     if not dirs:
         raise ValueError("At least one directory path must be specified.")
     for d in deduplicate(dirs):  # pyre-ignore[16]  deduplicate would return None only of `dirs` was None.
@@ -620,15 +630,15 @@ def onyo_mkdir(inventory: Inventory,
         ui.print(inventory.operations_summary())
 
         if ui.request_user_response("Save changes? No discards all changes. (y/n) "):
-            if not message:
+            if auto_message:
                 operation_paths = sorted(deduplicate([
                     op.operands[0].relative_to(inventory.root)
                     for op in inventory.operations
                     if op.operator == OPERATIONS_MAPPING['new_directories']]))
-                message = inventory.repo.generate_commit_message(
-                    format_string="mkdir [{len}]: {operation_paths}",
+                message = inventory.repo.generate_auto_message(
+                    format_string="mkdir [{len}]: {operation_paths}\n",
                     len=len(operation_paths),
-                    operation_paths=sorted(operation_paths))
+                    operation_paths=sorted(operation_paths)) + (message or "")
             inventory.commit(message=message)
             return
     ui.print('No directories created.')
@@ -672,7 +682,8 @@ def _maybe_rename(inventory: Inventory,
 def onyo_mv(inventory: Inventory,
             source: list[Path] | Path,
             destination: Path,
-            message: str | None = None) -> None:
+            message: str | None = None,
+            auto_message: bool | None = None) -> None:
     r"""Move assets or directories, or rename a directory.
 
     If `destination` is an asset file, turns it into an asset dir first.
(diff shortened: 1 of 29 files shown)


onyo/shared_arguments.py
```diff
@@ -4,8 +4,20 @@
     action='append',
     type=str,
     help=r"""
-        Use the given **MESSAGE** as the commit message (rather than the default).
+        Append the given **MESSAGE** to the commit message.
```
TobiasKadelka, line 7: I know this is something that was probably written by myself, but while reading it I thought `Append **MESSAGE** to commit message.` is better, because it is shorter and sounds more like the help below.

TobiasKadelka:
While this is a draft, I still had a look because I am excited for the functionality to be added.
Very cool, thanks for working on it :)

onyo/shared_arguments.py
aqw: I agree on the more terse language. I've been noticing other small programmer-isms in the help text, and it is good to remove them.

onyo/lib/commands.py
```diff
@@ -402,6 +403,8 @@ def onyo_edit(inventory: Inventory,
         Paths of assets to edit.
     message
         A custom commit message.
```
aqw, line 405: > A custom commit message.

This should also be updated, as it's not "custom" anymore. Perhaps:

> Commit message to append to the auto-generated message.

onyo/lib/commands.py
```diff
@@ -402,6 +403,8 @@ def onyo_edit(inventory: Inventory,
...
+    no_auto_message
+        If True, do not auto-generate a commit message subject line.
```
aqw, line 407: I know that this is matching the CLI flag `--no-auto-message` (which I do believe is the right name). But I am not a fan of passing `True` to enable the disabling of something.

I would prefer seeing the default as `auto_message=True` and we pass `False` to disable the `auto_message`.

Argparse supports `store_false` as the `action` and then `dest` to name the variable (so that it differs from the flag name).

onyo/lib/commands.py
```diff
@@ -426,16 +429,15 @@ def onyo_edit(inventory: Inventory,
 
     if inventory.operations_pending():
         if ui.request_user_response("Save changes? No discards all changes. (y/n) "):
-            if not message:
+            if not no_auto_message:
```
aqw, line 432: This scenario further supports my suggestion of renaming to `auto_message`.

onyo/lib/commands.py
```diff
@@ -604,6 +607,8 @@ def onyo_mkdir(inventory: Inventory,
         Paths to directories which to create.
     message
         An optional string to overwrite Onyo's default commit message.
```
aqw, line 609: This should be updated (and others). I only point it out again because it doesn't mention "custom". Rather, all `message` parameters should be checked/unified.

onyo/shared_arguments.py
```diff
@@ -4,8 +4,20 @@
...
+
+shared_arg_no_auto_message = dict(
+    args=('--no-auto-message',),
+    action='store_true',
+    help=r"""
+        Do not use auto-generated commit message subject.
+        This does not disable the inventory operations record
+        at the end of a commit message.
+        If there's also no **MESSAGE** given, the subject line
+        will be ``-``.
```
aqw, line 21: I prefer something explicit, rather than a seemingly arbitrary `-`. Perhaps `[Empty Subject]`?

onyo/lib/commands.py
```diff
@@ -401,16 +402,21 @@ def onyo_edit(inventory: Inventory,
     paths
         Paths of assets to edit.
     message
-        A custom commit message.
+        Commit message to append to the auto-generated message.
+    auto_message
+        If `True`, do auto-generate a commit message subject line.
+        The default can be configured via 'onyo.commit.auto-message'.
+        Onyo configures this to be `True` when initializing a repository.
```
aqw, line 409: I don't think this should be commented here. It can change at any time, and doesn't affect how this function works.

onyo/lib/commands.py
```diff
@@ -401,16 +402,21 @@ def onyo_edit(inventory: Inventory,
     paths
         Paths of assets to edit.
     message
-        A custom commit message.
+        Commit message to append to the auto-generated message.
+    auto_message
+        If `True`, do auto-generate a commit message subject line.
+        The default can be configured via 'onyo.commit.auto-message'.
```
aqw, line 408: Nit: "can be" -> is

IMO, this should more explicitly state that `None` means use the default that's configured at 'onyo.commit.auto-message'.

onyo/lib/onyo.py
```diff
@@ -141,6 +141,20 @@ def get_config(self,
...
+    @property
+    def auto_message(self) -> bool:
+        raw = self.get_config("onyo.commit.auto-message")
+        if raw:
+            from_cfg = raw.strip().lower()
+            if from_cfg in ["true", "1"]:
+                return True
+            if from_cfg in ["false", "0"]:
+                return False
+            ui.log(f"Invalid config value \"{raw}\" for 'onyo.commit.auto-message'. Using default.",
```
aqw, line 153: Nit: It would be nice to state what that default is.

onyo/shared_arguments.py
```diff
@@ -4,8 +4,22 @@
...
+    args=('--no-auto-message',),
+    action='store_true',
+    help=r"""
+        Do not use auto-generated commit message subject.
+        This does not disable the inventory operations record
+        at the end of a commit message.
+        If there's also no **MESSAGE** given, the subject line
+        will be ``[Empty subject]``.
+        The default can be configured via 'onyo.commit.auto-message'
+        and is set to ``true`` by ``onyo init``.
```
aqw, line 23: I propose:

> Do not auto-generate the commit message subject.
> If no **MESSAGE** is given, the subject line will be
> ``[Empty subject]``.
> The default can be configured via 'onyo.commit.auto-message'.

aqw (changes requested):
Despite how big the patch is, this was really straightforward to review. And the improvements to the tests are wonderful.

Most of my comments are about the docstrings/help-text.

One thing that is missing is adding the `--no-auto-message` flag to the tab completion.

onyo/lib/commands.py
```diff
@@ -401,16 +402,21 @@ def onyo_edit(inventory: Inventory,
     paths
         Paths of assets to edit.
     message
-        A custom commit message.
+        Commit message to append to the auto-generated message.
+    auto_message
+        If `True`, do auto-generate a commit message subject line.
```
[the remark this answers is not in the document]
bpoldrack: Agree on this wording being more concise. Otherwise: It is documenting a python function for programmers who want to use it. So, I find the argument a bit odd.

However, will do.

onyo/shared_arguments.py
aqw: I don't know if you intentionally resolved this without changes, or if it was an accident. Bumping this so you can say explicitly. :-)

aqw (approved):
Beautiful. LGTM.

I bumped one comment that you didn't change nor respond to.

Overall +1 on merging.

onyo/shared_arguments.py
bpoldrack: It is changed to what you said? I'm confused.

bpoldrack: Or do you want to remove the line about the operations record?

aqw: Ahh, I was intentionally removing the information about `onyo init` (as suggested elsewhere) and the operations record.

bpoldrack: Ah, okay. I changed other parts of the wording, but kept this part, since I thought this useful when it's actually showing up in the command line help.

bpoldrack: (As opposed to python functions where it's indeed more a burden on maintenance than anything else)

aqw: I'm fine with keeping the comment about the inventory operations. But the `init` comment should definitely be removed.

Otherwise, we should add that line to every other configuration option we have. It doesn't make sense. The config just "is" part of Onyo.

PR #702, opened 2024-10-25
NF/DOC: `onyo cat` improved documentation and enhance error codes

aqw:
This continues my journey to bring more features to the most irrelevant parts of `onyo`. ;-)

Really, this was an accident. I was looking at #493 and chose to work on `cat`'s docstrings. There was ambiguity about error cases and invalid content, and I wanted to clarify it.

Previously, `onyo cat` would exit `0` on success and `1` on error (no print content) or invalid asset content (but still print content).

I really like `onyo get`'s new ability to differentiate between producing valid output that is problematic (exit `1`) vs an error (exit `2`). And so I chose to bring that to `onyo cat`. Invalid asset content will continue to print and exit `1`. Errors will now exit `2`.

I also introduce a new exception (`InvalidAssetError`), as I felt `OnyoInvalidRepoError` was misleading and was being abused.

### DIFF
--- a/onyo/cli/cat.py
+++ b/onyo/cli/cat.py
@@ -4,6 +4,10 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from onyo.lib.commands import onyo_cat
+from onyo.lib.exceptions import (
+    InvalidAssetError,
+    OnyoCLIExitCode,
+)
 from onyo.lib.inventory import Inventory
 from onyo.lib.onyo import OnyoRepo
 
@@ -21,6 +25,11 @@ args_cat = {
 }
 
 epilog_cat = r"""
+.. rubric:: Exit Status
+
+The exit status is ``0`` if the asset contents are valid, ``1`` if the
+contents are invalid, and ``2`` if an error occurred.
+
 .. rubric:: Examples
 
 Display the contents of an asset (file or directory):
@@ -47,11 +56,17 @@ def cat(args: argparse.Namespace) -> None:
     r"""
     Print the contents of **ASSET**\ s to the terminal.
 
-    If any of the paths are invalid, then no contents are printed and an error
-    is returned.
+    If any of the paths are not assets, no content is printed and an error is
+    returned.
+
+    Assets with invalid content are printed with an error message. See **Exit
+    Status** for more information about return codes.
     """
     paths = [Path(p).resolve() for p in args.asset]
 
     inventory = Inventory(repo=OnyoRepo(Path.cwd(), find_root=True))
-    onyo_cat(inventory,
-             paths)
+    try:
+        onyo_cat(inventory,
+                 paths)
+    except InvalidAssetError as e:
+        raise OnyoCLIExitCode("'onyo cat' exits 1 when invalid asset content is found.", 1) from e
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -25,13 +25,14 @@ from onyo.lib.consts import (
     SORT_DESCENDING,
 )
 from onyo.lib.exceptions import (
+    InvalidAssetError,
+    InventoryDirNotEmpty,
     NotADirError,
     NotAnAssetError,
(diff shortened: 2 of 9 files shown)


onyo/cli/cat.py
```diff
@@ -47,11 +54,17 @@ def cat(args: argparse.Namespace) -> None:
...
     paths = [Path(p).resolve() for p in args.asset]
 
     inventory = Inventory(repo=OnyoRepo(Path.cwd(), find_root=True))
-    onyo_cat(inventory,
-             paths)
+    try:
+        onyo_cat(inventory,
+                 paths)
+    except InvalidAssetError:
+        sys.exit(1)
```
bpoldrack, line 70: This seems inconsistent with the rest of onyo. `InvalidAssetError` should bubble up and be caught in main. Or is there a reason to not go through that place that would usually decide how to render a message etc.?

bpoldrack (commented):
I think this should change the place where an exception is actually caught and handled.

Otherwise, LGTM. Agree on your introduction of a new error class, too.

onyo/cli/cat.py
aqw: This is actually intentional. It is a special case, just like `onyo get`. All other errors will exit `2` in `main.py`. This unique case for `cat` is caught, and handled by exiting `1`.

TobiasKadelka: Could you maybe add a note as a comment that it is intentional? :)
Things like that will definitely be forgotten by me when I look at the code in some months.

aqw: If Ben doesn't have a suggestion for a better way of approaching this, I am happy to add a comment stating its intentionality. :-)

onyo/lib/commands.py
```diff
@@ -200,7 +201,7 @@ def onyo_cat(inventory: Inventory,
     # TODO: "Full" asset validation. Address when fsck is reworked
     assets_valid = validate_yaml(deduplicate(files))
     if not assets_valid:
-        raise OnyoInvalidRepoError("Invalid assets")
+        raise InvalidAssetError("Invalid assets")
```
bpoldrack, line 204:
```suggestion
        e = InvalidAssetError("Invalid assets")
        e.returncode = 2
        raise e
```

bpoldrack: This could be done either here or in the CLI wrapper (catch the error, append the code `raise ... from`)
No need for special casing where exit anywhere other than main.

onyo/cli/cat.py
bpoldrack: See https://github.com/psyinfra/onyo/pull/702/files#r1818553447

bpoldrack: I would, of course, suggest the same thing for `get`. Simply missed that as I didn't get to review the get PR before it was merged.

aqw: I have fixed this, and it was much more of a rabbit hole than I expected. ;-)

I have updated the PR with three additional commits:
- `ui.error` no longer requires using the message of the exception
- add `OnyoCLIExitCode` exception
- switch `onyo get` over to `OnyoCLIExitCode`

Feedback much appreciated.

onyo/lib/commands.py
aqw: `pyre` chokes on this, and wants the Exception to have these attributes defined.

onyo/cli/cat.py
aqw: I have pushed again, this time splitting the traceback-to-string functionality into its own function. There already was a `ui.log_debug()`, that just wasn't being used by `ui.error()`.

I'm not thrilled with the name `format_traceback()`, but felt my other ideas weren't any better (`get_traceback_as_string()`, `traceback_to_string()`, etc). Ideas welcome here.

aqw: The `pyre `failures are new and unrelated to this PR.

I will investigate them but currently cannot reproduce locally.

PR #704, opened 2024-10-26
RF/DOC: normalize `onyo history`

### LINKED ISSUE #494: Tests + docstring for onyo_history
Properly document `onyo/lib/commands.py:onyo_history` and provide tests in `onyo/lib/tests/test_command_history.py`.
Based on that reworked docstring, provide a respective command line version in `onyo/commands/history.py`.

aqw:
This normalizes the `onyo history` command to gain (and call) an `onyo_history()` function and follow other common idioms.

Error codes now bubble up the same as all other shell calls.

`get_history_cmd()` has been moved from command_utils.py and adjusted to be a private helper function for `onyo_history()`.

The `interactive` parameter has been enhanced to allow forcing interactive on (`True`) or off (`False`), and auto-detection (`None`). Previously `True` meant auto-detection and one could not force on interactive mode.

The `onyo history` command is still unable to force interactive mode to be on. Such functionality is not considered to be valuable, but can be added trivially if it is later deemed desirable.

The `onyo history` command no longer checks if the target path exists. That's considered the responsibility of the configured history command.

I made a serious effort to allow multiple paths to be passed to `onyo history`, and have the configured program throw an error if it can only accept one path.  Both `tig` and `git log` allow multiple paths, but neither allow it with the `--follow` flag enabled, which is simply crucial for Onyo. Files move, a lot. This is a limitation of how git handles moved files (or rather doesn't). Any history tool will suffer from this, so I accepted it as a fixed limitation of `onyo history` itself.

close #494


TobiasKadelka:
This looks great, it is very cool to see this command normalized! :) 👍

onyo/lib/commands.py
```diff
@@ -575,6 +575,89 @@ def onyo_get(inventory: Inventory,
...
 
 
+@raise_on_inventory_state
+def onyo_history(inventory: Inventory,
+                 path: Path,
+                 interactive: bool | None = None) -> None:
+    r"""Display the history of a path.
+
+    Only one ``path`` is accepted as that's ``git log --follow``'s limitation.
+    This is due to git's design, and is independent of the tool used to query it.
```
bpoldrack, line 585: Hm. TBH: I don't buy it. Of course I can build a tool that takes several paths to follow. Not different from showing all current branch tips, etc. 
I don't know it exists, but it absolutely could.

aqw: I had a similar impulse, but then started reading and people seemed very convinced of the complexity of the issue.

I do feel comfortable keeping this limitation for now, as neither `git` nor `tig` support it.

That said, if you feel strongly about it, I can adjust this. I would take the approach of adding a script (perhaps to `.onyo/bin`) to run `tig`, as its error message for passing multiple arguments to `--follow` is not helpful. That may or may not be its own can of worms.

bpoldrack: I'm kinda fine with the limitation, even though I'd think it nicer to let the configured command error.
I'm not fine with that claim.
Obviously, I can say `git log --follow` on several paths, simply by having one call per path. To claim, that there can't be a tool that collects that information and show it in a useful way is a really bold claim, that I see no basis for.

bpoldrack: The only thing not trivial about it I ca see, is to figure out whether one can usefully draw a shared graph in 2D and/or how to break it up into a minimal number of such graphs. As opposed to treating the given paths separately (which is - again - trivial and obvious)

aqw: Pushed.

aqw:
I have updated the docstring, and added an unrelated commit that resolves a cognitive complexity complaint by code climate.

I basically convert a stack of if-else's with a case statement. I don't their advice here can be generalized, but in this relatively simple case I agree. It's easier to read as they're all identically stacked, and it's mentally enforced that all are the same type comparing just one char.

bpoldrack:
Thx, @aqw ! LGTM

PR #707, opened 2024-10-30
Getting rid of YAML implicit typing (for now)

### LINKED ISSUE #261: TST: "round trip mode"
We use`ruamel` because it enables "round trip mode", which means even when reading, updating, and writing assets again, it preserves existing comments in YAML files, as well as the order of contents (all information gets loaded into a "orderedDict").

We have some test cases that test this behaviour, and obviously I ran some manual tests and played around with it.
But It would be nice to have some specific test cases to verify that these functionalities are given and used by our implementation (especially the combinations with comments and order after code changes should be automatically verified).

This could mean:
- longer files with normal yaml and with and without additional comments, in alphabetical and non-alphabetical order
- what about reading "dummy files" that have no key/values, but just comments?
- what happens when the user `onyo unset`s the last key/value and just comments would stay?

While we do not want to "re-test" all functionalities from external libraries, we should still make sure that our tests cover our own use cases and the code we use for reading/writing of assets does pass the information between functions like wanted and assumed.

### LINKED ISSUE #643: `edit` and `new --edit` are inconsistent WRT leading zeroes
Creating an asset with an inventory number w/ leading zeroes works just fine. However, editing such a number by adding leading zeroes only works by quoting. 

Make this consistent and figure out how to deal with yaml implicit typing

### LINKED ISSUE #696: Bug: weird casting of `3_5` to `35` resulting in `KeyError`
This is easiest to explore in our inventory repo, but I will try to provide some context:

Some assets contain the following structure:
```
connectors:
  mDP: 1
  DP: 2 # 1 x out w/ MST
  HDMI: 1
  USB-A: 4
  USB-B: 1
  3_5: 1
  DVI: 1
```

However, when I try to perform any form of write action on that asset, it results in an error:
```
❱ onyo edit warehouse/display_dell_u2413f.abc123
ERROR: "'connectors.35'"
```
```
❱ onyo set --keys test=toast --asset warehouse/display_dell_u2413f.abc123
ERROR: "'connectors.35'"
```
Reading is a different story
```
❱ onyo cat warehouse/display_dell_u2413f.abc123
---
...
connectors:
  mDP: 1
  DP: 2 # 1 x out w/ MST
  HDMI: 1
  USB-A: 4
  USB-B: 1
  3_5: 1
  DVI: 1
```

The `connectors` key works, but something is weird/wrong with `3_5`:
```
❱ onyo get --keys connectors connectors.DVI connectors.3_5 connectors.35 --path warehouse/display_dell_u2413f.abc123
 ────────────────────────────────────────────────────────────── 
  connectors   connectors.DVI   connectors.3_5   connectors.35  
 ────────────────────────────────────────────────────────────── 
  <dict>       1                <unset>          <unset>        
 ────────────────────────────────────────────────────────────── 
```
When I edit the asset manually and change `3_5` to `3-5`, everything is fine again:
```
❱ onyo get --keys connectors connectors.3_5 connectors.35 connectors.3-5 --path warehouse/display_dell_u2413f.abc123
  ────────────────────────────────────────────────────────────── 
  connectors   connectors.3_5   connectors.35   connectors.3-5  
 ────────────────────────────────────────────────────────────── 
  <dict>       <unset>          <unset>         1               
 ────────────────────────────────────────────────────────────── 
```

When I enable debugging, I get the following traceback:
```
❱ onyo --debug set --keys test=toast --asset warehouse/display_dell_u2413f.abc123
...
ERROR: "'connectors.35'"
  File "/home/aqw/git/inm7/onyo/onyo/lib/utils.py", line 91, in __getitem__
    return effective_dict[parts[-1]]
           ~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/home/aqw/.venvs/onyo/lib/python3.12/site-packages/ruamel/yaml/comments.py", line 853, in __getitem__
    return ordereddict.__getitem__(self, key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: '35'

The above exception was the direct cause of the following exception:

  File "/home/aqw/git/inm7/onyo/onyo/main.py", line 543, in main
    args.run(args)
  File "/home/aqw/git/inm7/onyo/onyo/cli/set.py", line 109, in set
    onyo_set(inventory=inventory,
  File "/home/aqw/git/inm7/onyo/onyo/lib/commands.py", line 82, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/aqw/git/inm7/onyo/onyo/lib/commands.py", line 1063, in onyo_set
    inventory.modify_asset(asset, new_content)
  File "/home/aqw/git/inm7/onyo/onyo/lib/inventory.py", line 437, in modify_asset
    self.raise_required_key_empty_value(new_asset)
  File "/home/aqw/git/inm7/onyo/onyo/lib/inventory.py", line 709, in raise_required_key_empty_value
    if any(v is None or not str(v).strip()
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aqw/git/inm7/onyo/onyo/lib/inventory.py", line 710, in <genexpr>
    for k, v in asset.items()
                ^^^^^^^^^^^^^
  File "<frozen _collections_abc>", line 894, in __iter__
  File "/home/aqw/git/inm7/onyo/onyo/lib/utils.py", line 93, in __getitem__
    raise KeyError(f"'{key}'") from e
KeyError: "'connectors.35'"
```

This behavior is reproducible with `onyo new` and templates.

bpoldrack:
See commit message.

This is essentially ready, but: It wouldn't be ready for a release or actual usage (which speaks to a lack of testing).
The reason is the `is_asset_directory` key, which is expected to be a boolean rather than a string.
But with this PR, users can't easily pass a boolean anymore.
However, the clean solution for that, would rely on a proper implementation of pseudo keys in general, so I'd prefer to solve the problem in that PR rather than here.

Closes #696
Closes #643
Closes #261

### DIFF
--- a/onyo/lib/tests/test_utils.py
+++ b/onyo/lib/tests/test_utils.py
@@ -1,28 +1,45 @@
 import pytest
-from ruamel.yaml import YAML  # pyre-ignore[21]
 
 from onyo.lib.consts import PSEUDO_KEYS, RESERVED_KEYS
-from onyo.lib.utils import dict_to_asset_yaml
+from onyo.lib.utils import (
+    dict_to_asset_yaml,
+    get_asset_content,
+)
+
+asset_file_content = """---
+# top-level comment
+type: a  # key comment
+make: b
+model:  # comment at intermediate node
+  name: c # comment in nested dict-key
+  # comment in nested dict
+  more: d
+  integer: 1
+  explicit: !!int '2'
+  float: 1.2
+  list:
+  - 1.2
+  - b
+  - 003_5
+# keys and values need to preserve leading zeroes and underscores:
+serial: 00012_3456
+003_5: true
+explicit_null: null
+tilde_null: ~ # what now
+implicit_null:
+emptystring: ''
+description: |
+  This is a long text
+  containing multiple lines.
+"""
 
 
 def test_dict_to_asset_yaml() -> None:
     r"""Test Dict, CommentedMap, and empty dict."""
     # normal python dict
-    d = {'type': 'TYPE', 'make': 'MAKE', 'model': 'MODEL', 'serial': 8675309}
-    d_expected_output = "---\ntype: TYPE\nmake: MAKE\nmodel: MODEL\nserial: 8675309\n"
+    d = {'type': 'TYPE', 'make': 'MAKE', 'model': 'MODEL', 'serial': '008675309', 'explicit': 123}
+    d_expected_output = "---\ntype: TYPE\nmake: MAKE\nmodel: MODEL\nserial: 008675309\nexplicit: !!int '123'\n"
     assert d_expected_output == dict_to_asset_yaml(d)
-
-    # YAML with comments
-    yaml_string = "---\n" + \
-                  "model: MODEL\n" + \
-                  "# You can be my Yoko Onyo\n" + \
-                  "make: MAKE\n" + \
-                  "type: TYPE\n" + \
-                  "# service tag for Dell\n" + \
-                  "serial: SERIAL\n"
-    yaml = YAML(typ='rt', pure=True)
-    yaml_dict_w_comments = yaml.load(yaml_string)
-    assert yaml_string == dict_to_asset_yaml(yaml_dict_w_comments)
 
     # empty top-level dict should be stripped
     empty = {}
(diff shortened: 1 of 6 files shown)


onyo/lib/utils.py
```diff
@@ -201,7 +201,7 @@ def dict_to_asset_yaml(d: Dict | UserDict) -> str:
     return s.getvalue()
 
 
-def get_asset_content(asset_file: Path) -> dict[str, bool | float | int | str | Path]:
+def get_asset_content(asset_file: Path) -> dict:
```
aqw, line 204: In this case, I think we do want to specify the type of the key, since that's literally what we're doing. What about `dict[str, Any]:`

bpoldrack: Won't fix. Clarified in chat.

aqw (approved):
Overall LGTM. I have some questions that I'd like to explore around typing, but this PR is not the place for that. That's instead about validation/schemas/parsers.

So with that in mind: LGTM!

aqw:
Ahh, I forgot to also mention, this PR may also address the following two issues:
- #643 
- #261

onyo/lib/tests/test_utils.py
```diff
@@ -1,43 +1,99 @@
...
+  explicit: !!int '2'
+  float: 1.2
+  list:
+  - 1.2
+  - b
+  - 003_5
+# keys and values need to preserve leading zeroes and underscores:
+serial: 00012_3456
+003_5: true
+nothing: null
```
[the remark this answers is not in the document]
bpoldrack: TBH: I choose the "nothing-ness", so that the test can look for a consistent substring in the relevant keys.
Should be a bit nicer, though - agree.

PR #711, opened 2024-10-30
BF: retire `onyo get --path` alias, and other cleanups

aqw:
This PR
- fixes some outdated docs (in general, I'm leaving RTD alone for now, as it's so out of date)
- drops `onyo get --path`; it was an alias for `onyo get --include`
  - it made the CLI feel inconsistent, and was confusing
- adds `--include` and `--exclude` to `onyo get`'s tab completion (somehow overlooked)


bpoldrack (approved):
Thanks!

PR #712, opened 2024-10-31
TST: flake8 -> ruff

aqw:
[ruff](https://docs.astral.sh/ruff/) is a drop-in replacement for just about all Python linters.

It's blazingly fast (not an overstatement), and supports [far more rules](https://docs.astral.sh/ruff/rules/) than flake8.

This is a naive migration from flake8. More checks can be enabled as we see fit.

There is a github action for `ruff`. I didn't se the advantage of using it.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,8 +1,0 @@
-[flake8]
-ignore =
-    # E501: Line too long (82 > 79 characters)
-    E501,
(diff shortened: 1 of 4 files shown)


TobiasKadelka:
LGTM, very little room for controversy, so I'll merge :)

PR #725, opened 2024-11-08
DOC/BF: cleanup docstring for `rm` and `tree`; fix `onyo_tree()`'s awkwardness

aqw:


TobiasKadelka:
This looks all good to me, nice that these tasks are finally done!
I am not sure if Ben wants to have a look at it, too, so I do not merge just yet.

onyo/lib/tests/test_commands_tree.py
```diff
@@ -3,63 +3,79 @@
...
     asset_path = inventory.root / "somewhere" / "nested" / "TYPE_MAKER_MODEL.SERIAL"
     dir_path = inventory.root / 'empty'
 
     # no tree for files
     pytest.raises(ValueError,
                   onyo_tree,
                   inventory,
-                  paths=[(str(asset_path), asset_path)])
+                  path=asset_path,
+                  description='file')
```
bpoldrack, line 18: It's a bit of a nitpick, but putting an invalid `description` in a test that's supposed to test a failure on certain `path`s doesn't make a lot of sense. This does effectively not test what it says it tests, b/c the ValueError could be caused by the description rather than the path.

I realize that's not introduced by this commit, but still: Once we touch it  ...

bpoldrack: Same goes, of course, for the ones below.

bpoldrack (commented):
Not strictly a blocker, but may be the testing could be addressed.

Otherwise LGTM

onyo/lib/tests/test_commands_tree.py
aqw: Ahh, yeah. I see your point. Keep the failure as tight as possible.

I added the description to make it more explicit/self-documenting. But I will leave that to the comments.

I'll update the PR.

bpoldrack:
Thanks!

PR #726, opened 2024-11-08
BF: package and access skel, zsh completion, and TSVs as resources

### LINKED ISSUE #362: Turn non-source-code files into package resources
We have a bunch of files that aren't part of the actual source code.
These are accessed by code via hardcoded, relative paths within the repo's tree.
For example: https://github.com/psyinfra/onyo/blob/399bcb05a5873dcd0e742021a21bf4fe3c7653a9/onyo/lib/onyo.py#L236

That's not good, since the implementation of that function relies on its (as well as `skel`'s) location in the repo. Same goes for tests utilizing a bunch of tsv files. At least in the latter case, those appear to not even be installed.

This should be proper package data/resources and accessed as such.

aqw:
I found this guide to be quite useful: https://setuptools.pypa.io/en/latest/userguide/datafiles.html

fix #362


bpoldrack (approved):
OMG. Thank you!

Issue #689, opened 2024-10-22
Design: Aliases for Keys


aqw:
Allow an alias for shorthand to longer keynames (such as `path` for `onyo.path.relative`). They are merely aliases, and thus technically do not reserve key space. It's rather a squatting of namespace because it translates them to another space.

Though the original idea for this came from pseudo keys (see #688), Aliases work with any key names.

When `get`ing or `set`ing, there is no attempt to check if the alias name exists as real data in the target assets (or anywhere). Aliases are expanded immediately, and the internals of Onyo are largely ignorant of them.

It will be the responsibility of `fsck` to do a sanity check for aliases that squat on top of existing keys (see #679 #599).

By default, Onyo repositories will be init-ed with a single default alias: `path`->`onyo.path.relative` (see #688). It is such a dominant use-case that it deserves to be the default.

Aliases should be defined via onyo configuration, so all the usual user-preference hierarchy applies (git config then onyo config).

```
[onyo "key-alias"]
	path = onyo.path.relative
	o.g = onyo.git
```

Open question:
- Can aliases use other aliases? This would be powerful, but makes lookups more complicated.
- Should `onyo config onyo.key-alias.<alias>` do a sanity check for squatting on top of existing keys? (see #679)
- When aliases are used in `onyo get`, should the header for that column use the expanded name?

aqw:
We don't print the headers in machine readable mode.

bpoldrack:
Right. Ignore that. ;-)

So, not expanding is the answer, I think. We print what's queried. If the query uses an alias, so does the response.

aqw:
> We print what's queried. If the query uses an alias, so does the response.

I'm comfortable with that. And I think it allows for another use-case where users temporarily want a different representation (could be interesting in the context of globbing). Which is kinda cool when you think about it...

> However, that makes me think `onyo list-keys` should print aliases and what they resolve to.

Good idea. I agree that should be listable.

Issue #688, opened 2024-10-22
New Feature: `onyo.` namespace for Pseudo Keys


aqw:
All Pseudo Keys, now and in the future, should live under the `onyo.` namespace. See #678 for design background.

New Structure:

- `onyo.is.asset`: `True` or `False`
- `onyo.is.directory`: `True` or `False`
- `onyo.path.relative`: path relative to repo root (currently what `path` is)
- `onyo.path.parent`: parent directory (currently what `directory` is)
- `onyo.path.file`: full path relative to repo root (relevant for asset directories and templates)

Retired:
- `directory` (now `onyo.path.parent`)
- `path` (now `onyo.path.relative`)
- `template`
  - #692 is not a blocker for this issue, and `template` can be skipped or removed entirely, whatever is easier
- `is_asset_directory`

Note:
- All pseudo keys are queryable. Ideally they always have a value (`True`/`False`, etc). If they are truly irrelevant/impossible, then `<unset>` is the appropriate value.
- Asset Directories (will return `True` for both `is.asset` and `is.directory`).
- The concept of "reserved keys" should likely be retired.
- There is a possibility that some pseudo keys will become `set`able (perhaps `onyo.path.parent` (i.e. `mv`) and `onyo.is.directory` (i.e. `mkdir`/`rmdir`))

Future (out of scope for this issue; just context):
- #662 will live under `onyo.git`
- #675 will add `onyo.is.template`: `True` or `False`
- #674 will add `onyo.is.empty`: `True` or `False` (only ever `True` for directories and possibly templates)
- #692 will add `onyo.config.*`

aqw:
I have created #692 to allow `onyo new` to temporarily override onyo configuration key values. That solves the `onyo.template` problem.

bpoldrack:
I like this approach. Want to throw in one more thing, though: `onyo.is.template`.
This isn't actually related to the template "key" you mentioned above, but to #675.

aqw:
Yup! This issue I intentionally omitted keys for new features. There is an `onyo.is.template` in #675

bpoldrack:
Generally, I do support the idea of `onyo.cli.template`. I wouldn't put in the `--`, though. While it's a reference to a command option, the weirdness of `onyo.cli.--template` isn't necessary and furthermore: There is a difference in behavior. Such keys would be meant to specify per-asset options. Which is different from the switch itself, which refers to the entire command.

aqw:
Closing as a result of Ben's recent work. :-D

Issue #662, opened 2024-10-07
Feature Request: expose git metadata as pseudo-keys


aqw:
We have long discussed making various bits of git metadata queryable, most especially the creation date and last modification date.

Git already has all this information, and would support use-cases such as #428 .

The two "must haves" are creation date and last modification date. However, there is additional metadata we could expose such as committer and author (though I have no real use-cases for it).

All keys will live under the `onyo.git` namespace (see #688).

Proposed names:
1) - `created`
    - `modified`
2) - `date_created`
    - `date_modified`
3) - `created_date`
    - `modified_date`
4) - `created_date`
    - `created_author`
    - `created_committer`
    - `modified_date`
    - `modified_author`
    - `modified_committer`

aqw:
Ideas are most welcome. :-)

aqw:
After some though, I prefer option 1. It's the only known use-case.

I dislike adding `_date` as the type, aa the returned data will actually be date-time.

If we do eventually end up adding the other fields, I propose the following:
```
created
created_author
created_committer
modified
modified_author
modified_committer
```

bpoldrack:
I would suggest to go full namespace and not use underscores:

```
onyo.git.created.time  # datetime
onyo.git.created.committer  # these could be broken up into name and email
onyo.git.created.author
onyo.git.created.commit  # commit-SHA
```

WDYT?

TobiasKadelka:
While `onyo.created_author -> onyo.git.created.author.name` and others seem a bit lenghty, I think Bens idea is easier and more intuitive for the user.

aqw:
> I would suggest to go full namespace and not use underscores:

+1 I like this.

```
onyo.git.created.time  # datetime
onyo.git.created.committer  # these could be broken up into name and email
onyo.git.created.author
onyo.git.created.commit  # commit-SHA
```

And the same stack for `onyo.git.modified`. it would be metadata about the most recent modification. No need for the full history. That's what `git`'s commands (or `onyo history`) are for.

`onyo.created_author -> onyo.git.created.author.name`

I think it's OK, but not great.

My ideas are: `onyo.creator` and `onyo.modifier` as aliases. But I'm not in love with those either.

However, we don't have alias support yet, so there's no need to decide now. And aliases will not have any guarantee of stability. Only the full name,

bpoldrack:
Implementing this is conjunction with #688, b/c it's fundamental to allow lazy loading for `onyo.git.` keys in particular.

PR #735, opened 2024-11-20
New pseudo-keys:  Part I

bpoldrack:
Sits on top of PR #707 and resolves its issue WRT `is_asset_directory` by removing that key altogether. This is now split into the new `onyo.is.asset` and `onyo.is.directory` pseudo-keys. Although for the most part `onyo.is.directory` is what is used instead.

This is the first PR of a series, aiming to implement #662, #688, #689.
It's introducing an `Item` class and a new `PSEUDO_KEYS` mapping, that add pseudo-keys to dict-likes.
The aim for this PR is:
- Make `path` and `directory` keys aliases and replace their internal usage (except for test code, where they are still used at times in their role as default aliases)
- Remove `is_asset_directory` key in favor of `onyo.is.directory` (and `onyo.is.asset`)
- Replace `DotNotationWrapper` by `Item` in most places. `DotNotationWrapper` remains valid in some places where we don't want any pseudo-keys yet (For example: Reading yaml from an edited temporary file - only becomes an `Item` when we tie it to the actual repo. We don't want pseudo-keys referring to the temp file)
- Don't break anything other than `is_asset_directory` while throwing this class in.
- It does not yet utilize `Item`'s ability to represent directories, templates, etc. - just assets in the first round.
- It does not yet implement git metadata keys, but builds on a mechanic that is meant to account for the need to lazy load pseudo-keys.
- It does not yet introduce a proper mechanic for namespaces nor streamline the usage of this class yet. I want to see what this really needs first.

The scope is limited, in order to address that topic in digestable chunks. Next up should be:
- git metadata
- actually use `Item` consistently for anything at high-level code, so this becomes the datastructure, that `Inventory` and command implementations deal with, instead of arbitrary dicts and paths.
- Build different validators (at least in terms of design). Anything is an Item, but what it's used for is what makes the difference. If I want to use an asset as a template, I simply treat it that way (and put it into a TemplateValidator). As opposed to the initial idea of the object itself determining what it is. There may be more to do in a similar fashion along this line of thought.
- consequently properly implement namespaces and allow `get` to query anything (dirs, templates)

During these further steps, we/I can figure out additional functions to ease various ways of dealing with these objects (we need to repeatedly strip or reset pseudo-keys and we have notions of "you-can't-set-these", leading to a bunch of code duplication - I want to iron that out as we get to see more access patterns)

### DIFF
--- a/onyo/cli/get.py
+++ b/onyo/cli/get.py
@@ -160,11 +160,11 @@ def get(args: argparse.Namespace) -> None:
     In addition to keys in asset contents, **PSEUDO-KEYS** can be queried and
     matched.
 
-      * ``is_asset_directory``: is the asset an Asset Directory
-      * ``directory``: parent directory of the asset relative to repo root
-      * ``path``: path of the asset relative to repo root
+      * ``onyo.is.directory``: is the asset an Asset Directory
+      * ``onyo.path.parent`` (default alias: ``directory``): parent directory of the asset relative to repo root
+      * ``onyo.path.relative`` (default alias: ``path``): path of the asset relative to repo root
 
-    By default, the results are sorted by ``path``.
+    By default, the results are sorted by ``onyo.path.relative``.
     """
     includes = [Path(p).resolve() for p in args.include] if args.include else [Path.cwd()]
     excludes = [Path(p).resolve() for p in args.exclude] if args.exclude else None
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import copy
 import logging
 import subprocess
 from pathlib import Path
@@ -19,7 +18,6 @@ from onyo.lib.command_utils import (
     print_diff,
 )
 from onyo.lib.consts import (
-    PSEUDO_KEYS,
     RESERVED_KEYS,
     SORT_ASCENDING,
     SORT_DESCENDING,
@@ -34,7 +32,9 @@ from onyo.lib.exceptions import (
     OnyoRepoError,
     PendingInventoryOperationError,
 )
+from onyo.lib.items import Item
 from onyo.lib.inventory import Inventory, OPERATIONS_MAPPING
+from onyo.lib.pseudokeys import PSEUDO_KEYS
 from onyo.lib.ui import ui
 from onyo.lib.utils import (
     deduplicate,
@@ -293,12 +293,13 @@ def _edit_asset(inventory: Inventory,
     if not editor:
         editor = inventory.repo.get_editor()
 
-    # Store original reserved keys of `asset`, in order to re-assign
+    # Store original pseudo-keys of `asset`, in order to re-assign
     # them when loading edited file from disc. This is relevant, when
-    # `operation` uses them (`Inventory.add_asset`)
-    reserved_keys = {k: v for k, v in asset.items() if k in RESERVED_KEYS}
-    disallowed_keys = RESERVED_KEYS + PSEUDO_KEYS
-    disallowed_keys.remove("is_asset_directory")
+    # `operation` uses them (`Inventory.add_asset/modify_asset`).
+    reserved_keys = {k: v for k, v in asset.items() if k in list(PSEUDO_KEYS.keys()) + ['template']}
+    disallowed_keys = RESERVED_KEYS + list(PSEUDO_KEYS.keys())
+    disallowed_keys.remove('onyo.is.directory')
+    disallowed_keys.remove('onyo.path.parent')
 
     tmp_path = get_temp_file()
     write_asset_file(tmp_path, asset)
@@ -319,15 +320,19 @@ def _edit_asset(inventory: Inventory,
         subprocess.run(f'{editor} {quote(str(tmp_path))}', check=True, shell=True)
         operations = None
         try:
-            asset = DotNotationWrapper(get_asset_content(tmp_path))
-            if 'is_asset_directory' in asset.keys():
+            tmp_asset = DotNotationWrapper(get_asset_content(tmp_path))
+            if 'onyo.is.directory' in tmp_asset.keys():
                 # special case
-                # 'is_asset_directory' currently is the only modifiable, reserved key.
+                # 'onyo.is.directory' currently is the only modifiable, reserved key.
                 # TODO: This may either need a separate category or RESERVED_KEYS to
                 #       become a more structured thing than a plain list.
-                reserved_keys['is_asset_directory'] = asset['is_asset_directory']
-            if any(k in disallowed_keys for k in asset.keys()):
+                reserved_keys['onyo.is.directory'] = tmp_asset['onyo.is.directory']
+            # Check disallowed keys before we make an `Item` of this,
+            # because that will have all pseudo-keys.
+            if any(k in disallowed_keys for k in tmp_asset.keys()):
                 raise ValueError(f"Can't set any of the keys ({', '.join(disallowed_keys)}).")
+            asset = Item()
+            asset.data = tmp_asset.data  # keep exact objects for ruamel-based roundtrip
             # When reading from file, we don't get reserved keys back, since they are not
             # part of the file content. We do need the object from reading the file to be
             # the basis, though, to get comment roundtrip from ruamel.
@@ -437,7 +442,7 @@ def onyo_edit(inventory: Inventory,
         if ui.request_user_response("Save changes? No discards all changes. (y/n) "):
             if auto_message:
                 operation_paths = sorted(deduplicate([  # pyre-ignore[6]
-                    op.operands[0].get("path").relative_to(inventory.root)
+                    op.operands[0].get("onyo.path.relative")
                     for op in inventory.operations
                     if op.operator == OPERATIONS_MAPPING['modify_assets']]))
                 message = inventory.repo.generate_auto_message(
@@ -497,7 +502,7 @@ def onyo_get(inventory: Inventory,
       `sort` dictionary). Possible values are
       `onyo.lib.consts.SORT_ASCENDING` and `onyo.lib.consts.SORT_DESCENDING`.
       If other values are specified an error is raised.
-      Default: `{'path': SORT_ASCENDING}`.
+      Default: `{'onyo.path.relative': SORT_ASCENDING}`.
       One can sort by keys that are not in the output.
 
     Raises
@@ -527,7 +532,7 @@ def onyo_get(inventory: Inventory,
     if sort and not all(v in allowed_sorting for k, v in sort.items()):
         raise ValueError(f"Allowed sorting modes: {', '.join(allowed_sorting)}")
 
-    selected_keys = selected_keys or inventory.repo.get_asset_name_keys() + ['path']
+    selected_keys = selected_keys or inventory.repo.get_asset_name_keys() + ['onyo.path.relative']
     results = list(inventory.get_assets_by_query(include=include,
                                                  exclude=exclude,
                                                  depth=depth,
@@ -535,16 +540,12 @@ def onyo_get(inventory: Inventory,
                                                  # This is useless nonsense.
                                                  match=match))  # pyre-ignore[6]
 
-    # convert paths for sorting and printing
-    for r in results:
-        r['path'] = r['path'].relative_to(inventory.root)
-
     # Note: Sorting is done before any further filtering/replacing. Therefore, one can sort by keys that aren't actually
     #       in the output of the command. This behavior is utilized in tests.
     results = natural_sort(
         assets=results,
         # pyre can't tell SORT_ASCENDING is not an arbitrary string but matches the Literal declaration:
-        keys=sort or {'path': SORT_ASCENDING})  # pyre-ignore[6]
+        keys=sort or {'onyo.path.relative': SORT_ASCENDING})  # pyre-ignore[6]
 
     # Filter results for `selected_keys` first in order to not iterate over irrelevant parts of the assets in subsequent
     # replacements.
@@ -941,7 +942,6 @@ def onyo_new(inventory: Inventory,
     ValueError
         If information is invalid, missing, or contradictory.
     """
-    from onyo.lib.consts import PSEUDO_KEYS
     from copy import deepcopy
 
     if auto_message is None:
@@ -1015,11 +1015,6 @@ def onyo_new(inventory: Inventory,
     if clone and any('template' in d.keys() for d in specs):
         raise ValueError("Can't use 'clone' key and 'template' option.")
 
-    for pseudo_key in PSEUDO_KEYS:
-        for d in specs:
-            if pseudo_key in d.keys():
-                raise ValueError(f"Pseudo key '{pseudo_key}' must not be specified.")
-
     # Generate actual assets:
     if edit and not specs:
         # Special case: No asset specification defined via `keys` or `tsv`, but we have `edit`.
@@ -1035,7 +1030,6 @@ def onyo_new(inventory: Inventory,
         # 2. start from template
         if clone:
             asset = inventory.get_asset(clone)
-            asset.pop('path')
         else:
             t = spec.pop('template', None) or template
             asset = inventory.get_asset_from_template(Path(t) if t else None)
@@ -1057,7 +1051,7 @@ def onyo_new(inventory: Inventory,
         if edit or ui.request_user_response("Create assets? (y/n) "):
             if auto_message:
                 operation_paths = sorted(deduplicate([  # pyre-ignore[6]
-                    op.operands[0].get("path").relative_to(inventory.root)
+                    op.operands[0].get("onyo.path.relative")
                     for op in inventory.operations
                     if op.operator == OPERATIONS_MAPPING['new_assets']]))
                 message = inventory.repo.generate_auto_message(
@@ -1149,7 +1143,7 @@ def onyo_set(inventory: Inventory,
         asset, their value will be overwritten. If they do not exist the values
         are added.
         Keys that appear in asset names will result in the asset being renamed.
-        The key 'is_asset_directory' (bool) can be used to change whether an
+        The pseudo-key 'onyo.is.directory' (bool) can be used to change whether an
         asset is an asset directory.
     message
         Commit message to append to the auto-generated message.
@@ -1170,8 +1164,8 @@ def onyo_set(inventory: Inventory,
     if not keys:
         raise ValueError("At least one key-value pair must be specified.")
 
-    disallowed_keys = RESERVED_KEYS + PSEUDO_KEYS
-    disallowed_keys.remove("is_asset_directory")
+    disallowed_keys = RESERVED_KEYS + list(PSEUDO_KEYS.keys())
+    disallowed_keys.remove("onyo.is.directory")
     if any(k in disallowed_keys for k in keys.keys()):
         raise ValueError(f"Can't set any of the keys ({', '.join(disallowed_keys)}).")
 
@@ -1181,10 +1175,8 @@ def onyo_set(inventory: Inventory,
                          "\n".join(non_asset_paths))
 
     for asset in [inventory.get_asset(a) for a in assets]:
-        new_content = copy.deepcopy(asset)
+        new_content = Item(asset, inventory.repo)
         new_content.update(keys)
-        for k in PSEUDO_KEYS:
-            new_content.pop(k)
         try:
             inventory.modify_asset(asset, new_content)
         except NoopError:
@@ -1198,7 +1190,7 @@ def onyo_set(inventory: Inventory,
         if ui.request_user_response("Update assets? (y/n) "):
             if auto_message:
                 operation_paths = sorted(deduplicate([  # pyre-ignore[6]
-                    op.operands[0].get("path").relative_to(inventory.root)
+                    op.operands[0].get("onyo.path.relative")
                     for op in inventory.operations
                     if op.operator == OPERATIONS_MAPPING['modify_assets']]))
                 message = inventory.repo.generate_auto_message(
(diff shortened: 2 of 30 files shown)


onyo/lib/items.py
```diff
@@ -0,0 +1,163 @@
...
+    # def fill_created(self, what: str | None):
+    #     # self.repo.git.xxx   (self.get('onyo.path.file'))
+    #     # fill in all created keys
+    #     # switch `what` -> return whatever part was requesting this
+    #     pass#raise NotImplementedError
+    #
+    # def fill_modified(self, what: str | None):
+    #     pass#raise NotImplementedError
+
+    def get_path_absolute(self):
```
aqw, line 108: Does not need to be addressed in this PR, but I think I'd prefer these to all be private methods.

They're used internally to populate the pseudokeys, but everything else uses the pseudokey namespace.

onyo/lib/items.py
```diff
@@ -0,0 +1,163 @@
...
+# - values in templates maybe to be evaluated matching expression or even plugin calls  #714
+# - Path attribute caching at git/onyo layers (is_file, etc.):
+#   We actually know that everything we get from
+#   git ls-tree is in fact a file or symlink. And we can derive
+#   dirs from that path list (.anchor). That may be a lot faster.
+#   Implement cache dict at GitRepo level.
+# - Probably/Maybe: Stop passing dicts and Path objects around. All things relevant at higher level are Items, right?
+# - suck in DotNotationWrapper instead of deriving!? Probably not, because we have asset/template specs that should
+#   behave with dot notation, but can't or even must not have pseudo-keys.
+# - What about `__eq__` (see the horrible is_equal_dict helper)?
```
aqw, line 163: You read my mind. This was also something I was hoping the new object would allow us to address.

onyo/lib/pseudokeys.py
```diff
@@ -0,0 +1,101 @@
...
+                                    implementation=partial(delegate, attribute='get_path_relative')
+                                    ),
+    'onyo.path.parent': PseudoKey(description="Path of the directory the object is in relative to the repository root.",
+                                  implementation=partial(delegate, attribute='get_path_parent')
+                                  ),
+    'onyo.path.file': PseudoKey(description="Path to the file containing an asset's YAML."
+                                            "Different from 'onyo.path.relative' in case of an asset directory.",
+                                implementation=partial(delegate, attribute='get_path_file')
+                                ),
+    'onyo.is.asset': PseudoKey(description="Whether the object is an asset.",
```
aqw, line 47: Nit:
> Whether the object is an asset.
to
> Is the item an asset.

The purpose is to consolidate vocabulary. "Whether" is just a semi-synonym for "is", and "object" for "item".

The pseudokey vocabulary was well chosen. We can/should stick with it.

onyo/lib/pseudokeys.py
```diff
@@ -0,0 +1,101 @@
...
+    'onyo.is.template': PseudoKey(description="Whether the object is a template.",
+                                  implementation=partial(delegate, attribute='is_template')
+                                  ),
+    'onyo.is.empty': PseudoKey(description="Whether the object is empty.",
+                               implementation=partial(delegate, attribute='is_empty')
+                               ),
+}
+r"""Pseudo-Keys are key names that are addressable but not in asset content.
+
+All ``PSEUDO_KEYS`` are reserved.
```
aqw, line 62: This, in the near future will change. IMO, the entire `onyo.` namespace should be reserved (and nothing else).

onyo/lib/tests/test_commands_get.py
```diff
@@ -100,21 +101,22 @@ def test_onyo_get_reserved_keys(inventory: Inventory,
                                 capsys) -> None:
     r"""`onyo_get()` allows to specify all reserved keys to query
     and display information for."""
-    from onyo.lib.consts import RESERVED_KEYS, PSEUDO_KEYS
-    reserved = ["type", "make", "model.name", "serial"] + PSEUDO_KEYS + RESERVED_KEYS
+    from onyo.lib.consts import RESERVED_KEYS
+    from onyo.lib.pseudokeys import PSEUDO_KEYS
+    reserved = ["type", "make", "model.name", "serial"] + list(PSEUDO_KEYS.keys()) + RESERVED_KEYS
```
aqw, line 106: Oh ugh. I remember this.

Something we will be sure to unroll by the end of all of this pseudo-key work.

onyo/tests/demo/reference_git_log.txt
```diff
@@ -131,37 +131,37 @@ Date:   Sun Jan 1 00:00:00 2023 +0100
     - warehouse/laptop_apple_macbook.oiw629 -> accounting/Bingo Bob/laptop_apple_macbook.oiw629
     - warehouse/monitor_dell_PH123.86JZho -> accounting/Bingo Bob/monitor_dell_PH123.86JZho
 
-commit 1890e9029d8b7560b485ae83d1ed41b136704fc0
+commit 145755d76ab2f5ea2b0d0d97d03ae562293fefd7
 Author: Yoko Onyo <yoko@onyo.org>
 Date:   Sun Jan 1 00:00:00 2023 +0100
 
-    new [1]: warehouse/headphones_apple_airpods.uzl8e1
+    new [1]: None
```
aqw, line 138: Looks like a bug snuck in.

Is this known and will be fixed in the upcoming PRs?

aqw (approved):
Overall LGTM!

It's a huge PR, so it's difficult to review in depth.

In general, I chose to not be nitpicky, as this is all clearly heading in a positive direction, and it cannot fix everything on its way there. There will be some mess.

With that in mind, I saw no showstoppers.

Please review my comments. If you want to fix any of them in this PR, you are welcome to, but it's not necessary. They can be addressed in later PRs as more pieces land.

onyo/cli/get.py
```diff
@@ -160,11 +160,11 @@ def get(args: argparse.Namespace) -> None:
     In addition to keys in asset contents, **PSEUDO-KEYS** can be queried and
     matched.
 
-      * ``is_asset_directory``: is the asset an Asset Directory
-      * ``directory``: parent directory of the asset relative to repo root
-      * ``path``: path of the asset relative to repo root
+      * ``onyo.is.directory``: is the asset an Asset Directory
+      * ``onyo.path.parent`` (default alias: ``directory``): parent directory of the asset relative to repo root
```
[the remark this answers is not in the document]
bpoldrack: Yeah, I'm fine with that. Config cache + "real" aliases first, though.

onyo/lib/commands.py
```diff
@@ -293,15 +293,16 @@ def _edit_asset(inventory: Inventory,
...
 
-    # Store original reserved keys of `asset`, in order to re-assign
+    # Store original pseudo-keys of `asset`, in order to re-assign
     # them when loading edited file from disc. This is relevant, when
-    # `operation` uses them (`Inventory.add_asset`)
-    reserved_keys = {k: v for k, v in asset.items() if k in RESERVED_KEYS}
-    disallowed_keys = RESERVED_KEYS + PSEUDO_KEYS
-    disallowed_keys.remove("is_asset_directory")
+    # `operation` uses them (`Inventory.add_asset/modify_asset`).
+    reserved_keys = {k: v for k, v in tmp_asset.items() if k in list(PSEUDO_KEYS.keys()) + ['template']}
```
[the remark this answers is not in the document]
bpoldrack: True. Missed that spot. Will fix in this PR.

onyo/lib/commands.py
```diff
@@ -251,7 +251,7 @@ def onyo_config(inventory: Inventory,
 
 
 def _edit_asset(inventory: Inventory,
```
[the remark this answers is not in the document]
bpoldrack: Agree.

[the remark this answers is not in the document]
bpoldrack: Agree.

onyo/lib/items.py
bpoldrack: Yeah, just wanted to see to what extent DotNotationWrapper or even regular dicts keep a role first (where something like that helper may be needed). In any case: At least part if not all of that helper will be simplified and go into `DotNotationWrapper/Item.__eq__`.

onyo/lib/pseudokeys.py
bpoldrack: Agree. Will adjust.

onyo/lib/pseudokeys.py
bpoldrack: I guess, aliases are going to be a dynamic part of what is reserved, though. However, that may be a thing for fsck only.

onyo/tests/demo/reference_git_log.txt
bpoldrack: Good catch, thx! Missed that one. Need to fix.

onyo/lib/commands.py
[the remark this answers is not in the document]
bpoldrack: Actually, no. Not needed, because the incoming dict is supposed to be an `Item`. Hence, aliases used to create it in the first place were already resolved.
The "supposed to be an Item" is key, though. Happens to be the bug you saw in the demo-repo log ;-)

onyo/lib/differs.py
```diff
@@ -37,7 +37,7 @@ def diff_path_change(src: Path, dst: Path) -> Generator[str, None, None]:
 
 def diff_moved_asset(asset_old: dict | Path, asset_new: Path) -> Generator[str, None, None]:
     # could be same. Just check isinstance?
-    yield from diff_path_change(asset_old if isinstance(asset_old, Path) else asset_old.get('path'),
+    yield from diff_path_change(asset_old if isinstance(asset_old, Path) else asset_old.get('onyo.path.absolute'),
```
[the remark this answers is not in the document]
bpoldrack: Agree, but I'm not changing that yet, b/c that'll be inconsistent. We currently use absolute paths and they are only sometimes coming from an asset dict. For other operations the paths are the only thing passed around.
So, this can be consistently changed in one go, when we stop passing plain paths, but `Item`s instead.

onyo/lib/commands.py
```diff
@@ -1334,20 +1326,20 @@ def onyo_unset(inventory: Inventory,
         raise ValueError("The following paths aren't assets:\n%s" % "\n".join(non_asset_paths))
     if any(k in inventory.repo.get_asset_name_keys() for k in keys):
         raise ValueError("Can't unset asset name keys.")
-    if any(k in RESERVED_KEYS + PSEUDO_KEYS for k in keys):
-        raise ValueError(f"Can't unset reserved or pseudo keys ({', '.join(RESERVED_KEYS + PSEUDO_KEYS)}).")
+    # TODO: Actual Namespaces! Key must not start with `onyo.` What about aliases?
+    if any(k in RESERVED_KEYS or k.startswith("onyo.") for k in keys):
+        raise ValueError(f"Can't unset reserved keys ({', '.join(RESERVED_KEYS)}) "
```
aqw, line 1331: I think this might be too strict.

I agree that our current pseudo-keys under `onyo.` cannot be unset. Especially since we decided that we won't allow creation/deletion through set/unset. Just modifying within the guardrails.

But the reserved keys:
```
RESERVED_KEYS = ['template', 'onyo'] + list(PSEUDOKEY_ALIASES.keys())
...
PSEUDOKEY_ALIASES: Dict[str, str] = {
    'path': 'onyo.path.relative',
    'directory': 'onyo.path.parent',
}
```

Any aliases should be translated before they hit this far into `onyo_unset()`. And if they somehow do get this far, I don't see the problem of unsetting keys that match.

And a similar argument for `template`. It should never exist in the asset content. But if it does somehow get into assets, I don't see why we should disallow unsetting it.

onyo/lib/pseudokeys.py
```diff
@@ -0,0 +1,101 @@
...
+                                   implementation=partial(delegate, attribute='is_directory')
+                                   ),
+    'onyo.is.template': PseudoKey(description="Is the item a template.",
+                                  implementation=partial(delegate, attribute='is_template')
+                                  ),
+    'onyo.is.empty': PseudoKey(description="Is the directory empty. <unset> if the item is not a directory.",
+                               implementation=partial(delegate, attribute='is_empty')
+                               ),
+}
+r"""Pseudo-Keys are key names that are addressable but not in asset content.
```
aqw, line 60: I would adjust this just a little bit:

> Pseudo-Keys are key names that are addressable but not in asset content.

to 

> Pseudo-Keys are keys that are addressable but not written to disk as YAML.

Or something like that. it actually is a bit difficult to phrase right, since `onyo.is.directory` can be written to disk (just not as YAML). And using the word "file" is correct, but gets confusing WRT asset directories.

My goal was to make it clear that it was on-disk. The asset content in-memory clearly has these keys.

onyo/lib/pseudokeys.py
```diff
@@ -0,0 +1,101 @@
...
+    This Callable is expected to have a single parameter of type `Item`.
+    """
+
+
+def delegate(self: Item, attribute: str, *args, **kwargs):
+    # This is to avoid circular imports ATM.
+    return self.__getattribute__(attribute)(*args, **kwargs)
+
+
+PSEUDO_KEYS: Dict[str, PseudoKey] = {
```
aqw, line 33: Nit. And I'm not sure if it matters, but this list is not alphabeitized.

I understand they're sorted so that similar things are near each other. But we will be adding quite a few more keys. Eventually, alphabetized is going to be the only sane way to find anything.

onyo/lib/consts.py
```diff
@@ -1,20 +1,16 @@
...
-r"""Key names that are addressable but not in asset content.
-
-All ``PSEUDO_KEYS`` are reserved.
-
-See Also
---------
-RESERVED_KEYS
-"""
-RESERVED_KEYS = ['directory', 'is_asset_directory', 'template']
+RESERVED_KEYS = ['template', 'onyo'] + list(PSEUDOKEY_ALIASES.keys())
```
aqw, line 10: More "talking out loud" rather than anything else.

I'm not sure if aliases should be reserved. I just think they should be translated.

Though I see this is a nice way to make sure they don't somehow leak into the asset content un-translated.

I'm split.

Part of me wants them to just translate on the way in, and fsck does the actual checks.

onyo/lib/differs.py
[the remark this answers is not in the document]
aqw: Thanks for the explanation!

Can an explicit TODO be added to note this?

onyo/lib/inventory.py
```diff
@@ -304,15 +305,17 @@ def add_asset(self, asset: dict | UserDict) -> list[InventoryOperation]:
         self.raise_required_key_empty_value(asset)
         name = self.generate_asset_name(asset)
 
-        if asset.get('is_asset_directory', False):
-            # 'path' needs to be given, if this is about an already existing dir.
-            # Otherwise, a 'directory' to create the asset in is expected as with
+        if asset.get('onyo.is.directory', False):
```
aqw, line 308: `asset['onyo.is.directory']` should always return `True` or `False`, so the `get` doesn't provide any value here, except possibly obscuring some bugs.

Something to keep in mind when the later PRs come and start using these new pseudokeys internally.

onyo/lib/inventory.py
```diff
@@ -304,15 +305,17 @@ def add_asset(self, asset: dict | UserDict) -> list[InventoryOperation]:
...
+            # 'onyo.path.absolute' needs to be given, if this is about an already existing dir.
+            # Otherwise, a 'onyo.path.parent' to create the asset in is expected as with
```
aqw, line 310: Nit: this comment should be in the next if statement, where it happens.

onyo/lib/items.py
```diff
@@ -0,0 +1,167 @@
...
+def resolve_alias(key: Any) -> Any:
+    """Return the target of a key alias."""
+    try:
+        return onyo.lib.pseudokeys.PSEUDOKEY_ALIASES[key]
+    except KeyError:
+        return key
+
+
+class Item(DotNotationWrapper):
+    """An Item an Inventory can potentially track.
```
aqw, line 32: Interesting word "potentially".

onyo/lib/items.py
```diff
@@ -0,0 +1,167 @@
...
+            self.update(**kwargs)
+
+    def __setitem__(self, key, value):
+        key = resolve_alias(key)
+        super().__setitem__(key, value)
+
+    def __getitem__(self, key):
+        key = resolve_alias(key)
+        value = super().__getitem__(key)
+        if key in onyo.lib.pseudokeys.PSEUDO_KEYS and \
```
aqw, line 70: This seems to do a lookup every time, with no caching.

onyo/lib/onyo.py
```diff
@@ -637,29 +638,22 @@ def get_asset_content(self,
...
             raise NotAnAssetError(f"{path} is not an asset path")
         try:
-            if self.is_inventory_dir(path):
-                # It's an asset and an inventory dir -> asset dir
-                a = get_asset_content(path / self.ASSET_DIR_FILE_NAME)
-                a['is_asset_directory'] = True
-            else:
-                a = get_asset_content(path)
-                a['is_asset_directory'] = False
+            ## Where do we make sure to distinguish onyo.path.file from onyo.path.relative?
```
aqw, line 641: This sounds like a TODO. Can it be marked as such? Otherwise we will need to read all comments to make sure they don't contain unmarked TODOs.

aqw (approved):
LGTM!

I took the opportunity to review the non-test code more closely, as it's my second read through.

I left comments. None are blockers. Interested in your feedback. :-)

onyo/cli/get.py
```diff
@@ -160,11 +160,11 @@ def get(args: argparse.Namespace) -> None:
...
+      * ``onyo.path.relative`` (default alias: ``path``): path of the asset relative to repo root
```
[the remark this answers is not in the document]
bpoldrack: Yes, because this is just what it was before. The additional query abilities that this effort should ultimately lead to aren't tested yet. Also: Not sure that `file` works properly yet (probably not). 
In any case: I'd like to change/test the querying properly when the approach is applied to directories and templates that are now able to have pseudo-keys as well.

So, didn't want to put it into documentation before that's actually done and tested, even though at least `absolute` should already work.

onyo/lib/commands.py
bpoldrack: Yes, this is just to not break stuff yet. Ultimately, the concept of reserved/disallowed needs to change (it's context-dependent) anyway. `onyo` is just a crude measure to account for the top-level key and `template` is supposed to become `onyo.cli.template` and would then not be part of the dict indeed.

So, in short: Agree, this needs to change, but only after there's a better way to specify namespaces (`onyo.` in this case) and `template` has moved to where it now belongs.

onyo/cli/get.py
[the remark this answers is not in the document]
aqw: That makes sense. Thanks for explaining. :-)

onyo/lib/pseudokeys.py
bpoldrack: True. It's actually just taken from the previous `PSEUDO_KEYS` definition in `consts.py`.

> since onyo.is.directory can be written to disk (just not as YAML).

Yeah, and I think we said, we'd allow for several pseudo-keys to be set in templates. So, it's "on disc" and "in YAML", just not in an asset file.
So,

"Pseudo-Keys are key names that are addressable but not written to asset files."?

onyo/lib/inventory.py
bpoldrack: Nope.  That's the ultimate values for an existing thing in the inventory, yes. But when a new asset is created as an `Item`/dict and passed to `Inventory.add_asset` this property may not be specified (at least as of now) and can't be derived from the state of things on disk, because it doesn't exist yet. It's this very function that says "well, if it wasn't specified that counts as `False`".
We may want to change that and require this to be set by `onyo_new` instead, before it's passed here. Not opposing that, it's just not (yet) what's happening.

aqw: Ahh. Thanks for the explanation.

onyo/lib/inventory.py
bpoldrack: True. Will change.

onyo/lib/items.py
bpoldrack: :-D

Don't have a good phrasing yet. Point being, that this thing may be the result of `new --keys ..`, piped in YAML, merged templates etc. that are a state in memory that isn't any committed thing yet.

onyo/lib/items.py
bpoldrack: Do you mean WRT aliases or the testing for whether this is still `PseudoKey`?
The former: Yes, that'll need to rely on config caching.
The latter: Would be nice to cache, but not yet sure how. That's because I'm not yet sure how to properly do the various ways of stripping/resetting/ignoring pseudo-keys in different operations/commands. These patterns will hopefully lead to proper `Item` methods, that do that. Once that's clear, we know how to cache (and how o invalidate it)

onyo/lib/onyo.py
bpoldrack: Agree. Will do.

onyo/lib/pseudokeys.py
bpoldrack: Right. Will do. It does matter (kinda), simply because the main reason I wanted this kind of mapping in the first place, is the single entry point for humans.

onyo/lib/commands.py
```diff
@@ -251,7 +251,7 @@ def onyo_config(inventory: Inventory,
 
 
 def _edit_asset(inventory: Inventory,
-                asset: dict | UserDict,
+                tmp_asset: dict | UserDict,
```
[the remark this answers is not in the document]
bpoldrack: Thanks! I did in fact mean to only have the thing that's read from the temp file renamed. Will fix.

onyo/lib/items.py
aqw: > Do you mean WRT aliases or the testing for whether this is still PseudoKey?

Neither. I mean, once we've looked up the pseudo-key value once (such as `onyo.is.directory`), we don't store it, and will execute it again if requested again.

I can understand not doing this now, and waiting until more pieces land to determine if caching such lookups is practical. But yeah, that's what I was pointing out.

onyo/lib/consts.py
aqw: > However, the way RESERVED_KEYS is used ATM, this all needs to be in there. So, this isn't meant to be the final form, but the intermediate definition that allows this PR to not break shit. We reserved these keys before this PR and after. Just not at the end of the series.

Cool. I'm fine with that then. Last time I looked seriously into this, I found just how intertwined the logic is. So +1 on doing this piece by piece.

onyo/lib/items.py
bpoldrack: Have you seen line 75 below? It is stored. Not a cache exactly, but that's probably not the point.

onyo/lib/commands.py
```diff
@@ -319,15 +320,19 @@ def _edit_asset(inventory: Inventory,
         subprocess.run(f'{editor} {quote(str(tmp_path))}', check=True, shell=True)
         operations = None
         try:
-            asset = DotNotationWrapper(get_asset_content(tmp_path))
-            if 'is_asset_directory' in asset.keys():
+            tmp_asset = DotNotationWrapper(get_asset_content(tmp_path))
+            if 'onyo.is.directory' in tmp_asset.keys():
                 # special case
-                # 'is_asset_directory' currently is the only modifiable, reserved key.
+                # 'onyo.is.directory' currently is the only modifiable, reserved key.
```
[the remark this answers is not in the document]
bpoldrack: Semi-allowed at this point. I have neither fully implemented setting other pseudo-keys nor tested it. So, currently it would be settable via `--keys` but not from within the editor (like before). IOW: It's already stripped to not have it listed in the error message on setting disallowed keys, but if you set it in the edited YAML it would be ignored.

So, just a consequence of the intermediate state right now. Comment here is in-line with the code it's about. But that code will change.

onyo/lib/items.py
aqw: The magic for me is in line 71. I understand it now. Thanks!

Issue #506, opened 2023-11-15
Tests + docstring for `onyo_get()`


bpoldrack:
Properly document `onyo/lib/commands.py:onyo_get` and provide tests in `onyo/lib/tests/test_command_get.py`.
Based on that reworked docstring, provide a respective command line version in `onyo/commands/get.py`.

bpoldrack:
This requires #504 to be merged first.

Issue #488, opened 2023-11-08
Tests + docstring for `onyo_mkdir()`


bpoldrack:
Properly document `onyo/lib/commands.py:onyo_mkdir` and provide tests in `onyo/lib/tests/test_command_mkdir.py`.
Based on that reworked docstring, provide a respective command line version in `onyo/commands/mkdir.py`.

Issue #487, opened 2023-11-08
Tests + docstring for onyo_mv


bpoldrack:
Properly document `onyo/lib/commands.py:onyo_mv` and complete its tests in `onyo/lib/tests/test_command_mv.py`.
This is partially done but incomplete.
Based on that reworked docstring, provide a respective command line version in `onyo/commands/mv.py`.

bpoldrack:
This should also address #445

PR #743, opened 2025-01-30
DOC/RF: uplift docstrings of the remainder of `commands.py`

aqw:
This PR is focused on docstrings, but does include a few other things:

- uplift all remaining docstrings in `commands.py`
- uplift all docstrings in `consts.py`
- reword/remove comments as necessary
- refactor as needed

Fixes:
- #486
- #487
- #488
- #490 
- #506


onyo/cli/get.py
```diff
@@ -21,8 +21,8 @@
         required=False,
         default=0,
         help=r"""
-            Descend up to **DEPTH** levels into the directories specified. A
-            depth of **0** descends recursively without limit.
+            Descend up to **DEPTH** levels into the directories specified by
+            ``include``. A depth of **0** descends recursively without limit.
```
TobiasKadelka, line 25: Not sure if we want to mention both, but `--path` is an alias for `--include` and works the same way.

onyo/lib/commands.py
```diff
@@ -89,49 +89,43 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
...
-    By default, the following tests are performed:
+    The following tests are available:
 
-    * ``anchors``: verify that all directories (outside of ``.onyo/``) have an
-      ``.anchor`` file
-    * ``asset-unique``: verify that all asset names are unique
-    * ``asset-yaml``: verify that all asset contents are valid YAML
-    * ``clean-tree``: verify that git has no changed (staged or unstaged) or
-      untracked files
+    * ``anchors``: directories (outside of ``.onyo/``) have an ``.anchor`` file
```
TobiasKadelka, line 96: I think somewhere here `.gitignore` needs to be mentioned.

onyo/lib/commands.py
```diff
@@ -254,48 +245,40 @@ def _edit_asset(inventory: Inventory,
...
 
-    Utility function for `onyo_edit` and `onyo_new(edit=True)`.
-    This is editing a temporary file initialized with `asset`. Once
-    the editor is done, `asset` is updated from the file content and
-    `operation` is tried in order to validate the content for a
-    particular purpose (Currently used: Either `Inventory.add_asset`
-    or `Inventory.modify_asset`).
-    User is asked to either keep editing or accept the changes
-    (if valid).
+    The asset content is edited as a temporary file. Once the editor is done,
```
TobiasKadelka, line 252: I find this explanation sounds incomplete. It mentions temporary files that get edited, and that one can "accept" after being presented the diff with the promt. This as a whole makes it sound as if one accepts changes for the temporary file.

It should either mention:
1. the changes are applied to the original file
2. fully explain the four options, e.g. "skip" vs. "abort" as an option is not clear.

Nowhere it says that multiple assets can be edited at once, which makes these options even more unclear.

onyo/lib/commands.py
```diff
@@ -254,48 +245,40 @@ def _edit_asset(inventory: Inventory,
...
-      expected to raise, if the edited asset isn't valid for that
-      purpose.
-
-    Returns
-    -------
-    dict
-      The edited asset.
+        Function to validate the edited asset, which shall raise if the asset
+        isn't valid for that purpose.
+        In practice, see ``inventory.py``.
```
TobiasKadelka, line 267: "In practice" sounds super weird. No idea what it should tell me, or how it would be different "in theory".
Maybe "see ``inventory.py``." would be enough, or if this is important, "In practice" should be replaced for something more explanatory.

onyo/lib/commands.py
```diff
@@ -471,57 +450,47 @@ def onyo_get(inventory: Inventory,
...
     Parameters
     ----------
     inventory
-      The inventory to query.
+        The Inventory to query.
     include
-      Limits the query to assets underneath these paths.
-      Paths can be assets and directories.
-      If no paths are specified, the inventory root is used as default.
+        Assets or directories to query.
```
TobiasKadelka, line 460: `onyo get --path shelf warehouse/desktop_dell_D11S.G7MPSM2 --keys path` works just fine, i.e. one can mix assets and directories.

So I think instead of "Assets *or* directories" this should say "Assets *and* directories"

onyo/lib/commands.py
```diff
@@ -471,57 +450,47 @@ def onyo_get(inventory: Inventory,
...
+        Default: inventory root
     exclude
-      Paths to exclude, meaning that assets underneath any of these are not
-      being returned. Defaults to `None`. Note, that `depth` only applies to
-      `include`, not to `exclude`. `depth` and `exclude` are different ways
-      of limiting the results.
+        Assets or directories to exclude from the query.
```
TobiasKadelka, line 463: Here too "Assets and directories" seems more correct.

onyo/lib/commands.py
```diff
@@ -471,57 +450,47 @@ def onyo_get(inventory: Inventory,
...
     depth
-      Number of levels to descend into. Must be greater or equal 0.
-      If 0, descend recursively without limit.
+        Descend up to **depth** levels into the directories specified by
+        ``include``. A depth of ``0`` descends recursively without limit.
```
TobiasKadelka, line 466: `--include` and `--path` work both.

onyo/lib/commands.py
```diff
@@ -471,57 +450,47 @@ def onyo_get(inventory: Inventory,
...
+        separate values with a single tab) vs more human-friendly output
+        (headers and padded whitespace to align columns).
     match
-      Callables suited for use with builtin `filter`. They are
-      passed an asset dictionary and expected to return a `bool`,
-      where `True` indicates a match. The result of the query
-      consists of all assets that are matched by all callables in
-      this list. One can match keys that are not in the output.
+        Callables suited for use with builtin :py:func:`filter`. They are passed
+        an asset dictionary and expected to return a ``bool``.
```
TobiasKadelka, line 473: "They are passed an asset dictionary" sounds incorrect. Maybe it should have been "passed as an"?

TobiasKadelka (commented):
I left some suggestions in regards to phrasing as well as information provided.
Most of it are just details, but since some doc-strings are parsed into the documention and `--help` text they need be as clear and explanatory as possible :)

onyo/cli/get.py
aqw: IMO that would increase noise and reduce the comprehension. `--path` is indeed that, just an alias.

onyo/lib/commands.py
aqw: At this layer there is no alias.

onyo/lib/commands.py
aqw: In this case, the wording is correct: an Asset Dictionary is passed to the Callable.

TobiasKadelka: Ohh, this means "They get an asset dictionary passed"? Somehow I needed to re-read it 5 more times to get it. The active phrasing of "are passed", even though "They" (the callable) is receiving them passive did through me off.

onyo/lib/commands.py
aqw: Done.

onyo/lib/commands.py
aqw: I have added a sentence detailing how the temporary file works.

However, the rest of the comments don't really apply, as `_edit_asset()` is a helper function, and is non-user-facing. And because it's a private function, it's not even included in RTD rendering. It's documented to for developer convenience.

> Nowhere it says that multiple assets can be edited at once, 

Because this function only accepts one asset at a time.

onyo/lib/commands.py
aqw: I have replaced this with a direct reference to an example Operator.

onyo/lib/commands.py
aqw: This is a nitpick, because `or` is ambiguous in english. You're reading it as `xor`.

I have changed this (and a dozen other instances of this, across the project) to `and/or` for the sake of explicitness.

bpoldrack:
Thanks much!

Issue #709, opened 2024-10-30
Bug: renaming and moving a directory results in confusing commit subject


aqw:
I'm not sure if there's more than one misbehavior here, but at minimum, the commit subject is confusing.

Let's say I move a directory and rename it at the destination, such as:
```
onyo mv groups/dp/Alex\ Waite groups/psyinf/Alex\ Stellfox
```

This results in a commit subject of:
```
mv + ren [2]: Alex Waite,Alex Waite -> Alex Stellfox
```
The `Alex Waite,Alex Waite -> Alex Stellfox` is just weird. I /think/ the first `Alex Waite` is for `mv` (which is meaningless), and the second is for the `ren` (which is different than normal `ren` messages). I would like to see a rendering such as the one outlined in #708. However, I don't know if that's reasonable to expect if we indeed consider this to be two different operations.

In short, should this be considered one or two operations?
1) if one, then this should render like any other rename (which perhaps #708 will improve upon)
2) if two, I don't have a suggestion yet, and we need to some up with a more meaningful message

TobiasKadelka:
I think you meant to provide an example for which the directory changes, too? Because `onyo mv groups/dp/Alex\ Waite groups/dp/Alex\ Stellfox` does not move the directory, it just renames it. For just renaming, it is no move, I only get `ren [1]: groups/dp/Alex\ Waite groups/dp/Alex\ Stellfox`.

`onyo mv groups/dp/Alex\ Waite groups/sbb/Alex\ Stellfox` leads to 
`mv + ren [2]: Alex Waite,Alex Waite -> Alex Stellfox`

For this, I personally do think two operations makes sense. It is only on the terminal/computer level that it is one operation, but in real life this a rather rare situation, and I think it will almost always be caused by two things happening at once. (The only exception might be if we completely would change our whole inventory structure maybe, e.g. rename groups, merge them, change from the current `spaces/` to something different. But I do not think that in every day operation the two things happen often at once)

--------------

For assets, I really do think this should be considered and displayed as two operations in the history.

When I reflect on when this would happen, it seems always like a short form of we received the device, changed/repaired it, and gave it out to a new person, or some other action like that, that would describe multiple actions in real life.

So in summary, I think a rename+move in one call does reflect two actions, and should be displayed as such.
I only think we should make the commit message subject more meaningful, for example even a `groups/dp/Alex\ Waite -> groups/sbb/Alex\ Stellfox` feels more correct an meaningful.

aqw:
[...]

Issue #708, opened 2024-10-30
New Feature: shorter commit subject for renames


aqw:
When `onyo mv` currently performs a rename, it generates a commit subject like the following:
```
ren [1]: groups/dp/Alex Waite -> groups/dp/Alex Stellfox
```

`tig`'s UI summarizes this as:
```
groups/dp/{Alex Waite => Alex Stellfox}
```

I propose that onyo do something very similar for the commit subject:
```
ren [1]: groups/dp/{Alex Waite -> Alex Stellfox}
```

I prefer this because:

1) it's shorter, and space is at a premium in the subject
2) it highlights what part of the path changed

Implementation:
- if there overlap in the path names, use `{}` to group the differences
  - separate with `->`
  - e.g. `groups/{dp/Alex Waite -> psyinf/Alex Stellfox}/`
- if there's no overlap in the path names, don't use any `{}`
  - e.g. `shelf -> bucket` (i.e. current behavior)

Other commit subject related issue: #475

TobiasKadelka:
This idea makes a lot of sense. I really value our subject lines and they are one of the main things I use to quickly identify what action was when performed, and this idea only makes it easier.

And I had to have a look, it also fits into what we already have for a moving (instead of renaming) `onyo mv`:

`mv [1]: repair/workbench/laptop_apple_mbp2021.serial -> shelf`

Here it is also somewhat shortened, we just state the new directory, not the whole destination path again.

bpoldrack:
> I propose that onyo do something very similar for the commit subject:

I agree. And I would suggest to use the exact same pattern of shortening for `mv`. In some sense it only looks similar now in some cases by accident - it doesn't really work that way yet.

aqw:
The pattern works great for `mv` when moving only one asset. Multiple assets won't really work.

Issue #475, opened 2023-11-03
Find an alternative short-form for devices for `generate_commit_message()`


TobiasKadelka:
#462 points out that the flexible name schemes for assets mean that `paths = [x.name.split('_')[0] for x in path` do not necessarily lead to the `type` name field, this is just based on our current default implementation.

While not high priority, I wonder if there is not a reasonable replacement for this; e.g. selecting the first element of `onyo.assets.filename` which defines the filename for assets, and/or the biggest overlap in the name of all things the list of parameters given into `generate_commit_message()` or something like that.

I do not see a way that always works ATM, but somebody else might come up with a reasonable alternative :)

aqw:
I recently rediscovered this issue, and this functionality is really from a different era of Onyo. Nowendays, no particular key has any more meaning than others. So specifying a particular key for the summarize to use when the subject is too long... it's pretty arbitrary and I don't think will work terribly well in practice.

If I make a change to many (let's say 42) assets, every single asset is listed in the subject. That does feel... intense.
```
set [42] (type): laptop_apple_mba2022... [42 assets are listed here]
```

I see a few options:
1) leave the behavior as is; it's not hurting anything
  - it's also not useful in any way; the Operations Summary is much more useful
2) naively decide what "too long" is and simply say "42 assets"
  - this feels pointless, as `set [42] (type)` already contains all that information
3) attempt some form of reduction of the asset name:
  - `set [42] (type): laptop_{apple,framework,lenovo,toshiba}_{16}_{42}`
  - short lists (some char length?) will list all fields, longer ones will say the number of variants
  - This doesn't feel all that much better. The primary improvement is that I can review how much variance there is.

If we can't design something that's clearly an improvement, then we should close this issue.

Thoughts?

TobiasKadelka:
> If we can't design something that's clearly an improvement, then we should close this issue.

Lol, I just went to old issues with the goal to find at least one old issue that can be closed. Seeing your comment seems to justify my believe that this one is indeed out of date.

In praxis, I did experience this too (we have some very long commit messages), but it does not hurt, and there is still no better way this could be done. I will close here :)

aqw:
I'm going to reopen this, because I would like to explore option 3 more before we abandon it.

aqw:
My problem here is that the subject is filled with redundant information. It's all repeated in the body, and in a much better format for both humans and scripts.

[...]

PR #745, opened 2025-02-02
NF/BF: support inline-diffs for `onyo mv`

aqw:
It now generates more explicit and visually meaningful diffs between two paths (see #708).

Multi-path moves are not affected by this change.

Fixes:
- #708
- #709

Part of the broader effort for #475

### DIFF
--- a/onyo/lib/command_utils.py
+++ b/onyo/lib/command_utils.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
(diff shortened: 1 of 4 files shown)


aqw:
There's a good chance that this conflicts with #743.

That PR should be merged first, and I will rebase this one as necessary.

bpoldrack (approved):
Nice! Thanks!

Issue #732, opened 2024-11-13
New Feature: `onyo new --yaml` (with supporting flags)


aqw:
This is a WIP. Saving now. Will finish later:

----

This issue is about adding the `--yaml` flag to `onyo new`, but it has multiple knock-on effects that need to be addressed at the same time.

I have made the following modifications from our whiteboarding session:
- I inverted `--yaml-preserve-comments` to `--yaml-ignore-comments` 
- Comments are overwritten like key-values. When there's conflicts, then the new comments are used. This makes `--yaml-ignore-comments` more meaningful. And one can use `--yaml` with comments to mass update comments.
- `--recursive` no longer exists. If one wants to use a template dir without children, then `--yaml` is the only way to do it. I can't imagine a non-asset/template-dir passed to --yaml that needs to be recursed into. And the problems of multiple arguments passed to `--yaml`.

Relevant parts of `onyo new --help`:
```
Asset contents are populated in a waterfall pattern and can overwrite
values from previous steps:

  1) --clone or --template
  2) --yaml
  3) --keys
  4) --edit (i.e. manual user input)
```
```
  -c, --clone CLONE
                        Path of an asset to clone. The equivalent of --recursive --yaml.
                        
                        This cannot be used with the --template flag nor the
                        onyo.new.template Pseudo Key.
  -t, --template TEMPLATE [...]
                        Path of a template to populate the contents of new assets.
                        The equivalent of --recursive  --yaml .onyo/templates/.
                        
                        This cannot be used with the --clone flag nor the
                        onyo.new.template Pseudo Key.
  -y, --yaml YAML
                        Path to a YAML file (or asset or template) to populate
                        the contents of new assets.
                        
                        Comments are set on a best-effort basis. Any merge
                        conflicts will use the new comment to overwrite the
                        previous comment. See also --yaml-ignore-comments.
  -i, --yaml-ignore-comments
                        Disable setting comments.
  -r, --recursive
                        Recurses through all child assets, templates, and directories.
                        Behaves as if it's a multi-document YAML file.

                        Can only be used with the --yaml flag.
```

Tasks:
- [ ] `--template` only operates on templates, and can accept multiple templates
- [ ] `--template` is only relative to the template dir. It no longer tries CWD as a fallback.
- [ ] `--clone` only operates on assets, and only one asset
- [ ] `--yaml` (and thus `--template` and `--clone`) can accept multi document YAML files (see #724)
[...]

Issue #703, opened 2024-10-25
Design: allow `onyo new` to accept values in YAML files


aqw:
`onyo new` can have values passed into it via `--template`, `--keys`, or `--tsv`. However, despite generating a YAML file, there is no way to pass a YAML file into `onyo new`.

Except one can, by abusing `--template`. A template does not need to be in `.onyo/templates`. One could pass a YAML file to `--template` and proceed from there.

The downside is that one then cannot use a template.

The current two favorite options are:
1) add a flag (such as `--yaml`) to accept YAML file(s)
2) allow multiple templates to be provided, allowing them to layer

Option 1 is more powerful as a batch import (basically alternative TSV).
Option 2 provides interesting new capabilities for templates, to either use them as lego blogs (assembling components) or layers to selectively overwrite or redact values.

These are two very different directions, and I don't think we should pursue both.

Open Questions:
- what to do about empty values?
  - `--keys` requires both key and value; what if we have a key with a value in the template and the key without a value in the YAML file? Does it retain or overwrite the value with empty?
- can multiple YAML files be provided?
  - can multiple YAML documents be provided inside of one file? (separated by `---`)

TobiasKadelka:
I agree that option 2 makes most sense.

I reflected on the "multiple templates" idea over the weekend. And I began to wonder if for situations when the user enters 2 templates with conflicting information (i.e. in asset 1 `key: value1`, and in asset 2 `key: value2`) if it would be possible to do basically what git does for merge conflicts.

As an example, I imagine this call: `onyo new --template yaml1 yaml2 --edit` opens the editor with:
```
type: laptop
make: apple
model: mbp
serial: 1234

<<<<
key: value1
====
key: value2
>>>>
```

aqw:
Oof. I am pretty against the merge conflict syntax. That means we generate a document that is invalid YAML. Loads of issues stem from that. And users who are only casually familiar with `git` will be confused.

I'd prefer a clear behavior of inheritance and that we just follow those rules.

TobiasKadelka:
> I'd prefer a clear behavior of inheritance and that we just follow those rules.

This makes sense to me, but I instantly see some problems. Even if we have some solution like "overwrite keys by order of yaml files given", I am absolutely sure that during the next weeks/months at least somebody in our group will accidently overwrite something while using two files, and we will not notice.
I do not think this is really possible to solve with informing people and good documentation. I can see you making the case to go that route, but I wanted to share my perspective.

[...]

PR #747, opened 2025-02-04
NF/TST/DOC: add `onyo tsv-to-yaml`

### LINKED ISSUE #731: New Feature: `onyo tsv-to-yaml`
A dedicated command to convert tabular data (TSV and CSV) into YAML. It will supplant `onyo new --tsv`.

Usage:
```
onyo tsv-to-yaml [-h] [-d, --delimiter] TSV [TSV ...]

Convert tabular files (e.g. TSV, CSV) to YAML files suitable for passing
to onyo new and onyo set.

The header declares the key names to be populated. The values to
populate assets are declared with one line per asset.

The output is printed to stdout as a multiple document YAML file
(each document is separated by a `---` line).

Positional Arguments:
  TSV
                        Paths of tabular files to convert to YAML.

Options:
  -h, --help
                        show this help message and exit
  -d, --delimiter
                        Delimiter for the tabular file. Default is tab.
```

The output should be printed to the terminal. The user can redirect to a file or pipe the YAML as needed.

Notes:
- The design issue for this feature is #722.
- This depends on #724 (will turn into a feature request)

aqw:
This PR removes TSV support entirely from `onyo new` and creates a dedicated command (creatively named `tsv-to-yaml`) to convert TSV files into YAML.

The result is a substantial reduction in the (horrendous) complexity of `onyo new`.

A small enhancement is added to `DotNotationWrapper` to allow ingesting dicts that have keys in dotnotation and we wish to be interpreted.

This PR does /not/ add the feature to ingest YAML files with `onyo new` (see #703 and #732). That is coming soon, and this PR is intended to make that effort easier.

Fixes #731


bpoldrack (approved):
LGTM. Love the amount of deletion this comes with. ;-)

I would have preferred something like this over an init-switch:

```
class DotNotationWrapper 
    ...
    @classmethod
    def from_dot_keys(cls, d: dict) -> DotNotationWrapper:
        ...
```
because `__init__` now does two conceptually quite different things.
But it's fine - mostly a matter of taste really.

aqw:
> LGTM. Love the amount of deletion this comes with. ;-)

As do I. It's beautiful. :-D 

> I would have preferred something like this over an init-switch:
> 
> ```
> class DotNotationWrapper 
>     ...
>     @classmethod
>     def from_dot_keys(cls, d: dict) -> DotNotationWrapper:
>         ...
> ```
> 
> because `__init__` now does two conceptually quite different things. (Also: Would be easier to enhance that, so that this way of interpreting the input would apply to any form of input - kwargs, sequence of key-value, dict ...) But it's fine - mostly a matter of taste really.

Yeah, I had originally written something like that. And I ended up moving to the `__init__` solution. For two reasons:
1) `from_dot_keys()` is the exact same as `set` in a for-loop. It's no enhancement.
2) `__init__` allows passing a dict. In this case, we want to do exactly that, but disable the retention of the pristine dict passed in.

In the end, it felt like the most appropriate place to put the functionality.

If you feel strongly about it, I'm willing to hear more and be persuaded. But if it's a matter of taste, then I suggest that we use this opportunity to celebrate our differences. ;-)

bpoldrack:
> If you feel strongly about it, I'm willing to hear more and be persuaded. But if it's a matter of taste, then I suggest that we use this opportunity to celebrate our differences. ;-)

Let's celebrate :)

PR #748, opened 2025-02-06
RF: `Item()`-ize `onyo_mkdir()` and `inventory.add_directory()`

aqw:
This takes to approach of: minimal changes to make things work in areas of the code that are not ready to be flipped fully over to `Item()` yet.

This will lead to a bit of churn, but is the only rational way to convert the codebase incrementally rather than all at once.

### DIFF
--- a/onyo/lib/inventory.py
+++ b/onyo/lib/inventory.py
@@ -357,16 +357,33 @@ class Inventory(object):
                 asset['onyo.path.absolute'] = path.parent / self.generate_asset_name(asset)
             else:
                 # The directory does not yet exist.
-                operations.extend(self.add_directory(path))
+                operations.extend(self.add_directory(Item(path, repo=self.repo)))
         elif not self.repo.is_inventory_dir(path.parent):
-            operations.extend(self.add_directory(path.parent))
+            operations.extend(self.add_directory(Item(path.parent, repo=self.repo)))
 
         # record operation
         operations.append(self._add_operation('new_assets', (asset,)))
         return operations
 
-    def add_directory(self, path: Path) -> list[InventoryOperation]:
+    def add_directory(self, item: Item) -> list[InventoryOperation]:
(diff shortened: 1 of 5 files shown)


onyo/lib/inventory.py
```diff
@@ -472,8 +490,9 @@ def modify_asset(self, asset: dict | UserDict | Path, new_asset: dict | UserDict
         # If a change in is.directory is implied, do this first:
         if asset.get("onyo.is.directory", False) != new_asset.get("onyo.is.directory", False):
             # remove or add dir aspect from/to asset
-            ops = self.add_directory(asset["onyo.path.absolute"]) if new_asset.get("onyo.is.directory", False)\
-                else self.remove_directory(asset["onyo.path.absolute"])
+            ops = self.add_directory(Item(asset["onyo.path.absolute"], repo=self.repo)) \
```
bpoldrack, line 493: Not sure, but I think `asset` should already be an `Item` at this point. Hence, `self.add_directory(asset)` should be fine. Did you check?

bpoldrack:
Overall looks right. Didn't check the test failure, but the general approach is how it's intended indeed.

onyo/lib/inventory.py
bpoldrack: See line 470: 
asset = self.get_asset(path) if isinstance(asset, Path) else asset

So, yes: `asset` already is the item you want to pass.

aqw: It's not an `Item` though. It's a `UserDict`.

bpoldrack: Ok. Then leave it to be straightened in context of `set` (main user of `modify_asset`).

aqw:
> Overall looks right. Didn't check the test failure, but the general approach is how it's intended indeed.

Test failure is fixed (by simply removing the `onyo_cat()` commit).

Issue #673, opened 2024-10-15
Meta: `onyo_get` support for directories and templates


aqw:
This is a tracking, meta issue to allow `onyo get` to select the type of data source to query (other than just assets).

The objective is to add support for directories and templates. This result would obviate the need for an `onyo list` (#53), `onyo template` (#552), and address #672.

- [ ] #676 (default to `asset`; the present behavior)
- [ ] #674
- [ ] #675

---

Open Questions:
- what are the implications for `onyo set` and `onyo unset`
  - they both exclusively operate on assets (which is very, very sane)
  - but if one can query `template` keys, one naturally expects that one can `set` them...

bpoldrack:
Just adding another open question:

Directories could be "matchable" in various ways. They do have some metadata themselves (empty, not empty, glob, creation/modification time, etc.)
So, how is `--match` interpreted, when `--type=directory`?

- not allow it at all?
- have special expressions that `--match` would accept in that mode: some notion of  `contains`  (including `no asset`)?
- match assets within via `--match` and have a separate option to be used for directory matching

bpoldrack:
> but if one can query template keys, one naturally expects that one can set them...

Seems valid to me. I think this would suggest to add a `--type=template|asset` to `set`, `unset` and potentially even `new`.
I'd give that low priority ATM, though.

nhjjr:
I'm in favor of this; I missed this recently when adding devices but wasn't sure what the templates were called. It's of course solved with `ls .onyo/templates` but that's only because I already know where the folder is located.

aqw:
Yeah, tab completion is meant to alleviate this somewhat when creating assets with `onyo new`. But I agree that it's an incomplete solution.

PR #749, opened 2025-02-06
Allow querying OnyoRepo for paths of all kinds of items

bpoldrack:
- This enables `OnyoRepo` to be queried for paths of different types of items (Assets, Dirs, Templates) in preparation of the respective query modes for `onyo get` (see #673).
- In the process `Inventory.get_asset()` and `OnyoRepo.get_asset_paths()` are renamed b/c they account for all items, not just assets.
- Enhances the `onyorepo` fixture to take an `inventory_templates` marker

### DIFF
--- a/onyo/lib/inventory.py
+++ b/onyo/lib/inventory.py
@@ -426,7 +426,7 @@ class Inventory(object):
         #             registered modify operations. It's easier to not force compliance here, but simply let
         #             modify_asset generate the name and pass it.
         generated_name = self.generate_asset_name(
-            self.get_asset(path)
+            self.get_item(path)
             if isinstance(asset, Path) else asset
         )
         if name and name != generated_name:
@@ -449,7 +449,7 @@ class Inventory(object):
         path = asset if isinstance(asset, Path) else Path(asset.get('onyo.path.absolute'))
         if not self.repo.is_asset_path(path):
             raise ValueError(f"No such asset: {path}")
-        asset = self.get_asset(path) if isinstance(asset, Path) else asset
+        asset = self.get_item(path) if isinstance(asset, Path) else asset
 
         # A path change must not be specified w/ modify_asset. A move is a different operation and
         # a renaming has to be derived from content:
@@ -556,8 +556,8 @@ class Inventory(object):
     # non-operation methods
     #
 
-    def get_asset(self, path: Path) -> UserDict:
-        # read and return Asset
+    def get_item(self, path: Path) -> Item:
+        r"""Get an inventory ``Item`` from ``path``."""
         return Item(path, self.repo)
 
     def get_assets(self,
@@ -585,9 +585,9 @@ class Inventory(object):
         Generator of dict
            All matching assets in the inventory.
         """
-        for p in self.repo.get_asset_paths(include=include, exclude=exclude, depth=depth):
+        for p in self.repo.get_item_paths(include=include, exclude=exclude, depth=depth, types=['assets']):
             try:
-                yield self.get_asset(p)
+                yield self.get_item(p)
             except NotAnAssetError as e:
                 # report the error, but proceed
                 ui.error(e)
--- a/onyo/lib/items.py
+++ b/onyo/lib/items.py
@@ -211,7 +211,7 @@ class Item(DotNotationWrapper):
         """Initializer for the 'onyo.is.empty' pseudo-key."""
         if self['onyo.is.directory'] and self.repo and self._path:
             # TODO: This likely can be faster when redoing/enhancing caching of repo paths.
-            return not any(p.parent == self._path for p in self.repo.get_asset_paths())
+            return not any(p.parent == self._path for p in self.repo.asset_paths)
         return None
 
 # TODO/Notes for next PR(s):
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
         Generator,
         Iterable,
         List,
+        Literal,
     )
     from collections import UserDict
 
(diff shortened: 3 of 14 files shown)


onyo/lib/inventory.py
```diff
@@ -556,8 +556,8 @@ def rename_directory(self, src: Path, dst: str | Path) -> list[InventoryOperatio
     # non-operation methods
     #
 
-    def get_asset(self, path: Path) -> UserDict:
-        # read and return Asset
+    def get_item(self, path: Path) -> Item:
+        r"""Get an inventory `Item` from `path."""
```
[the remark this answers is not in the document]
bpoldrack: Thx!

onyo/lib/onyo.py
```diff
@@ -563,49 +564,73 @@ def validate_anchors(self) -> bool:
...
         # Note: The if-else here doesn't change result, but utilizes `GitRepo`'s cache:
         files = self.git.get_subtrees(include) if include else self.git.files
         if depth:
-            roots = include if include else [self.git.root]
             files = [f
                      for f in files
-                     for r in roots
-                     if r in f.parents and len(f.parents) - len(r.parents) <= depth]
+                     for root in include
+                     if root in f.parents and len(f.parents) - len(root.parents) <= depth]
```
aqw, line 610: Nit: I prefer to group the statement explicitly:
```python
f root in f.parents and (len(f.parents) - len(root.parents) <= depth)]
```

onyo/lib/onyo.py
```diff
@@ -563,49 +564,73 @@ def validate_anchors(self) -> bool:
...
             exclude = [exclude] if isinstance(exclude, Path) else exclude
             files = [f for f in files if all(f != p and p not in f.parents for p in exclude)]
 
-        # This only checks for `is_inventory_path`, since we already
-        # know it's a committed file:
-        return [f for f in files if self.is_inventory_path(f)] + \
-               [f.parent for f in files if f.name == self.ASSET_DIR_FILE_NAME]
+        paths = []
+        if 'assets' in types:
+            # This only checks for `is_inventory_path`, since we already
```
aqw, line 618: I don't find this comment helpful. Nothing in this function would lead to believe that we're modifying the repo. So a statement that files are committed is superfluous, and could just as relevantly precede the majority of lines in the codebase.

onyo/lib/onyo.py
```diff
@@ -563,49 +564,73 @@ def validate_anchors(self) -> bool:
 
         return True
 
-    def get_asset_paths(self,
-                        include: Iterable[Path] | None = None,
-                        exclude: Iterable[Path] | Path | None = None,
-                        depth: int = 0
-                        ) -> List[Path]:
-        r"""Select all assets in the repository that are relative to the given
+    def get_item_paths(self,
```
aqw, line 567: In general, seeing all the logic in this method, I wonder if this method should be retired in favor of `get_items()` (which has not yet been converted) and have users query `[onyo.path.*]` according to their needs.

That very well be out of the scope of this PR, but I wanted to mention it.

aqw (approved):
Overall looks great. Quite straightforward.

I left some nits, mostly regarding the docstring and comments. None are blockers, but all are relatively quick to fix.

I did leave some commentary regarding how some of these functions work. But those are not part of the body of work you're trying to get done in this PR. So they are not blockers at all. But I wanted to mention them while we're beginning to warp our minds (and this code) to think in Item-space.

onyo/lib/onyo.py
```diff
@@ -563,49 +564,73 @@ def validate_anchors(self) -> bool:
...
+        #       there's no need to iterate over alö files in git. Conversely, w/o 'templates' there's nothing of
+        #       interest underneath `.onyo/` at all.
+        if types is None:
+            types = ['assets']
+        if include is None:
+            include = [self.git.root]
         if depth < 0:
             raise ValueError(f"depth must be greater or equal 0, but is '{depth}'")
         # Note: The if-else here doesn't change result, but utilizes `GitRepo`'s cache:
         files = self.git.get_subtrees(include) if include else self.git.files
```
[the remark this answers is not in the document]
bpoldrack: Agree, that this distinction should be made here but in `GitRepo` (if at all).

onyo/lib/onyo.py
bpoldrack: Well, may be I should rephrase instead. The intuitive thing to check here, seems to be `is_asset_path` rather than `is_inventory_path`. The point here is, that we don't check an arbitrary, user-provided path, but one that we already know to be a committed file. Hence, what remains to check is cheaper.

onyo/lib/onyo.py
bpoldrack: Not sure how it really should be done, but I agree that this probably should be done differently.
Just not an important part of switching over to the new powers.

onyo/lib/items.py
```diff
@@ -211,7 +211,7 @@ def is_empty(self) -> bool | None:
         """Initializer for the 'onyo.is.empty' pseudo-key."""
         if self['onyo.is.directory'] and self.repo and self._path:
             # TODO: This likely can be faster when redoing/enhancing caching of repo paths.
-            return not any(p.parent == self._path for p in self.repo.get_asset_paths())
+            return not any(p.parent == self._path for p in self.repo.asset_paths)
```
aqw, line 214: Should this be `self.repo._asset_paths`?

aqw (approved):
LGTM. One thing I'm unsure on. Let me know, and then I'll merge.

onyo/lib/items.py
aqw: Perfect. Merging!

PR #752, opened 2025-02-10
DOC/RF: Lots of docstrings; some `Item`-ization

aqw:
This was an accidental adevnture while I was working on something else. I was motivated for a time to improve docstrings, and so i ran with it while I had the motivation.

When I work on docstrings, I usually do a review (of varying depth) of the code it covers. Thus, there are a few code changes sprinkled in here as well.

And then a few more functions in `inventory.py` have been `Item`-ized.

### DIFF
--- a/onyo/lib/command_utils.py
+++ b/onyo/lib/command_utils.py
@@ -47,20 +47,13 @@ def allowed_config_args(git_config_args: list[str]) -> bool:
         A disallowed flag is detected.
     """
 
-    forbidden_flags = ['--system',
-                       '--global',
-                       '--local',
-                       '--worktree',
-                       '--file',
-                       '--blob',
-                       '--help',
-                       '-h',
-                       ]
+    forbidden_flags = ['--system', '--global', '--local', '--worktree',
+                       '--file', '--blob',
+                       '--help', '-h']
+    if any([x in forbidden_flags for x in git_config_args]):
+        raise ValueError("The following options cannot be used with onyo config:\n%s\nExiting. Nothing was set." %
+                         '\n'.join(forbidden_flags))
 
-    for a in git_config_args:
-        if a in forbidden_flags:
-            raise ValueError("The following options cannot be used with onyo config:\n%s\nExiting. Nothing was set." %
-                             '\n'.join(forbidden_flags))
     return True
 
 
--- a/onyo/lib/command_utils.py
+++ b/onyo/lib/command_utils.py
@@ -27,24 +27,26 @@ log: logging.Logger = logging.getLogger('onyo.command_utils')
 def allowed_config_args(git_config_args: list[str]) -> bool:
     r"""Check a list of arguments for disallowed ``git config`` flags.
 
-    ``git-config`` stores configuration information in a variety of locations.
-    This makes sure that such location flags aren't in the list (and ``--help``).
+    ``git-config`` stores configuration information in a variety of locations
+    using location flags (e.g. ``--system``). Onyo uses ``--file`` to write to
+    :py:data:`onyo.lib.onyo.OnyoRepo.ONYO_CONFIG`.
 
-    A helper for ``onyo_config()``.
+    This function makes sure that such flags (and ``--help``) aren't in the
+    list.
+
+    A helper for py:func:`onyo.lib.commands.onyo_config`.
 
     Parameters
     ----------
     git_config_args
-        The list of arguments to pass to ``git config``.
+        List of arguments to pass to ``git config``.
 
     Raises
     ------
     ValueError
-        If a disallowed flag is detected.
+        A disallowed flag is detected.
     """
 
-    # git-config supports multiple layers of git configuration. Onyo uses
-    # ``--file`` to write to .onyo/config. Other options are excluded.
     forbidden_flags = ['--system',
                        '--global',
                        '--local',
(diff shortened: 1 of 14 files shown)


onyo/lib/git.py
```diff
@@ -354,65 +329,63 @@ def get_hexsha(self,
...
-          Path to a file containing exclude patterns to evaluate.
+            Path to a file containing exclude patterns (in the style of ``.gitignore``).
         paths
-          Paths to check against the patterns in `ignore`.
-
-        Returns
-        -------
-        list of Path
-          Paths in `paths` that are excluded by the patterns in `ignore`.
+            Paths to check
```
TobiasKadelka, line 353: Normally we still have a dot at the end of these lines.

onyo/lib/git.py
```diff
@@ -354,65 +329,63 @@ def get_hexsha(self,
...
-    def _parse_log_output(self, lines: list[str]) -> dict:
-        """Generate a dict from the output of ``git log`` for one commit.
+        A helper for :py:func:`history`.
 
-        Internal helper that converts a list of git-log-output lines of
-        a single commit to a dictionary.
+        Parameters
+        ----------
+        lines
+            List of text from ``git log`` of a single commit.
```
TobiasKadelka, line 383: I think I know what this explanation means, but it is the least clear parameter explanation I read so far, maybe it can be descripted more clearly.

Is each element in the list one line of the `git log` output, including some for empty lines, some for the operation record and so on?

onyo/lib/git.py
```diff
@@ -450,24 +423,28 @@ def _parse_log_output(self, lines: list[str]) -> dict:
...
 
         Parameters
         ----------
-        path:
-          What file to follow. If `None`, get the history of HEAD instead.
-        n:
-          Limit history going back ``n`` commits.
-          ``None`` for no limit (default).
+        path
+            The Path to get the history of. Defaults to ``HEAD`` (default).
```
TobiasKadelka, line 437: Now this sounds as if "HEAD" was a path like every other.

I know explaining this clearly is tricky, but I still want to highlight that I don't fully get it and feel like I have to guess.

onyo/lib/command_utils.py
```diff
@@ -141,18 +135,18 @@ def inline_path_diff(source: str | Path,
 
 def intersect_index(seq1: Sequence,
                     seq2: Sequence) -> Tuple[int | None, int | None]:
-    """
-    Compare two sequences. If a match is found, return both indexes of the match.
+    r"""Get the indexes of the next matching value of two sequences.
```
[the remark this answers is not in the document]
aqw: Good catch. Fixed. :-)

onyo/lib/git.py
```diff
@@ -157,33 +151,27 @@ def get_subtrees(self,
...
         return files
 
     def is_clean_worktree(self) -> bool:
-        r"""Check whether the git worktree is clean.
-
-        Returns
-        -------
-        bool
-          True if the git worktree is clean, otherwise False.
+        r"""Whether the git worktree is clean.
```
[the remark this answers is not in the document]
aqw: Yeah, the thing is that having a "Returns" section suppresses the auto generation of a Returns section from the type annotations.

So we either:
1) define everything twice, and have Returns be unlike all other documentation
2) don't define Returns, and describe in the description if there anything interesting about the contents of what is returned (beyond the already-generated type).

We have been pursuing the latter for awhile now, and IMO is the better choice.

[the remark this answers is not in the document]
TobiasKadelka: Okay, thanks for clarifying!
That makes sense, I just wasn't aware of the change :)

onyo/lib/command_utils.py
```diff
@@ -141,18 +135,18 @@ def inline_path_diff(source: str | Path,
...
 
     Returns ``None`` as the indexes if no match is found.
 
     Parameters
     ----------
     seq1
-        First sequence.
+        First sequence
```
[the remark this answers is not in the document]
aqw: Oh, and for what it's worth, it's not defined in numpy no enforced by `ruff`.

onyo/lib/git.py
aqw: Yeah. I bumped on that too. Fixed.

onyo/lib/git.py
aqw: I have tweaked the wording here a bit, but it's not worth much effort as it is a private function that is only ever used in the context of someone who is actively altering `GitRepo.history()`.

Indeed, your understanding is correct. It is a list of lines of text for one commit that is output from `git log`.

onyo/lib/command_utils.py
```diff
@@ -27,48 +27,42 @@
...
 
 def inline_path_diff(source: str | Path,
                      destination: str| Path) -> str:
-    """
-    Generate an inline diff of two paths.
+    r"""Generate an inline diff of two paths.
 
-    Renaming (i.e. changing the last element) is its own action, and does not
-    group with other changes. Moves group with adjacent changes when possible.
+    A rename (i.e. changing the last element) is its own action, and does not
```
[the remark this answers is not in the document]
aqw: I have put "move" and "rename" in bold.

onyo/lib/git.py
aqw: Thanks for the review! I have addressed most issues.

I responded to the `Returns` section question. I feel fairly strongly about the path we have chosen (and is the path chosen by other projects).

The periods I just don't care enough about.

aqw:
Thanks for the review! I have addressed most of your comments.

I responded to the Returns section question. I feel fairly strongly about the path we have chosen (and is the path chosen by other projects).

The periods I just don't care enough about.

PR #757, opened 2025-02-11
Itemize rename_asset and modify_asset

bpoldrack:
This is making sure `Inventory.modify_asset()`, `Inventory.rename_asset()`, and their callers pass `Item`s around rather than paths.


aqw (approved):
Beautiful. Really straightforward, and it's nice to see things beginning to simplify.

PR #758, opened 2025-02-11
Itemize add_asset, remove_asset, remove_directory

bpoldrack:
This ensures that the above mentioned `Inventory` methods are always passed `Item`s by their callers.
In the process two more aspects are addressed:

- Register both `remove_*` operations with an `Item` as operand.
- Change `onyo_rm` and `Inventory.remove_directory()` to not remove any non-inventory files regardless of `recursive`. What's `.onyoignore`'d is ignored by onyo. Period.

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -24,6 +24,7 @@ from onyo.lib.consts import (
     SORT_DESCENDING,
 )
 from onyo.lib.exceptions import (
+    InvalidArgumentError,
     InvalidAssetError,
     InventoryDirNotEmpty,
     NotADirError,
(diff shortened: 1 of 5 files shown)


onyo/lib/commands.py
```diff
@@ -1062,7 +1067,7 @@ def onyo_rm(inventory: Inventory,
         if ui.request_user_response("Save changes? No discards all changes. (y/n) "):
             if auto_message:
                 operation_paths = sorted(deduplicate([  # pyre-ignore[6]
-                    op.operands[0].relative_to(inventory.root)
+                    op.operands[0]['onyo.path.relative']
```
aqw, line 1070: This just looks delicious.

aqw (approved):
Really like the simplification. Especially the reduction of cognitive complexity in `remove_directory()`.

Will merge. :-D

Issue #250, opened 2022-11-02
DOC: test fixtures etc


aqw:
We have now multiple fixtures to generate existing repositories with contents for our tests.
These should be explained, either in the readme, or in the readthedocs under a new section about development.

TobiasKadelka:
I updated the initial empty description with what I think this issue wants to see addressed.

PR #760, opened 2025-02-13
RF/DOC: another sprawling uplift

aqw:
This was not the PR I intended to open, but it happened. :laughing: 

The body of the work is docstrings (and various other string work). However, this also does some refactoring, as I review all the code while writing docs for it.

There is one major change in functionality:
- asset names are no longer required to be unique. This finishes up on work removing that test from `fsck` a few PRs ago.

Assorted bits:
- (almost) all test fixtures are available in all scopes
- 3 more `ruff` docstring rules are enabled.
- the Faker provider is now available at `onyo.lib.faker`
- merge `inventory.get_items_by_query()` -> `inventory.get_items()`
- assorted `Item`-ization

Fixes:
- #501 
- #250


onyo/lib/tests/test_commands_new.py
```diff
@@ -159,7 +161,7 @@ def test_onyo_new_creates_directories(inventory: Inventory) -> None:
 
     # new assets are added
     for s in specs:
-        p = new_directory / inventory.generate_asset_name(s)
+        p = new_directory / inventory.generate_asset_name(Item(s, repo=inventory.repo))
```
bpoldrack, line 164: I would suggest to use this pattern instead:

```suggestion
        p = new_directory / inventory.generate_asset_name(inventory.get_item(s))
```

bpoldrack: Obviously not really important. But I feel it's easier/smoother reading.

bpoldrack (approved):
Awesome!
Don't see anything wrong with it. Only few comments - feel free to decide what of that to ignore or include.

onyo/cli/get.py
```diff
@@ -54,11 +54,9 @@
         type=str,
         default=None,
         help=r"""
-            Criteria to match in the form ``KEY=VALUE``, where **VALUE**
-            is a python regular expression. Pseudo-keys such as ``path`` can
-            also be used. Dictionary subkeys can be addressed using a period
-            (e.g. ``model.name``, ``model.year``, etc.) One can match keys that
-            are not in the output. Special values supported are:
+            Criteria to match in the form ``KEY=VALUE`` — where **VALUE** is a
+            python regular expression. All keys can be matched, and are not
```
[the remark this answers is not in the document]
aqw: Thanks for pointing this out. I have fixed it.

onyo/lib/tests/test_commands_new.py
aqw: Much smoother reading. I saw this in your other PR.  Fixed. :-)

aqw: Actually, I take it back. In this case, we should be doing the awkward thing.

`s` is a spec, and `inventory.get_item()` accepts a `Path`. The point of this entire assignment is to build the path of the asset (that we don't have yet).

So in this case, we're having `generate_asset_name` build the name for us.

In the end, this is all the wrong way to test whether the asset exists, but I don't want to address that in this PR.

PR #762, opened 2025-02-17
DOC/TST/RF: one last giga uplift of docstrings

aqw:
This is the last PR in my recent series of giga PRs making sweeping changes everywhere. Thank you for putting in the effort to review them.

The primary purpose of this PR is to bring docstrings to everything that is not a test. That, along with a spree through the tests to fix the lowest hanging fruit, led to 8 more `ruff` linting rules being enabled.

As usual, there is some refactoring in this PR (details in commit messages):

- some Item()-ization.
- normalization of differs, executors, and recorders to be alphabetized and renamed to be singular.
- removal of dead code from `differs.py` and `argparse_helpers.py`
- `StoreSortOption` sort direction is explicit rather than magical
- add `Item.__eq__()` and retire `is_equal_assets_dict()` and `YAMLDumpWrapper()`

### DIFF
--- a/onyo/lib/items.py
+++ b/onyo/lib/items.py
@@ -45,7 +45,9 @@ class Item(DotNotationWrapper):
     def __init__(self,
                  item: Mapping[_KT, _VT] | Path | None = None,
                  repo: onyo.lib.onyo.OnyoRepo | None = None,
-                 **kwargs: _VT):
+                 **kwargs: _VT) -> None:
+        r"""Initialize an Item."""
+
         super().__init__()
         self.repo: onyo.lib.onyo.OnyoRepo | None = repo
         self._path: Path | None = None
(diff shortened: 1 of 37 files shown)


onyo/lib/items.py
```diff
@@ -94,16 +101,41 @@ def __getitem__(self,
...
-                     key):
+                     key: _KT) -> bool:
+        """Whether ``key`` is in self."""
+
         return super().__contains__(resolve_alias(key))
 
-    def get(self,
-            key,
-            default=None):
+    def __eq__(self,
```
bpoldrack, line 115: First: Thanks! ;-)

Second, though: Technically, this isn't quite correct yet IMO, because it only considers YAML content.
Given how `__eq__` is currently used, that's fine, but there are two issues going forward:
- We want to change to pseudo-keys being settable (at least some of them), so an asset modification could be a move. This implementation would consider a location change no change at all. (`onyo.path.absolute/relative`)
- Now, that this is `Item.__eq__` it seems a bit weird to consider all plain dirs equal regardless of their name and location, when these are exactly the things that would distinguish dirs from eachother. Which boils down to the same thing: Path matters for a general `Item.__eq__`.

This can be addressed later. I don't consider it a blocker. But if you want to ... :)
Alternatively, it could be `Item.equal_content()` instead and figure a notion of `==` later.

bpoldrack (approved):
Great! Thanks much!

There's only one thing and I'd be fine with that not being addressed in this PR.

onyo/lib/items.py
aqw: This is a good point. I intentionally made it a list to allow more pseudo-keys to be checked.

I will add an `equal_content()` function, make it public, and have `__eq__()` check the pseudokeys (including paths) and then call `equal_content()`.

bpoldrack: Wonderful :)

PR #765, opened 2025-02-20
NF/DOC/TST: add `rmdir` command

### LINKED ISSUE #648: Design: `onyo rmdir`
In the spirit of having mirrors/counterparts of commands, we would like `onyo rmdir` as a counterpart to `onyo mkdir`.

```
onyo rmdir [-h] [-m MESSAGE] DIRECTORY [DIRECTORY ...]

Delete empty DIRECTORYs or convert empty Asset Directories into an Asset File.

If the DIRECTORY does not exist, the path is protected, or the asset is
already an Asset File, then Onyo will error and leave everything
unmodified.

Positional Arguments:
  DIRECTORY
                        Directories to delete; or assets to convert into an Asset File.

Options:
  [...]
```

`onyo rmdir` shall:
- convert a childless asset dir into an asset file
- delete an empty dir (that is not an asset dir)

`onyo rmdir` will error when passed:
- a non-directory
- a directory (normal or asset) with any children (asset or dir)

---

`--recursive` was dropped, but preserving the description for posterity.

`--recursive` will follow the same rules as above, just starting from the bottom of the tree. For example:
```
$ onyo tree ASSET_DIR
ASSET_DIR
└── one
    └── two
```
- `onyo rmdir -r ASSET_DIR` will check if `two` is empty, delete, check if `one` is empty, delete, check if `ASSET_DIR` is empty, and convert to an Asset File
  - the actual implementation is simpler:
    - get all children of `ASSET_DIR`
    - if any are non-directory: fail
    - if any are asset (aka asset dir): fail (children cannot be converted because parent will soon be a non-asset-dir)

aqw:
It's the counterpart to `mkdir`. It removes empty directories and (more interestingly) converts asset directories into asset files.

fix #648

### DIFF
--- a/onyo/cli/rmdir.py
+++ b/onyo/cli/rmdir.py
@@ -1,0 +1,62 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from onyo.lib.onyo import OnyoRepo
+from onyo.lib.commands import onyo_rmdir
+from onyo.lib.inventory import Inventory
+from onyo.shared_arguments import (
+    shared_arg_message,
+    shared_arg_no_auto_message,
+)
+
+if TYPE_CHECKING:
+    import argparse
+
+args_rmdir = {
+    'directory': dict(
+        metavar='PATH',
+        nargs='+',
+        help=r"""
+            Directories to delete; or Asset Directories to convert into Asset Files.
+        """
+    ),
+    'message': shared_arg_message,
+    'no_auto_message': shared_arg_no_auto_message,
+}
+
+epilog_rmdir = r"""
+.. rubric:: Examples
+
+Remove a user from a group:
+
+.. code:: shell
+
+    $ onyo rmdir accounting/Bingo\ Bob/
+
+Convert an empty Asset Directory into an Asset File:
+
+.. code:: shell
+
+    $ onyo rmdir accounting/Bingo\ Bob/laptop_apple_macbook.oiw629/
+
+"""
+
+
+def rmdir(args: argparse.Namespace) -> None:
+    r"""
+    Delete empty **DIRECTORY**\ s or convert empty Asset Directories into Asset
+    Files.
+
+    If the **DIRECTORY** does not exist, the path is protected, or the asset is
+    already an Asset File, then Onyo will error and leave everything
+    unmodified.
+    """
+
+    dirs = [Path(d).resolve() for d in args.directory]
+    inventory = Inventory(repo=OnyoRepo(Path.cwd(), find_root=True))
+    onyo_rmdir(inventory,
+               dirs=dirs,
+               message='\n\n'.join(m for m in args.message) if args.message else None,
+               auto_message=False if args.no_auto_message else None)
--- a/onyo/cli/tests/test_rmdir.py
+++ b/onyo/cli/tests/test_rmdir.py
@@ -1,0 +1,184 @@
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from onyo.lib.items import Item
+
+
+assets = [
+    Item({
+        "type": "type",
+        "make": "make",
+        "model.name": "model",
+        "serial": "001",
+        "onyo.is.directory": True,
+        "onyo.path.relative": Path("subdir") / "type_make_model.001",
+    }),
+    Item({
+        "type": "type",
+        "make": "make",
+        "model.name": "model",
+        "serial": "002",
+        "onyo.is.directory": True,
+        "onyo.path.relative": Path("subdir") / "type_make_model.002",
+    }),
+    Item({
+        "type": "type",
+        "make": "make",
+        "model.name": "model",
+        "serial": "003",
+        "onyo.is.directory": False,
+        "onyo.path.relative": Path("subdir") / "type_make_model.003",
+    }),
+    Item({
+        "type": "type",
+        "make": "make",
+        "model.name": "model",
+        "serial": "004",
+        "onyo.is.directory": False,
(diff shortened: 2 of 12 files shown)


onyo/lib/commands.py
```diff
@@ -1084,6 +1084,67 @@ def onyo_rm(inventory: Inventory,
...
+    ------
+    ValueError
+        ``dirs`` is empty.
+    """
+
+    if auto_message is None:
+        auto_message = inventory.repo.auto_message
+
+    if not dirs:
+        raise ValueError("At least one directory path must be specified.")
```
bpoldrack, line 1119: I do wonder whether that should be a `NoopError`, too.
If one programmatically assembles a list of dirs to remove, then ending up with an empty list or a list of things that lead to `NoopError`s is probably not different in how one wants to deal with that.
So:
```
try:
    onyo_rmdir(mylist)
except NoopError:
    # that's fine
    pass
```

would be convenient. Whereas catching `NoopError` and `ValueError` needs string comparison in addition, b/c there could be a different `ValueError` from deeper down that's actually a problem and not "nothing to do".

bpoldrack: Not a blocker at all, just something that I think would be nice if we get that sort of thing sorted out better.

bpoldrack (commented):
Nice! Left a couple things for your consideration. :)

onyo/cli/rmdir.py
```diff
@@ -0,0 +1,62 @@
...
+
+"""
+
+
+def rmdir(args: argparse.Namespace) -> None:
+    r"""
+    Delete empty **DIRECTORY**\ s or convert empty Asset Directories into Asset
+    Files.
+
+    If the **DIRECTORY** does not exist, the path is protected, or the asset is
```
TobiasKadelka, line 52: > the path is protected

Does this include paths that are `.onyoignore`d?

onyo/cli/rmdir.py
```diff
@@ -0,0 +1,62 @@
...
+.. code:: shell
+
+    $ onyo rmdir accounting/Bingo\ Bob/laptop_apple_macbook.oiw629/
+
+"""
+
+
+def rmdir(args: argparse.Namespace) -> None:
+    r"""
+    Delete empty **DIRECTORY**\ s or convert empty Asset Directories into Asset
```
TobiasKadelka, line 49: Does it work on dirs without an anchor file, i.e. actual empty dirs, or only if tracked with onyo?

onyo/cli/rmdir.py
```diff
@@ -0,0 +1,62 @@
...
+
+if TYPE_CHECKING:
+    import argparse
+
+args_rmdir = {
+    'directory': dict(
+        metavar='PATH',
+        nargs='+',
+        help=r"""
+            Directories to delete; or assets to convert into an Asset File.
```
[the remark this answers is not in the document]
aqw: I agree. Thanks for catching this.

onyo/cli/rmdir.py
aqw: It errors:
```
❱ mkdir tst
❱ onyo rmdir tst 
ERROR: Not an inventory directory: /home/aqw/git/inm7/inventory/tst
```

Which IMO is the right thing to do. `onyo mkdir` adds a directory to the inventory; it doesn't cares if it already exists. However, `rmdir` is about removing a directory from the inventory. It makes sense if it errors and days "I know know about this thing; failing".

onyo/lib/commands.py
```diff
@@ -1084,6 +1084,67 @@ def onyo_rm(inventory: Inventory,
...
+
+    dirs = [dirs] if not isinstance(dirs, list) else dirs
+    for d in deduplicate(dirs):  # pyre-ignore[16]
+        try:
+            inventory.remove_directory(Item(d, repo=inventory.repo), recursive=False)
```
bpoldrack, line 1124: One more thing ;-)

```suggestion
            inventory.remove_directory(inventory.get_item(d), recursive=False)
```

onyo/cli/rmdir.py
aqw: Probably? The support for `.onyoignore` is spotty across the codebase. It's present, but I don't think there's been a comprehensive audit yet. IMO, that's beyond the scope of this PR.

onyo/cli/rmdir.py
TobiasKadelka: Yes, perfect. Thank you!

onyo/cli/rmdir.py
bpoldrack: Yes. Nothing works on `.onyoignore`'d stuff.

onyo/cli/tests/test_rmdir.py
```diff
@@ -0,0 +1,192 @@
...
+import pytest
+
+from onyo.lib.items import Item
+
+
+assets = [
+    Item({
+        "type": "type",
+        "make": "make",
+        "model": {"name": "model"},
```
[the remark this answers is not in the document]
aqw: Thanks! I knew that I could use dot notation, but brain farted in the moment. ;-)

onyo/lib/commands.py
aqw: I agree. I have adjusted this for both `rmdir` and `mkdir`. I have not gone through the other commands.

A complete review of all exceptions is its own task.

onyo/cli/rmdir.py
bpoldrack: This is covered in the respective `Inventory` methods.
`Inventory.remove_directory()` in this case.

onyo/lib/commands.py
aqw: Ahh yes. Thank you. Copy-paste from a different age of onyo. ;-)

Issue #713, opened 2024-10-31
Design: deconflate `<unset>` into `<unset>` and `<null>`


aqw:
Currently, `onyo get` returns `<unset>` for both a) keys that are missing and b) keys that exist but have an empty value.

This is also the case when querying (`--match`), making it impossible to differentiate the two states.

This limitation also becomes apparent when deciding upon behavior of stacked templates (#703). A more explicit syntax would remove the ambiguity.

The issue is naming. `<unset>` can be read as both active (unset this key) or passive (I am an unset value). On the positive side, the active form is the exact same verb as `onyo unset`. For that reason, I propose:

- `<unset>`: the key does not exist
- `<null>`: the key exists, but there is no value

The name `<null>` feel intuitive, as `null` is a valid keyword in YAML. We usually use an empty value, but others could use the keyword. And just for trivia, because nothing is ever simple in YAML, the following all mean null:

```
- null  # or Null or NULL
- ~     # yes, a tilde means null...
-       # empty
```

The use cases I see are:

1) `onyo get` output (the values in the output columns)
2) `onyo get --match a='<unset>' b='<null>'`
3) `onyo set --keys a='<unset>' b='<null>'`
4) `onyo new` templates (and thus also subsequent templates, keys, and templates)

These keys accompany `<list>` and `<dict>` (which only declare type, not content; `[]` and `{}` declare empty content).

None of these type-tags shall be written to an asset (as they have no meaning to YAML). They are allowed to be present in a template, because templates (and TSVs) are just alternate forms of passing data to `onyo new --keys`, and they provide meaningful instructions to `onyo new`.

Open Questions:
1) I'm still not thrilled about the name `<unset>`. My next best are `<absent>` and `<missing>`, neither of which feel all that much better.
2) There is an additional type-tag discussed in #703: `<inherit>`. This would leave existing content untouched, but create an empty key if missing. This would be of little to no use for `onyo set`, as `onyo get` can already filter for key/value states. So the only advantage is for `onyo new`. However... everything in `onyo new` acts like a waterfall. `<inherit>` would allow that waterfall to go upstream. So I am skeptical. Plus, the next idea would be "create a key with `X` value if missing." Which there is no clear syntax for. In general, the more I think about this the more I am against this. But I wanted to raise it.

bpoldrack:
Agree on the aim here.

Re naming:
- I kinda get the trouble with `<unset>` , but I can't think of anything better.
[...]

PR #766, opened 2025-02-20
NF/TST: add `<`, `<=`, `>`, and `>=` support to `--match` 

### LINKED ISSUE #719: Design: `onyo get --match` should be able to do `>`/`<`
This is split out from #361.

`onyo get` should be able to do the following comparisons:

- `<`
- `>`
- `<=`
- `>=`

It should use `natsort`'s sorting (which uses the locales `COLLATE`). This also matches the behavior added to the 2024 POSIX spec.

The question is the CLI syntax itself:
```
onyo get --match cpu.cores<4
```

Shell interprets 3/4 of the above symbols as redirection, so they have to be quoted:
```
onyo get --match 'cpu.cores<4'
```

The above syntax no longer always has `=` as a delimiter. Which makes determining what we're telling onyo to do more complicated.  `=` is already implicitly disallowed in the keyname (see #718), so the weirdest we can get is:

- `weird<key=weird>=value`

I propose extending the protected characters list in the keyname to include `<` and `>` (for a total of `.`, `=`, `<`, and `>`.)

At that point, the weirdest onyo will see is:

- `key><weird>=<value`

I propose the following logic:
- find the first `>`, `<`, or `=`
  - if `=`, check if the preceding character is `<` or `>`
    - this determines the comparison mode
- split and/or trim accordingly

Note:
I strongly prefer suffixing the equals sign (e.g. `<=` and `>=`) rather than prefixing.
- everyone does it this way
- it matches how one says it in english "greater than or equal"
- I am generally against limiting characters in the `value`, but am ok with limiting characters in the `key` to reduce complexity.
- reduce ambiguity in the situation `key=<unset>`

Barely related, if onyo is passed a tag (such as `<nope>`) that is not an onyo-reserved tag (such as `<unset>`, `<list>`, etc), should onyo warn? This would only be the case if the entire value is a tag, not just a substring. If so, an issue should probably be opened for that.

aqw:
The primary purpose of this is to allow filtering numerical fields (e.g. `cpu.cores < 4`, etc). However, we get string comparisons for free as well, so filtering according to alphabetical order, etc, is possible (e.g. `type>laptop`).

fixes #719

### DIFF
--- a/onyo/cli/get.py
+++ b/onyo/cli/get.py
@@ -55,13 +55,21 @@ args_get = {
         default=None,
         help=r"""
             Criteria to match in the form ``KEY=VALUE`` — where **VALUE** is a
-            literal string or a python regular expression. All keys can be
+            literal string or a python regular expression. Valid operands are
+            ``=``, ``!=``, ``>``, ``>=``, ``<``, and ``<=``. All keys can be
             matched, and are not limited to those specified by ``--keys``.
-            Special values supported are:
+            Tags supported are:
 
-              * ``<dict>``
-              * ``<list>``
-              * ``<unset>``
+              * ``<bool>``: a boolean
+              * ``<dict>``: a dictionary
+              * ``<empty>``: null or an empty string, dictionary, or list
+              * ``<false>``: ``false``, ``False`` or ``FALSE``
+              * ``<list>``: a list
+              * ``<null>``: ``null``, ``Null``, ``NULL``, ``~``, or nothing
+              * ``<true>``: ``true``, ``True``, or ``TRUE``
+              * ``<unset>``: the key does not exist
+
+            Tags and regular expressions work only with ``=`` and ``!=``.
         """
     ),
 
--- a/onyo/conftest.py
+++ b/onyo/conftest.py
@@ -44,6 +44,7 @@ def clean_env(request) -> None:
     except KeyError:
         pass
 
+
 def params(d: dict) -> MarkDecorator:
     r"""Parameterize a dictionary with human-friendly names.
 
@@ -123,6 +124,7 @@ class AnnotatedGitRepo(GitRepo):
 
         super().__init__(path, find_root)
         self.test_annotation = None
+
 
 @contextmanager
 def fixture_gitrepo(tmp_path: Path,
@@ -208,6 +210,7 @@ def gitrepo_session_scope(tmp_path: Path,
 
     with fixture_gitrepo(tmp_path, request) as result:
         yield result
+
 
 ########################################
 #
@@ -739,6 +742,7 @@ def fake_class_scope() -> Generator:
 @pytest.fixture(scope='module')
 def fake_module_scope() -> Generator:
     r"""Scope the ``fake`` parameter fixture for modules."""
+
     with fixture_fake() as result:
         yield result
 
@@ -746,5 +750,6 @@ def fake_module_scope() -> Generator:
 @pytest.fixture(scope='session')
 def fake_session_scope() -> Generator:
     r"""Scope the ``fake`` parameter fixture for sessions."""
+
     with fixture_fake() as result:
         yield result
--- a/onyo/conftest.py
+++ b/onyo/conftest.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import locale
 import os
 import subprocess
 from collections.abc import Iterable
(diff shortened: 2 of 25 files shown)


onyo/lib/filters.py
```diff
@@ -109,3 +144,132 @@ def match(self,
...
+        Parameters
+        ----------
+        item
+            Item to compare against.
+        """
+
+        if self.value == UNSET_VALUE:
+            # match if:
+            #   - key is present (even if it is `None` or an empty string)
+            return self.key in item
```
bpoldrack, line 163: I think, it would make sense to say `<unset>` is **not** greater than `<unset>`.
Hence:
```suggestion
            return self.key in item and item[self.key] != UNSET_VALUE
```

WDYT?

aqw: I took the literal meaning of `<unset>` (in that the key does not exist). In my opinion, something is greater than nothing. The same logic follows a bit later with non-empty lists being greater than empty lists.

However, if you see `<unset>` as `<null>`, then it's confusing. IMO, that's why we need #713, to remove this ambiguity. 

If you're ok with it, I'm hoping to work on #713 soon-ish and can address this then.

bpoldrack: Fine with deconflating first, yes.  In that case, the point would be about `<null>` instead.
So, greater than null shouldn't match null.

bpoldrack: As long as we don't forget that w/o the follow-up `_match_equal` is based on a different notion of `<unset>` than `_match_greater` in this commit - fine with merging.

aqw: I was already planning on working on `<null>` next, so I'll actually just add it to this PR.

Then this'll all be consistent. :-)

aqw: Done. :-D

aqw:
Differences with the previous version:

- tags are added (`<empty>`, `<bool>`, etc)
- deconflation of `<unset>` into `<unset>`, `<null>`, and `<empty>` (fix #713)
- bools are implicitly resolved into `True` and `False` (controversial; I opened #769 about this; but it should not block this PR)
- regex is no longer used for `>=` and `<=`
- fixes to bugs I found in test fixtures
- docstring fixes to enable one more `ruff` rule
- absolute shitloads of tests
- controlling locale in the test environment

onyo/cli/get.py
```diff
@@ -55,13 +55,21 @@
         default=None,
         help=r"""
             Criteria to match in the form ``KEY=VALUE`` — where **VALUE** is a
-            literal string or a python regular expression. All keys can be
+            literal string or a python regular expression. Valid operands are
```
bpoldrack, line 58:
```suggestion
            literal string or a python regular expression. Valid operations are
```

Or just "comparisons"?

onyo/conftest.py
```diff
@@ -44,6 +45,26 @@ def clean_env(request) -> None:
     except KeyError:
         pass
 
+
+@pytest.fixture(scope="function", autouse=True)
+def clean_locale(request) -> None:
```
bpoldrack, line 50: Thanks for implementing this as a fixture :)

onyo/lib/filters.py
```diff
@@ -33,30 +39,60 @@ class Filter:
     _arg: str = field(repr=False)
     key: str = field(init=False)
     value: str = field(init=False)
+    operand: str = field(init=False)
```
bpoldrack, line 42: You use "operand" everywhere. But it's the operator / operation. The operands are the arguments to it. So, `value` and `item[self.key]`.

```suggestion
    operator: str = field(init=False)
```

onyo/lib/filters.py
```diff
@@ -76,36 +112,235 @@ def _re_match(text: str,
         except re.error:
             return False
 
-    def match(self,
-              item: Item) -> bool:
-        r"""Does ``item`` match this ``Filter``.
+    def _tags_or_types_match(self,
+                             item: Item) -> bool | None:
+        r"""Whether the tags or types of ``self.value`` equals ``Item[self.key]``.
```
bpoldrack, line 117: Before I saw `_match_equal` it was unclear to me, why there's a potential `None` returned.
May be add to the docstring something like "None if not applicable"?
Alternative:
Don't return None, but a raise a dedicated error when `self.value` isn't any of these tags, so `_match_equals` can
```
try:
    return _tags_or_types_match(...)
except NewException:
    return item[self.key] == self.value
```

bpoldrack (commented):
I like it very much!

Thoroughness of test review is very limited, but I'm happily trusting you with that ;-)

onyo/lib/filters.py
aqw: Embarrassing. Thanks for the catch!

onyo/lib/filters.py
aqw: Yeah, this got weird when I removed regex equality matching from `>=` and `<=`. I wanted to avoid duplicating code, but there's a lot of meaningful checks in here.

I have chosen to raise `ValueError`, as indeed the function expects a tag or an empty literal, and only reaches that point if the value is not that.

I have also added sections to the docstrings for exceptions.

onyo/cli/get.py
aqw: I chose operator. Fixed in the latest push.

PR #770, opened 2025-02-27
BF/NF/TST/DOC: allow multiple `onyo get --match` statements (enabling OR)

### LINKED ISSUE #361: `get` can't do >=, OR
Currently, we can't do queries that involve comparing values via `>=` or combine filters with a logical OR via the CLI.

A query for laptops made in a certain date range (that are not retired or lost - see #359) would need to look something like this:

```
> onyo get -H --filter type=laptop --keys build-date -s | grep -v retired | grep -v lost | grep -v unset

20160301   some/path/laptop_brand_model.serial
...
```

Which would then need to be wrapped in a loop that does string slicing and comparison per line in order to match the range.
Note, also, that the standard reporting of `<unset>` needs to be filtered out in order to allow such processing (or whatever processing in top would need to incorporate knowledge about onyo rather than just being able to handle integers or dates).

I'm not sure whether and to what extend the CLI needs to be able to support this type of query. It may be sufficient to provide a list of asset dicts via python interface, since further filtering/matching is pretty much a one-liner in python where one deals with actual data structures (and types) rather than plain strings. However, at the very least this issue should be record of the problem and eventually point to whatever solution.

### LINKED ISSUE #649: `onyo get --match` siliently discards multiple invocations
I ran an `onyo get` command similar to the following:
```
onyo get --match type=laptop --match type=desktop
```

The result was that `--match type=laptop` was silently ignored and only `--match type=desktop` was used.

I expected that `onyo get` would error, and tell me that `--match` can only be used once.

This may affect other option flags as well.

### LINKED ISSUE #697: Bug: query of subkey of dict that is in a subset of assets results in `TypeError`
Like #696, this is easiest to explore in our inventory repo.

Some assets contain the following structure:
```
connectors:
  mDP: 1
  DP: 2 # 1 x out w/ MST
  HDMI: 1
  USB-A: 4
  USB-B: 1
  3_5: 1
  DVI: 1
```
However, some do not have the `connectors` key.

When addressing the `connectors` key directly, all is well:
```
❱ onyo get --keys connectors
 ──────────── 
  connectors  
 ──────────── 
  <dict>      
  <unset>     
  <dict>      
...
  <dict>      
  <unset>     
 ──────────── 
```
However, addressing a subkey (valid or invalid) fails:
```
❱ onyo get --keys connectors.DVI
ERROR: 'connectors' is not a dictionary.
```

Traceback shows `TypeError`:
```
❱ onyo --debug get --keys connectors.DVI
...
ERROR: 'connectors' is not a dictionary.
  File "/home/aqw/git/inm7/onyo/onyo/lib/utils.py", line 92, in __getitem__
    return effective_dict[parts[-1]]
           ~~~~~~~~~~~~~~^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable

The above exception was the direct cause of the following exception:

  File "/home/aqw/git/inm7/onyo/onyo/main.py", line 543, in main
    args.run(args)
  File "/home/aqw/git/inm7/onyo/onyo/cli/get.py", line 187, in get
    results = onyo_get(inventory=inventory,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aqw/git/inm7/onyo/onyo/lib/commands.py", line 83, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/aqw/git/inm7/onyo/onyo/lib/commands.py", line 537, in onyo_get
    results = [{k: r[k] if k in r and r[k] not in [None, ""] else UNSET_VALUE
                           ^^^^^^
  File "/home/aqw/git/inm7/onyo/onyo/lib/utils.py", line 136, in __contains__
    self.__getitem__(key)
  File "/home/aqw/git/inm7/onyo/onyo/lib/utils.py", line 96, in __getitem__
    raise TypeError(f"'{'.'.join(parts[:-1])}' is not a dictionary.") from e
TypeError: 'connectors' is not a dictionary.
```

aqw:
The big feature is enabling `onyo get` to have multiple `--match` invocations. This allows for an "OR" operator in addition to what was previously just "AND".

The behavior is the exact same as before, and this feature is the equivalent of running multiple `onyo get` statements and deduplicating the results. Having this supported directly in `onyo get` is more performant, and saves the user the effort of deduplicating and re-sorting the merged outputs.

An unrelated (but annoying) bug was fixed, WRT to querying subkeys of a missing key.

fix #361
fix #649
fix #697

### DIFF
--- a/onyo/argparse_helpers.py
+++ b/onyo/argparse_helpers.py
@@ -109,6 +109,39 @@ class StoreSingleKeyValuePairs(argparse.Action):
         setattr(namespace, self.dest, results)
 
 
+class StoreMatchOption(argparse.Action):
+    r"""Store match statements and retain their order across multiple invocations."""
+
+    def __call__(self,
+                 parser: argparse.ArgumentParser,
+                 namespace: argparse.Namespace,
+                 keys: list[str],
+                 option_string: str | None = None) -> None:
+        r"""Store a list of match statements in a list.
+
+        This way multiple invocations of ``--match``  won't have their lists
+        merged together, and can be considered independently of one another
+        (i.e. ``OR``).
+
+        Parameters
+        ----------
+        parser
+            ArgumentParser object that contains this action.
+        namespace
+            Namespace object returned by :py:meth:`argparse.ArgumentParser.parse_args`.
+        keys
+            List of strings containing match statements.
+        option_string
+            Option string used to invoke this action.
+        """
+
+        items = getattr(namespace, self.dest, None)
+        items = list() if items is None else items
+        items.append(keys)
+
+        setattr(namespace, self.dest, items)
+
+
 class StoreSortOption(argparse.Action):
     r"""Store keys-to-sort and retain their order across multiple invoking options.
 
--- a/onyo/cli/get.py
+++ b/onyo/cli/get.py
@@ -3,7 +3,10 @@ from __future__ import annotations
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from onyo.argparse_helpers import StoreSortOption
+from onyo.argparse_helpers import (
+    StoreMatchOption,
+    StoreSortOption,
+)
 from onyo.lib.commands import onyo_get
 from onyo.lib.exceptions import OnyoCLIExitCode
 from onyo.lib.filters import Filter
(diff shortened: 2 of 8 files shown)


bpoldrack (approved):
Another nice enhancement!
Looks good.

PR #772, opened 2025-02-27
BF/RF: use pathspec module for `.{git,onyo}ignore` files

### LINKED ISSUE #767: BUG: onyo edit asks to accept changes even if no changes are made
Not really a problem, but a bit weird: If I open an asset with `onyo edit` and make no changes, just leave my editor again, onyo asks me to accept changes:

```
❱ onyo edit <path>
Accept changes? (y)es / continue (e)diting / (s)kip asset / (a)bort command [Default: yes] y
No assets updated.
```

Since this does not lead to problems and is only a bit awkward, it is okay if we decide this is not worth changing anything, and just close the issue.

But the expected behavior IMO would be that onyo detects that no changes where made, skips the question, and just says "no assets updated".

aqw:
Though `GitRepo.check_ignore()` allows a list of paths to be passed, upper levels of Onyo only ever check one path at a time.

This inefficiency was made even more expensive, as each call spawned a subprocess for `git check-ignore` to check the path against the ignore file.

On our internal repository, this caused over 1,100 calls to `git check-ignore` for a single `onyo get` call.

Thankfully, there is a (seemingly) mature project (pathspec) that understands the patterns used in git ignore files.

Here, I swap out `git check-ignore` for pathspec.

In our internal inventory repository (which has three `.onyoignore` files), it brought that same `onyo get` call (with 1,100 `git check-ignore` invocations) down from an average of 2.80s to 1.18s.

I would love to show numbers from the benchmarking suite, however, the generated repos did not include `.onyoignore` files, and so were unaffected by this.

Also includes two unrelated bugfixes.

fix #767

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -351,11 +351,14 @@ def _edit_asset(inventory: Inventory,
                     # should not be possible
                     raise RuntimeError(f"Unexpected response: {response}")
 
+        # if no edits were made, move on
+        if not operations:
+            break
+
         # show diff and ask for confirmation
-        if operations:
-            ui.print("Effective changes:")
-            for op in operations:
-                print_diff(op)
+        ui.print("Effective changes:")
+        for op in operations:
+            print_diff(op)
 
         response = ui.request_user_response(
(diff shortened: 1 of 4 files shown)


bpoldrack:
Nice find!

PR #776, opened 2025-03-04
BF/RF: fix command length overflow; cache `OnyoRepo.get_config()`

### LINKED ISSUE #727: Bug: onyo fails when creating "too many" (e.g. 5,000) assets
Git explodes; line is too long.

Assigning Ben, because he's already solved this before.

aqw:
Fixes in this PR are:

1) don't error when creating a boatload of assets and thus overflowing the maximum command length (#727) 
2) introduce caching for `OnyoRepo.get_config()`
3) cleanup around temp files (dangling file handles, leaked paths, etc)

---
Regarding 1: I introduced a test to overflow with paths (since the overflow can happen with either the message or paths), but left it disabled by default because it takes 30+ minutes to run on my machine. The paths need to be relatively short (beneath the size of the most limiting OS/FS) in order to run reliably. And then in total exceed the maximum argument length of the most capable OS/FS. This results in 150k+ files, and that understandably takes a long time to run.

So the test is there for reference if we are ever uncertain. But is skipped by default.

---
Regarding 2: The motivation is the significant overhead of querying `onyo.assets.name-format`, which is used on a per-asset basis by `set` and `new`.

Initially, I attempted to introduce a per-file file cache at the GitRepo layer, and track the mtime of files that would be queried. However, in the case of a bare `git config` call, the number of locations is quite a few and are modified by environmental variables. These environmental variables would need to be resolved and tracked as well for changes.

CLI users will not have any issues managing the cache (and the new behavior is arguably an improvement, as Onyo's behavior can no longer change mid-flight if the system config changes in parallel). For users of the API, they only encounter issues if they explicitly avoid `OnyoRepo.set_config()`; in which case `clear_cache()` solves their problems.

This speeds up the benchmark considerably:

- `test_cli_onyo_new_fifty[1000] `: 2.09s -> 1.46s (30%)
- `test_cli_onyo_new_fifty[100]`: 1.12s -> 0.42s (63%)
- `test_cli_onyo_set_fifty[1000]`: 2.51s -> 2.31s (8%)
- `test_cli_onyo_set_fifty[100]`: 1.06s -> 0.79s (25%)

Old times:
```
------------------------------------------ benchmark 'test_cli_onyo_new_fifty': 2 tests ------------------------------------------
Name (time in s)                     Min               Max              Mean            StdDev            Median            Rounds
----------------------------------------------------------------------------------------------------------------------------------
test_cli_onyo_new_fifty[1000]     1.8910 (1.74)     2.1347 (1.88)     2.0508 (1.84)     0.0981 (5.01)     2.0877 (1.87)          5
test_cli_onyo_new_fifty[100]      1.0864 (1.0)      1.1338 (1.0)      1.1162 (1.0)      0.0196 (1.0)      1.1177 (1.0)           5
----------------------------------------------------------------------------------------------------------------------------------

------------------------------------------ benchmark 'test_cli_onyo_set_fifty': 2 tests ------------------------------------------
Name (time in s)                     Min               Max              Mean            StdDev            Median            Rounds
----------------------------------------------------------------------------------------------------------------------------------
test_cli_onyo_set_fifty[1000]     2.4798 (2.38)     2.6454 (2.44)     2.5503 (2.40)     0.0700 (4.48)     2.5134 (2.37)          5
test_cli_onyo_set_fifty[100]      1.0427 (1.0)      1.0849 (1.0)      1.0613 (1.0)      0.0156 (1.0)      1.0589 (1.0)           5
----------------------------------------------------------------------------------------------------------------------------------
```

New times:
```
------------------------------------------ benchmark 'test_cli_onyo_new_fifty': 2 tests ------------------------------------------
Name (time in s)                     Min               Max              Mean            StdDev            Median            Rounds
----------------------------------------------------------------------------------------------------------------------------------
test_cli_onyo_new_fifty[1000]     1.4381 (3.47)     1.6802 (3.61)     1.5081 (3.48)     0.0999 (4.20)     1.4568 (3.49)          5
test_cli_onyo_new_fifty[100]      0.4146 (1.0)      0.4654 (1.0)      0.4329 (1.0)      0.0238 (1.0)      0.4178 (1.0)           5
----------------------------------------------------------------------------------------------------------------------------------

------------------------------------------ benchmark 'test_cli_onyo_set_fifty': 2 tests ------------------------------------------
Name (time in s)                     Min               Max              Mean            StdDev            Median            Rounds
----------------------------------------------------------------------------------------------------------------------------------
test_cli_onyo_set_fifty[1000]     2.2197 (2.83)     2.3716 (2.99)     2.2920 (2.90)     0.0629 (15.11)    2.3075 (2.92)          5
test_cli_onyo_set_fifty[100]      0.7846 (1.0)      0.7940 (1.0)      0.7897 (1.0)      0.0042 (1.0)      0.7911 (1.0)           5
----------------------------------------------------------------------------------------------------------------------------------
```

---
fix #727

### DIFF
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -95,6 +95,7 @@ class OnyoRepo(object):
 
         # caches
         self._asset_paths: list[Path] | None = None
+        self._config_cache: dict[str, dict[str, str]] = {'git': {}, 'onyo': {}}
 
     def set_config(self,
                    key: str,
@@ -124,6 +125,10 @@ class OnyoRepo(object):
         if self.version == '1' and key == 'onyo.assets.name-format':
             key = 'onyo.assets.filename'
 
+        # clear the config cache
+        self._config_cache = {'git': {}, 'onyo': {}}
+
+        # set
         loc = ONYO_CONFIG if location == 'onyo' else location
         return self.git.set_config(key=key, value=value, location=loc)
 
@@ -133,6 +138,12 @@ class OnyoRepo(object):
 
         This first checks git's normal git-config locations and then
         :py:data:`onyo.lib.consts.ONYO_CONFIG` as a fallback.
+
+        The results are cached, which are cleared automatically by
+        :py:func:`commit` and :py:func:`set_config`.
+
+        If changes are made by other means, use :py:func:`clear_cache` to reset
+        the cache.
 
         Parameters
         ----------
@@ -144,11 +155,56 @@ class OnyoRepo(object):
                   KEY = VALUE
         """
 
+        value = None
+
         # repo version shim
         if self.version == '1' and key == 'onyo.assets.name-format':
             key = 'onyo.assets.filename'
 
-        return self.git.get_config(key) or self.git.get_config(key, self.onyo_config)
+        #
+        # check cache
+        #
+        cache_hit = False
+
+        try:
+            # git
+            value = self._config_cache['git'][key]
+            ui.log_debug(f"config '{key}' acquired from cache of git config: '{value}'")
+            cache_hit = True
+        except KeyError:
+            cache_hit = False
+            pass
+
+        if value is None:
+            # onyo
+            try:
+                value = self._config_cache['onyo'][key]
+                ui.log_debug(f"config '{key}' acquired from cache of onyo config: '{value}'")
+                cache_hit = True
+            except KeyError:
+                cache_hit = False
+                pass
+
+        if cache_hit:
+            return value
+
+        #
+        # query actual config files
+        #
+        ui.log_debug(f"config '{key}' cache miss")
+
+        # query the full git config stack
+        value = self.git.get_config(key)
+        self._config_cache['git'][key] = value  # pyre-ignore[6]
+
+        if value is not None:
+            return value
+
+        # query .onyo/config
+        value = self.git.get_config(key, self.onyo_config)
+        self._config_cache['onyo'][key] = value  # pyre-ignore[6]
+
+        return value
 
     @property
     def auto_message(self) -> bool:
@@ -234,6 +290,7 @@ class OnyoRepo(object):
         """
 
         self._asset_paths = None
+        self._config_cache = {'git': {}, 'onyo': {}}
         self.git.clear_cache()
 
     @staticmethod
--- a/onyo/lib/onyo.py
+++ b/onyo/lib/onyo.py
@@ -343,10 +343,10 @@ class OnyoRepo(object):
     def asset_paths(self) -> list[Path]:
         r"""Get the absolute ``Path``\ s of all assets in this repository.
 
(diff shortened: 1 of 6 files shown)


onyo/lib/onyo.py
```diff
@@ -144,11 +155,56 @@ def get_config(self,
...
+        #
+        # query actual config files
+        #
+        ui.log_debug(f"config '{key}' cache miss")
+
+        # query the full git config stack
+        value = self.git.get_config(key)
+        self._config_cache.setdefault('git', {})[key] = value  # pyre-ignore[6]
+
+        if value is not None:
```
TobiasKadelka, line 200: Couldn't this be simply `if value:`?

TobiasKadelka:
LGTM! :)

onyo/lib/onyo.py
aqw: No. What if `value` is `''` or `0`?

bpoldrack: @TobiasKadelka : This would break the moment `self.git.get_config()` returns python types rather than just strings. A config may be set  to `"false"` and the getter could return `False`, since this is in fact what we'd want to deal with. So, your condition would trigger b/c it can't distinguish between nothing being returned and `False` being returned.

onyo/lib/onyo.py
```diff
@@ -124,6 +125,10 @@ def set_config(self,
         if self.version == '1' and key == 'onyo.assets.name-format':
             key = 'onyo.assets.filename'
 
+        # clear the config cache
+        self._config_cache: dict[Path, dict[str, str]] = dict()
```
bpoldrack, line 129: Type annotation is incorrect here. (`Path`)

onyo/lib/onyo.py
```diff
@@ -95,6 +95,7 @@ def __init__(self,
 
         # caches
         self._asset_paths: list[Path] | None = None
+        self._config_cache: dict[str, dict[str, str]] = dict()
```
bpoldrack, line 98: Why not here (and below in `set_config()`) initialize to:
```suggestion
        self._config_cache: dict[str, dict[str, str]] = {'onyo': {}, 'git':{}}
```

instead of calling `setdefault()` everytime (L198 and 205)?

bpoldrack (commented):
Cool!
Two minor comments, otherwise LGTM.

onyo/lib/onyo.py
aqw: Oh nice catch. I fixed the other one, but forgot this. Thanks. :-)

onyo/lib/onyo.py
aqw: I agree. Part of my leftover desire to use `None`, that I abandoned.

I have fixed this as well, and will push soon once the tests pass locally.

PR #780, opened 2025-03-06
Recursive (directory) templates

bpoldrack:
Sits on top of PR #779

### DIFF
--- a/onyo/lib/inventory.py
+++ b/onyo/lib/inventory.py
@@ -857,12 +857,24 @@ class Inventory(object):
                 # report the error, and proceed
                 ui.error(e)
 
-    def get_asset_from_template(self,
-                                template: Path | str | None) -> Item:
-        r"""Get a template as an Item."""
+    def get_templates(self,
+                      template: Path | None,
+                      recursive: bool = False) -> Generator[Item, None, None]:
+        r"""Get templates as Items.
 
-        # TODO: Possibly join with get_asset_content()
-        return Item(self.repo.get_template(template))  # , repo=self.repo ?? Probably not. Template is not yet bound.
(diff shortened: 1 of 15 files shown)


onyo/lib/onyo.py
```diff
@@ -574,19 +577,23 @@ def is_onyo_ignored(self,
 
         return False
 
-    def get_template(self,
-                     path: Path | str | None = None) -> dict:
-        r"""Select a template file and return an asset dict from it.
+    def get_templates(self,
+                      path: Path | None = None,
+                      recursive: bool = False) -> Generator[ItemSpec, None, None]:
+        r"""Select a template file and return a dict-like from it.
```
aqw, line 583: "dict-like" -> ItemSpec

onyo/lib/inventory.py
```diff
@@ -857,12 +857,17 @@ def get_items(self,
...
-    def get_asset_from_template(self,
-                                template: Path | str | None) -> Item:
-        r"""Get a template as an Item."""
-
-        # TODO: Possibly join with get_asset_content()
-        return Item(self.repo.get_template(template))  # , repo=self.repo ?? Probably not. Template is not yet bound.
+    def get_templates(self,
+                      template: Path | None,
+                      recursive: bool = False) -> Generator[Item, None, None]:
+        r"""Get templates as Items."""
```
aqw, line 863: I don't want to get you bogged down in docstrings. But it is preferable if the parameters are also documented.

Not a blocker.

aqw (approved):
LGTM.

`get_templates()` really exposes the limitations we have with `Item` and `ItemSpec`. We'll make that better.

PR #783, opened 2025-03-19
`new`: Rewrite

bpoldrack:
This can't possibly pass any tests yet. Just drafting the thing for now.
Sits on top of #780.

### DIFF
--- a/onyo/lib/commands.py
+++ b/onyo/lib/commands.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import subprocess
+from copy import deepcopy
 from pathlib import Path
 from typing import (
     ParamSpec,
@@ -11,7 +12,7 @@ from typing import (
 from functools import wraps
 
 from rich import box
-from rich.table import Table  # pyre-ignore[21] for some reason pyre doesn't find Table
+from rich.table import Table
 
 from onyo.lib.command_utils import (
     inline_path_diff,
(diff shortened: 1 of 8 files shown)


onyo/lib/commands.py
```diff
@@ -897,63 +907,75 @@ def onyo_new(inventory: Inventory,
...
-        auto_message = inventory.repo.auto_message
-
-    keys = keys or []
-    if not any([keys, edit, template, clone]):
+    if not any([keys, edit, template, clone, yaml]):
         raise ValueError("Key-value pairs or a template/clone-target must be given.")
     if template and clone:
         raise ValueError("'template' and 'clone' options are mutually exclusive.")
 
+    auto_message = auto_message or inventory.repo.auto_message
```
aqw, line 919: Huh. This is new to me.

Just to confirm, this is a shorthand for:
```python
auto_message = inventory.repo.auto_message if auto_message is None else auto_message
```
(or more literally)
```python
auto_message = auto_message if auto_message is not None else iventory.repo.auto_message
```

onyo/lib/commands.py
```diff
@@ -897,63 +907,75 @@ def onyo_new(inventory: Inventory,
...
+    keys = keys or []
+    base = base or Path.cwd()
     # get editor early in case it fails
     editor = inventory.repo.get_editor() if edit else ""
 
-    # Note that `keys` can be empty.
-    specs = deepcopy(keys)
+    layers = ([(t_path, inventory.get_templates(t_path, recursive=recursive)) for t_path in template]
```
aqw, line 925: This would benefit from a comment. :-)

onyo/lib/commands.py
bpoldrack: Not exactly.

It's short for `auto_message if not auto_message else inventory.repo.auto_message`. Which isn't actually what we want, so this was wrong ;-)

Issue #789, opened 2025-04-01
update helptext for each command (coordination issue)


TobiasKadelka:
Commands to update:
- [ ] config
- [ ] edit
- [ ] fsck
- [ ] get
- [ ] history
- [ ] init
- [ ] mkdir
- [ ] mv
- [ ] new
- [ ] rm
- [ ] rmdir
- [ ] set
- [ ] shell-completion
- [ ] show
- [ ] tree
- [ ] tsv-to-yaml
- [ ] unset

TobiasKadelka:
"protected paths" need to be documented in manpage, readthedocs, etc.

PR #790, opened 2025-04-01
WIP: update: command helptext

TobiasKadelka:
Works on https://github.com/psyinfra/onyo/issues/789

Currently updates the helptext of the following commands:
- mkdir


onyo/cli/mkdir.py
```diff
@@ -52,11 +52,8 @@ def mkdir(args: argparse.Namespace) -> None:
     Intermediate directories are created as needed (i.e. parent and child
     directories can be created in one call).
 
-    An empty ``.anchor`` file is added to each directory, to ensure that git
-    tracks them even when empty.
-
-    If a **DIRECTORY** is already a directory or a protected path, then Onyo
-    will error and leave everything unmodified.
+    If a **DIRECTORY** is already a directory or a protected path (see
+    "protected paths"), then Onyo errors and leaves everything unmodified.
```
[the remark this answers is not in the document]
TobiasKadelka: Yeah, I meant this as a place holder to a later manpage about protected paths. It does not exist yet, but this location should lead to it IMO.

I might remove it for now, until we get manpages or decide actually how to do it.
I think there might be a really good short form to mention in the onyo top level command that these concepts exist, but I cannot imagine how to actually explain anything about them in a useful way there, without just spamming the terminal with too much text.

onyo/cli/tree.py
```diff
@@ -35,6 +35,13 @@
 .. code:: shell
 
     $ onyo tree shelf
+
+Show only the room structure of a building without listing assets:
```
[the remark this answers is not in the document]
TobiasKadelka: I will be curious about what Ben says.

But as an option with piping that would later for example show only the items (assuming it is all asset dirs) which match "type is either user, group, or room" (or the reverse, excluding laptops, desktops, monitors) would be quite cool and useful for me.

> reduces the need/value of onyo tree

As for this statement, it is maybe right in the abstract, but I really do like to see the tree-view. `onyo get` can give the information already in a table, but in my opinion I can normally read/parse this specific information faster when having the tree.

onyo/cli/tsv_to_yaml.py
```diff
@@ -37,13 +39,15 @@
...
-    The header declares the key names to be populated. The values to populate
-    documents are declared with one line per YAML document.
+    The output is printed to stdout as a multi-document YAML file, and each
+    document is separated by a ``---`` line. The YAML output can be passed into
+    ``onyo new`` and ``onyo set``.
 
-    The output is printed to stdout as a multiple document YAML file (each
-    document is separated by a ``---`` line).
+    The TSV file's header row declares the key names for all resulting YAML
+    documents, and each line following describes the values OF one the document.
```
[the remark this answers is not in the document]
TobiasKadelka: Lol, capslock error, thanks for catching it.

I'll try to rephrase that sentence in a better way! :)

onyo/cli/mkdir.py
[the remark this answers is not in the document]
TobiasKadelka: I decided to just remove that part for now. We will have to add references to the manpages later anyway in multiple locations, no real sense in trying to do it here now. Thanks! :)

TobiasKadelka:
@aqw Thanks for your feedback!

I went through it and adjusted the help texts :)

onyo/cli/rm.py
```diff
@@ -51,18 +51,24 @@
 .. code:: shell
 
     $ onyo rm --message "Bob retired; he won at bingo" admin/Bingo\ Bob/
+
+Remove a building and all of its assets:
+
+.. code:: shell
+
+    $ onyo rm --recursive --message "The big fire..." buildings/20.3/
```
aqw, line 59: Suggestion: "big" -> "great".

It's always a "great fire". ;-P

https://en.wikipedia.org/wiki/Great_Chicago_Fire
https://en.wikipedia.org/wiki/Great_Seattle_Fire
https://en.wikipedia.org/wiki/Great_Fires_of_1871
https://en.wikipedia.org/wiki/Great_Fire_of_London
https://en.wikipedia.org/wiki/Great_fire_of_Hamburg

onyo/cli/rm.py
```diff
@@ -51,18 +51,24 @@
...
 """
 
 
 def rm(args: argparse.Namespace) -> None:
     r"""
     Delete **ASSET**\ s and/or **DIRECTORY**\ s.
 
-    Directories and asset directories are deleted along with their contents,
-    if the ``--recursive`` flag is set. Otherwise, fails on non-empty directories.
+    If the ``--recursive`` flag is set, directories are deleted along with their
```
aqw, line 68: Not sure if this is a digression... but I'm reflecting on the writing style here.

What do you think about the following?

> When ``--recursive`` is used, directories are deleted along with their

onyo/cli/tree.py
```diff
@@ -45,8 +52,9 @@ def tree(args: argparse.Namespace) -> None:
     If no directory is provided, the tree for the current working directory is
     listed.
 
-    Directories are printed sequentially. If one does not exist, no further
-    trees will be printed and an error is returned.
+    **DIRECTORY**\ s are printed sequentially. If one does not exist, or is not
```
aqw, line 55: Very good correction. The previous version was ambiguous and arguably incorrect. Nice catch. +1.

onyo/cli/tsv_to_yaml.py
```diff
@@ -22,12 +22,13 @@
 epilog_tsv_to_yaml = r"""
 .. rubric:: Examples
 
+Convert a table to a YAML file:
```
aqw, line 25: I bump on the word "convert", since it doesn't convert the file.

`tsv_to_yaml` prints a TSV file as YAML.

And what it produces is YAML text, not "a YAML file".

onyo/cli/tsv_to_yaml.py
```diff
@@ -22,12 +22,13 @@
...
+
 .. code:: shell
 
     $ onyo tsv-to-yaml table.tsv
 
-The resulting multi-document YAML file can be split into multiple YAML files (if
-desired) using ``csplit``.
+Split the resulting multi-document YAML file into separate YAML files:
```
aqw, line 31: There is no "YAML file". Just output.

> Split the resulting multi-document YAML output and save as separate YAML files:

onyo/cli/tsv_to_yaml.py
```diff
@@ -37,13 +38,15 @@
...
 def tsv_to_yaml(args: argparse.Namespace) -> None:
     r"""
-    Convert a TSV file into YAML suitable to pass to ``onyo new`` and ``onyo set``.
+    Print a TSV file's contents as YAML, suitable to pass into ``onyo new`` and
+    ``onyo set``.
 
-    The header declares the key names to be populated. The values to populate
-    documents are declared with one line per YAML document.
+    The output is printed to stdout as a multi-document YAML file, and each
+    document is separated by a ``---`` line.
```
aqw, line 45: Huge nit, but "separated with" is preferred over "separated by". (I had to look this one up).

"By" is an agent. The men fighting were separated by Horatio.

"With" is for instruments. They were separated with a crowbar. The documents were separated with a `---`.

aqw (changes requested):
Looks way better.

I left a few nits. Overall getting close.

I'm also interested in @bpoldrack's thoughts about `tree`.

onyo/cli/rm.py
TobiasKadelka: Thanks, after the first 4 sources I was skeptical, but the fifth one really convinced me! ;)

onyo/cli/rm.py
TobiasKadelka: I personally prefere to start with "if" instead of "when" because the next sentence starts with an "otherwise". The rest of your suggestion I applied :)

onyo/cli/tsv_to_yaml.py
TobiasKadelka: Lol, I love it. It's adjusted!

TobiasKadelka:
Everything was updated and pushed :)

onyo/cli/tree.py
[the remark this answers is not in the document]
bpoldrack: > And now that I say that.... I look it up and unix's tree supports --fromfile. So right now, we already can do:

> `onyo get --machine-readable --types assets directories --keys onyo.path.relative --match type!=laptop | tree --fromfile`

I do agree. The idea of what is or isn't a "structural" dir really depends on the user. There's nothing inherently different about it. So, leaving it to a customizable `get` is the right way, I think.

> Which.... IMO really reduces the need/value of onyo tree. To the point where we might want to discuss retiring it.

Just talked to @aqw about this. `onyo tree` could become a thin wrapper around that construction, allowing to pass options for `onyo get` and possibly `tree`. Simply because this feature of `tree` isn't something frequently encountered/used. Docs should then point out, that the command is just convenience around `onyo get | tree`.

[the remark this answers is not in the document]
aqw: I have opened #793

aqw:
Please fix the `ruff` lint failures and then this PR is ready to merge. :-D

TobiasKadelka:
> Please fix the ruff lint failures and then this PR is ready to merge. :-D

I did fix the ruff link failures and now I think this PR is ready to merge :-D

PR #797, opened 2025-08-25
NF: use uv to install packages

TobiasKadelka:
Install packages with `uv` instead of `pip`.

As a bonus, this allows us to use `uvx ruff check` (ruff build in into uv) which removes a dependency from `pyproject.toml`.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -29,13 +29,14 @@ jobs:
         python -m pip install --upgrade pip
         # get proper PyICU (for natsort)
         sudo apt-get install pkg-config libicu-dev
+        pip install uv
         pip install --no-binary=:pyicu: pyicu
         # install onyo
-        pip install -e ".[tests]"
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -29,13 +29,16 @@ jobs:
         python -m pip install --upgrade pip
         # get proper PyICU (for natsort)
         sudo apt-get install pkg-config libicu-dev
+        pip install uv
         pip install --no-binary=:pyicu: pyicu
         # install onyo
-        pip install -e ".[tests]"
+        uv venv --python 3.13 onyo-venv
```
aqw, line 35: This is problematic, as it effectively overrides the python version matrix declared above.

aqw (changes requested):

aqw (approved):
LGTM. Thanks!

aqw:
lol; thanks for the ping. I thought I had clicked "merge" yesterday. :-)

